PR #35, opened 2024-07-11
Fix/conveyor auth without OIDC

stijn-janssens:


backend/app/core/conveyor/notebook_builder.py
```diff
@@ -13,6 +13,7 @@ def __init__(self):
         self.conveyor_api = "https://app.conveyordata.com/api/v2"
 
         self.oidc = OIDCConfiguration(
+            oidc_disabled=False,
```
andrii-ivaniuk, line 16: Should we use `get_boolean_variable("OIDC_DISABLED", True)` and not in `OIDCConfiguration.__init__`?

andrii-ivaniuk (approved):

PR #40, opened 2024-07-16
Added isort and pydantic-settings

andrii-ivaniuk:


backend/pyproject.toml
```diff
@@ -27,6 +27,12 @@ shortuuid = "^1.0.13"
 pytz = "^2024.1"
```
stijn-janssens, line 27: you will need to update the poetry version to 3.12 as well

backend/app/settings.py
```diff
@@ -0,0 +1,39 @@
+from functools import lru_cache
```
stijn-janssens, line 1: Great improvement!

backend/app/settings.py
```diff
@@ -0,0 +1,39 @@
...
+from typing import Optional
+
+from pydantic import HttpUrl
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class Settings(BaseSettings):
+    model_config = SettingsConfigDict(env_file=".env")
```
stijn-janssens, line 9: This will be overwritten by the same ENV variables being populated right?

backend/app/data_products/service.py
```diff
@@ -246,7 +244,8 @@ def update_data_product_about(
         current_data_product.about = data_product.about
         db.commit()
 
-    def ensure_owner(self, authenticated_user: User, data_product: DataProduct):
+    @staticmethod
```
stijn-janssens, line 247: I would not necessarily make this a static method? Even though it does not use `self` ?

backend/pyproject.toml
andrii-ivaniuk: What do you mean by saying to update the poetry version?

backend/app/settings.py
andrii-ivaniuk: yes

backend/app/data_products/service.py
andrii-ivaniuk: I just updated it here. It has no impact. Just no extra param is used in the function definition

stijn-janssens: We might need to think about this in a broader picture. e.g. in the `DatasetService` class, there is no method that can't be a `staticmethod` currently. But I don't think that warrants making all of the methods `staticmethod`. Not sure what you think. It does not harm to do it, but there is also limited advantages

stijn-janssens (approved):

PR #41, opened 2024-07-16
Add admin support

andrii-ivaniuk:


backend/app/users/schema.py
```diff
@@ -10,6 +10,7 @@ class UserCreate(ORMModel):
     external_id: str
     first_name: str
     last_name: str
+    is_admin: bool
```
stijn-janssens, line 13: I would add the `is_admin` field to the `UserGet` and the `User` schema, but not to the `UserCreate` schema. This way we (currently) can only add admins directly in the database or through seeding, and the frontend is not required to send the boolean parameter at creation time.

andrii-ivaniuk: fixed

stijn-janssens (approved):

PR #42, opened 2024-07-16
CORS allowed hosts need to be passed as list

stijn-janssens:


stijn-janssens:
Needs to be fixed in settings.py @andrii-ivaniuk will look at it

PR #44, opened 2024-07-18
Added unit tests to test is_admin flag

andrii-ivaniuk:


backend/tests/conftest.py
```diff
@@ -164,6 +184,36 @@ def default_data_product(
...
+def admin_data_product_membership(session, admin_user):
+    data_product_membership = DataProductMembershipModel(
+        user_id=admin_user.id,
+        role=DataProductUserRole.MEMBER,
+    )
+    return data_product_membership
+
+
+@pytest.fixture
+def admin_data_product(
```
stijn-janssens, line 197: what is the usecase of this admin data product?
To me the admin data product is just a normal data product that the admin is not the owner / member of, but should still be able to interact with. Am I missing something else?

backend/tests/data_products_datasets/test_router.py
```diff
@@ -26,6 +28,24 @@ def test_request_data_product_link(
         )
         assert link.status_code == 200
 
+    def test_request_data_product_link_by_admin(
```
stijn-janssens, line 31: I feel like this tests the wrong thing, because of the admin dataset and admin product, but I might be wrong.
I would test that the admin user can link any product and dataset together, regardless of ownership

backend/tests/data_products_datasets/test_router.py
```diff
@@ -48,6 +68,35 @@ def test_approve_data_product_link(
         approved_link = self.approve_default_data_product_dataset_link(client, link_id)
         assert approved_link.status_code == 200
 
+    def test_approve_data_product_link_by_admin(
```
stijn-janssens, line 71: I think this tests the use case I described above? But then I don't know what the other use case does :D

stijn-janssens:
Looks good to me overall, but I have some questions regarding some of the unit tests. It's just for my understanding so I'll approve already

stijn-janssens (approved):

PR #47, opened 2024-07-19
FE. Start supporting admin user

andrii-ivaniuk:


frontend/src/pages/data-product/components/data-product-tabs/dataset-tab/dataset-tab.tsx
```diff
@@ -42,7 +42,7 @@ export function DatasetTab({ dataProductId }: Props) {
     const isDataProductOwner = useMemo(() => {
         if (!dataProduct || !user) return false;
 
-        return getIsDataProductOwner(dataProduct, user.id);
+        return getIsDataProductOwner(dataProduct, user.id) || user.is_admin;
```
stijn-janssens, line 45: I would think about the naming of your methods.
Here your method is `isDataProductOwner` and returns both owner or admin.
Above in `frontend/src/components/data-products/data-product-form/data-product-form.component.tsx` the `getIsDataProductOwner` only returns owner.
I would think of a name that suggests you have some form of priviliged access, this can then return both owner check or admin and that method should be used almost everywhere, removing the longer or chain

andrii-ivaniuk: isDataProductOwner is not a method. It is just a variable. Technically, the admin is the owner of all the projects. So, in this case, I believe  the naming is suitable.

andrii-ivaniuk: Different entities use different approaches to get the owner. So, it will be quite difficult to create some general function.

stijn-janssens: alright, makes sense. I hope it does not hurt us in the future

PR #48, opened 2024-07-19
Nginx Configuration with Additional Proxy Headers for Swagger 

alishams8:
…ver Block

- Added `proxy_set_header` directives to the /api and / locations in the main server block to forward headers:
  - `Host` to preserve the original host header.
  - `X-Real-IP` to pass the client’s real IP address.
  - `X-Forwarded-For` to pass the original client’s IP address through proxy servers.
  - `X-Forwarded-Proto` to pass the original protocol (HTTP or HTTPS).

- Introduced a new server block listening on port 5050 for direct backend access:
  - Configured to forward necessary headers similar to the main server block.

These changes improve the transparency and request forward to  access swagger API on port 5050.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -33,6 +33,7 @@ services:
     build: ./nginx
     ports:
       - 8080:80
+      - 5050:5050
     depends_on:
       - backend
       - frontend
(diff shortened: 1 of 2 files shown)


stijn-janssens (approved):
Looks good to me! Thanks for your contribution

PR #49, opened 2024-07-19
feat: Add dataset_type field to datasets model

alishams8:
Datasets come in various types such as tables, documents, webpages, and system logs. This update adds the dataset_type field to the datasets model in the backend, allowing for better classification and handling of different dataset formats.


stijn-janssens:
Hi,
Thanks for your contribution!

We are aware that this is indeed a missing concept and I really like your suggested implementation.
However, the vision we have for the future of Portal is a little different.

We aim to introduce a third concept, next to **data products** and **datasets**, called **data outputs**.
These data outputs will be the link between data products and their technical implementation.
e.g. it can be an s3 path, a table in Glue, a complete database , ...
These data outputs will have types similar to what you implemented here.

Datasets will be a combination of data outputs, grouped together for shareability. Assigning a technical type to these datasets therefor does not make a lot of sense

If you are interested to discuss this concept in more detail, or want to contribute to the development of these data outputs, we warmly invite you to [join us in our Slack channel!](https://join.slack.com/t/dataproductportal/shared_invite/zt-2lo245t3l-bJxcAz1trqrcQNnGARWMDg)

PR #50, opened 2024-07-22
Add APIs to manage settings

andrii-ivaniuk:


backend/app/database/database.py
```diff
@@ -57,5 +57,9 @@ def get_db_session():
     db = SessionLocal()
     try:
         yield db
+        db.commit()
```
stijn-janssens, line 60: I am not sure we should do the commiting and the rollback here?
What is your reasoning behind adding this code? Isn't it safer to always explicitly commit where needed?

backend/app/dependencies.py
```diff
@@ -0,0 +1,12 @@
+from fastapi import Depends, HTTPException, status
```
stijn-janssens, line 1: I think /core/helpers might be a better location for this file. And I would also name it something like admin functionality. Would not group all the dependencies together.

backend/app/environments/enums.py
```diff
@@ -0,0 +1,5 @@
+from enum import StrEnum
+
+
+class PlatformTypes(StrEnum):
```
stijn-janssens, line 4: this should be in the /platforms folder

backend/app/environments/model.py
```diff
@@ -1,17 +1,34 @@
...
     # has Snowflake # AWS, Snowflake, Azure, Databricks
     # Custom specs per type?
 
     # Mvp: name and Iam role with data product name as a template
     # e.g. role-arn: iam-{{DataProductName}}-dev-iam-role
+
+    platforms: Mapped[list["Platform"]] = relationship("Platform", backref="env")
+
+
+class Platform(Base, BaseORM):
```
stijn-janssens, line 28: I would not strongly couple these models and move all platform functionality to /platforms folder

backend/app/environments/router.py
```diff
@@ -1,17 +1,56 @@
...
+from app.environments.schema import CreatePlatform, Environment, Platform
+from app.environments.service import EnvironmentService
 
 router = APIRouter(prefix="/envs", tags=["environments"])
 
 
 @router.get("")
 def get_environments(db: Session = Depends(get_db_session)) -> list[Environment]:
-    return EnvironmentService().get_environments(db)
+    return EnvironmentService(db).get_environments()
```
stijn-janssens, line 17: is there a performance gain with moving the db?
I'm not sure what the best practice way is in this case. Can you still be sure you will always use the latest session? I see no reason why not but want to double check

backend/app/environments/router.py
```diff
@@ -1,17 +1,56 @@
...
+)
+def delete_environment_settings(
+    platform_id: UUID = Path(alias="id"),
+    db: Session = Depends(get_db_session),
+):
+    EnvironmentService(db).delete_settings(platform_id)
+
+
+@router.get("/{environment}/platforms", dependencies=[Depends(only_for_admin)])
+def get_environment_platforms(
```
stijn-janssens, line 51: I would maybe separate the `get all` from the `get_platform_by_name` call. Then you can also use the correct single return type in both routes

backend/app/environments/schema.py
```diff
@@ -1,7 +1,46 @@
...
+from .enums import PlatformTypes
+
 
 class Environment(ORMModel):
     name: str
-    context: str
     is_default: bool = False
+
+
+class S3Config(ORMModel):
```
stijn-janssens, line 16: I would put all of these in a separate folder, similar to what I'm doing in the data output PR

backend/app/environments/schema.py
```diff
@@ -1,7 +1,46 @@
...
+    glue: GlueConfig
+
+
+class CreatePlatform(ORMModel):
+    name: PlatformTypes
+    settings: AWSSettings
+
+    @field_validator("settings", mode="before")
+    @classmethod
+    def parse_settings(cls, v: str | dict) -> dict:
```
stijn-janssens, line 39: this is very nice!

stijn-janssens (commented):
Overall idea is solid, but I think we need to tweak it a little bit.
Also might run into some overlap issues with my PR, but that is something we need to sort out later. This PR should be the starting point

backend/app/database/database.py
andrii-ivaniuk: Why do you need to write it explicitly everywhere if you can manage it in one place here?

The invocation always returns to line #59 when the request is over.

backend/app/dependencies.py
andrii-ivaniuk: Usually, all generic dependencies are kept in a file called `dependencies.py`. It allows you to figure out which dependencies exist in your app and manage them more easily, especially when one dependency depends on another.

backend/app/environments/router.py
andrii-ivaniuk: It was agreed on our call. I can always return a list if it is preferable.

backend/app/environments/schema.py
andrii-ivaniuk: This is an initial implementation and we don't exactly know the final shape, thus I'd keep it as is for now.

backend/app/database/database.py
stijn-janssens: For now I don't think we have a use case yet, but maybe not all db sessions want to commit at closing. You can leave for now and we'll see how it behaves

backend/app/dependencies.py
stijn-janssens: Sure, but all our auth dependencies are in core/auth. So these are not in dependencies.py
Maybe we should have a access, permissions or rights folder? Where we put the dependencies on access level in?
I'm not against a global dependencies.py file, just not sure if this should be in it

backend/app/environments/router.py
stijn-janssens: we agreed that we would specify a path parameter IIRC? Currently it is a body / query parameter and it is not always required.
The frontend is set up with two separate calls for getting everything and getting a single one.
The single call should use a path parameter as discussed. If I said body / query parameter previously then that is my bad

backend/app/dependencies.py
andrii-ivaniuk: It looks more readable to me when I see `dependencies` files in the app root that includes all the global dependencies.

https://fastapi.tiangolo.com/tutorial/bigger-applications/#an-example-file-structure

backend/app/environments/service.py
```diff
@@ -1,12 +1,61 @@
...
+    def delete_settings(self, platform_id: UUID) -> None:
+        platform = self.db.get(PlatformModel, str(platform_id))
+        if platform:
+            platform.delete()
+            self.db.flush()
 
-    def create_environment(self, environment: Environment, db: Session):
-        db.add(EnvironmentModel(**environment.model_dump()))
-        db.commit()
+    def get_environment_platforms(
```
stijn-janssens, line 49: Final remark: I would personally split this method also in two separate ones, removing the extra if statement and making the logic within a method almost trivial

stijn-janssens (approved):

PR #53, opened 2024-07-23
Add first implementation for data outputs as example

stijn-janssens:


backend/app/data_outputs/router.py
```diff
@@ -0,0 +1,39 @@
...
+from app.data_outputs.schema import DataOutput, DataOutputCreate
+from app.data_outputs.service import DataOutputService
+from app.database.database import get_db_session
+
+router = APIRouter(prefix="/data_outputs", tags=["data_outputs"])
+
+
+@router.get("")
+def get_data_outputs(db: Session = Depends(get_db_session)) -> list[DataOutput]:
+    return DataOutputService().get_data_outputs(db)
```
andrii-ivaniuk, line 15:
```suggestion
    return DataOutputService(db).get_data_outputs()
```
Since you already instantiate an object I suggest passing in 'db' to `init` method. In such a case if you will need to use db in multiple methods you can use self.db instead of passing it as an argument to another method

backend/app/data_outputs/router.py
```diff
@@ -0,0 +1,97 @@
...
+    authenticated_user: User = Depends(get_authenticated_user),
+    db: Session = Depends(get_db_session),
+):
+    return DataOutputService().link_dataset_to_data_output(
+        id, dataset_id, authenticated_user, db
+    )
+
+
+@router.delete(
+    "/{id}/dataset/{dataset_id}",
```
andrii-ivaniuk, line 73:
```suggestion
    "/{data_output_id}/dataset/{dataset_id}",
```
Can we use more specific names instead of id?

backend/app/data_outputs_datasets/service.py
```diff
@@ -0,0 +1,83 @@
...
+            and not authenticated_user.is_admin
+        ):
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail="Only dataset owners can execute this action",
+            )
+        current_link.status = DataOutputDatasetLinkStatus.DENIED
+        current_link.denied_by = authenticated_user
+        current_link.denied_on = datetime.now(tz=pytz.utc)
+        db.commit()
```
andrii-ivaniuk, line 63:
```suggestion
        db.flush()
```

db.commit() will be executed in db session getter generator

PR #64, opened 2024-07-30
Update poetry file to add FactoryBoy

stijn-janssens:


backend/pyproject.toml
```diff
@@ -28,6 +28,7 @@ pytz = "^2024.1"
 boto3 = "^1.34.90"
 jinja2 = "^3.1.3"
 pydantic-settings = "^2.3.4"
+factory-boy = "^3.3.0"
```
andrii-ivaniuk, line 31: Why did you add it here? `factory-boy` is added under `test` group, since it is required only for testing

stijn-janssens: Sorry, forgot about that. Will change this PR to updating the README then to install this locally

andrii-ivaniuk (approved):

PR #65, opened 2024-07-30
Add settings related tables

andrii-ivaniuk:


andrii-ivaniuk:
Details:
Screenshot 2024-07-30 at 13 51 35
Screenshot 2024-07-30 at 13 51 53

backend/app/database/alembic/versions/2024_07_22_1630-2e48ad16e9f7_add_platforms_settings_table.py
```diff
@@ -22,6 +22,9 @@
 
 
 def upgrade() -> None:
+    class PlatformTypes(StrEnum):
```
stijn-janssens, line 25: Shouldn't this be part of the source code instead of an Alembic migration?

stijn-janssens: edit: it is probably there to be able to remove the old enum? weird syntax but fine for now

backend/app/environments/model.py
```diff
@@ -1,35 +1,12 @@
...
+from sqlalchemy import Boolean, Column, String
 
 from app.database.database import Base
 from app.shared.model import BaseORM
 
-from .enums import PlatformTypes
-
 
 class Environment(Base, BaseORM):
     __tablename__ = "environments"
```
stijn-janssens, line 9: I would add id to environment table to use instead of name as foreign key

backend/app/platforms/router.py
```diff
@@ -0,0 +1,60 @@
...
+                    "example": {"detail": "Service configuration not found"}
+                }
+            },
+        }
+    },
+)
+def get_service_config(
+    platform_id: UUID, service_id: UUID, db: Session = Depends(get_db_session)
+) -> PlatformServiceConfigSchema:
+    if not (
```
stijn-janssens, line 51: this method should not contain the logic. The 404 functionality should happen inside of the service

backend/app/platforms/schemas.py
```diff
@@ -0,0 +1,30 @@
...
+from pydantic import field_validator
+
+from app.shared.schema import ORMModel
+
+
+class Identifiers(ORMModel):
+    identifiers: list[str]
+
+
+class PlatformServiceConfigSchema(ORMModel):
```
stijn-janssens, line 13: This is a temp config I pressume? You will probably create S3 specific and Glue specific schemas in the future?

backend/app/database/alembic/versions/2024_07_22_1630-2e48ad16e9f7_add_platforms_settings_table.py
andrii-ivaniuk: Yes, this is required to remove redundant enums.py

backend/app/environments/model.py
andrii-ivaniuk: I planned to add ID but it affects `permissionsets` functionality. I propose to add it in a separate dedicated PR

backend/app/platforms/router.py
andrii-ivaniuk: I can't entirely agree. I think all fastapi-related logic should be part of a router and not of a service. What if I need to reuse the service in another place and I want to check that a record doesn't exist? Why should the service return HTTPException?

backend/app/platforms/schemas.py
andrii-ivaniuk: No, this is the schema used to return generic config.

andrii-ivaniuk: Identifiers might be converted to something specific later

backend/app/platforms/router.py
stijn-janssens: This should never be the case right? The 404 happens because something unexpected occurs. It should not be thrown in case we deliberately check if something exists. Then you should use another method. The get_service_config always expects the service config to be there, so it raises an exception if it is not.

backend/app/platforms/schemas.py
stijn-janssens: I think we are talking about the same thing :D
Fine for now as a first implementation

stijn-janssens (approved):

PR #67, opened 2024-07-31
Managing Environment platform service config

andrii-ivaniuk:


backend/app/environments/service.py
```diff
@@ -41,12 +41,12 @@ def create_config(
             PlatformServiceConfig.service_id == service_id,
         )
 
-        pltf_srvc_config = self.db.scalar(stmt)
-        if not pltf_srvc_config:
+        pltfrm_srvc_config = self.db.scalar(stmt)
```
stijn-janssens, line 44: I'm not a huge fan of abbreviating these vars. Would just use `platform_service_config` in this case

backend/tests/environments/test_router.py
```diff
@@ -25,8 +33,8 @@ def test_get_environments(self, client):
         assert len(environments) == 1
         assert environments[0]["name"] == env_obj.name
 
+    @pytest.mark.usefixtures("admin")
```
stijn-janssens, line 36: why do you use fixtures here instead of the factory boy?

backend/tests/environments/test_router.py
```diff
@@ -42,3 +50,94 @@ def test_create_environment_with_repeated_name(self, client):
     def create_environment(client, environment):
         response = client.post(ENDPOINT, json=environment)
         return response
+
+    def test_get_environment_platform_service_config_forbidden(self, client):
```
stijn-janssens, line 54: I don't think this endpoint should be admin protected. Only the post, put and delete should be protected.
Getting this config will be necessary for the creation of data outputs

stijn-janssens (approved):

backend/app/environments/service.py
andrii-ivaniuk: ok

backend/tests/environments/test_router.py
andrii-ivaniuk: Fixture uses FactoryBoy :)

backend/tests/environments/test_router.py
andrii-ivaniuk: I think you will not use this endpoint outside of the settings page.  Data output will handle it on BE side. So, let's keep it protected for now.

stijn-janssens: Alright, easy enough to fix should we still need it later indeed

PR #84, opened 2024-09-09
Improve endpoint security with data product role dependency

stijn-janssens:
Some endpoints were completely unprotected. They did not check whether or not you were a member / owner for a product.
E.g. the request credentials from AWS endpoint.
This is a big security risk.

Now all endpoints from the data products route are at least secured. The rest can be refactored once this method is approved and tested

### DIFF
--- a/backend/app/data_products/router.py
+++ b/backend/app/data_products/router.py
@@ -4,14 +4,16 @@ from fastapi import APIRouter, Depends
 from sqlalchemy.orm import Session
 
 from app.core.auth.auth import get_authenticated_user
+from app.data_product_memberships.enums import DataProductUserRole
 from app.data_products.schema import (
+    DataProductAboutUpdate,
     DataProductCreate,
     DataProductUpdate,
-    DataProductAboutUpdate,
 )
 from app.data_products.schema_get import DataProductGet, DataProductsGet
 from app.data_products.service import DataProductService
 from app.database.database import get_db_session
+from app.dependencies import OnlyProductRoles
 from app.users.schema import User
 
 router = APIRouter(prefix="/data_products", tags=["data_products"])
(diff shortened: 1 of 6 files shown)


backend/app/dependencies.py
```diff
@@ -10,3 +18,49 @@ async def only_for_admin(authenticated_user: User = Depends(get_authenticated_us
...
+    ) -> None:
+        if id:
+            try:
+                data_product = db.get_one(DataProductModel, id)
+            except NoResultFound:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND, detail="product not founds"
+                )
+        elif data_product_name:
+            data_product = db.query(DataProductModel).get_one(
```
andrii-ivaniuk, line 46: I thought we decided to use a new approach to query data (SQLAlchemy 2.0), isn't it?

backend/app/data_products/router.py
```diff
@@ -189,11 +201,17 @@ def get_signin_url(
...
+    dependencies=[Depends(OnlyProductRoles([DataProductUserRole.MEMBER]))],
+)
 def get_conveyor_notebook_url(id: UUID, db: Session = Depends(get_db_session)) -> str:
     return DataProductService().get_conveyor_notebook_url(id, db)
 
 
-@router.get("/{id}/conveyor_ide_url")
+@router.get(
+    "/{id}/conveyor_ide_url",
+    dependencies=[Depends(OnlyProductRoles([DataProductUserRole.MEMBER]))],
```
andrii-ivaniuk, line 214: It's a bit confusing that you restrict access to a member but allow access to a member and owner. 
I'd suggest providing the role(s) only if you want to restrict access for the specific role, otherwise by default it allow access to member and owner.

stijn-janssens: Great suggestion

backend/app/dependencies.py
stijn-janssens: Forgot about it. To me it also feels less intuitive, but it probably just takes some getting used to

andrii-ivaniuk (approved):

PR #97, opened 2024-09-19
Added environment platform config

andrii-ivaniuk:


backend/app/environments/schema.py
```diff
@@ -18,27 +17,43 @@ class GetEnvironment(Environment):
...
+    bucket_name: str
     arn: str
-    kms: str
+    kms_key: str
+    is_default: bool
 
 
-class Config(RootModel[Dict[str, _AWSS3Config]]):
-    pass
+class _AWSGlueConfig(ORMModel):
```
stijn-janssens, line 27: I would put all of these configs in separate files

stijn-janssens: similar to my approach in data outputs branch

backend/app/environments/schema.py
```diff
@@ -18,27 +17,43 @@ class GetEnvironment(Environment):
...
+    identifier: str
+    database_name: str
+    bucket_name: str
+    s3_path: str
 
 
 class EnvPlatformServiceConfig(ORMModel):
-    config: Config
+    config: list[_AWSS3Config | _AWSGlueConfig]
```
stijn-janssens, line 35: will we save this in the database as a single entry per environment and platform? Or will we also split per service?
Here you chose to do it as a single entry, why did you choose that and not the other option?

backend/app/environments/schema.py
```diff
@@ -18,27 +17,43 @@ class GetEnvironment(Environment):
...
-class CreateConfigSchema(EnvPlatformServiceConfig):
-    platform_id: UUID
-    service_id: UUID
+class _AWSConfig(ORMModel):
+    account_id: str
+    region: str
+    can_read_from: list[str]
+
+
+class EnvPlatformConfig(ORMModel):
```
stijn-janssens, line 51: I would also split up the platform and platform service configs. This feels like a god file now

andrii-ivaniuk: Why do you mean under `platform` and `platform service configs`?

backend/app/environments/schema.py
andrii-ivaniuk: It is env,  platform, and service granularity. 

class EnvPlatformServiceConfig(Base, BaseORM):
    __tablename__ = "env_platform_service_configs"

    id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
    environment_id: Mapped[UUID] = mapped_column(ForeignKey("environments.id"))
    platform_id: Mapped[UUID] = mapped_column(ForeignKey("platforms.id"))
    service_id: Mapped[UUID] = mapped_column(ForeignKey("platform_services.id"))
    config = Column(String)

backend/app/environments/schema.py
andrii-ivaniuk: How do you suggest splitting it? Based on which entity?

stijn-janssens: Have a file per service. I have S3DataOutput and GlueDataOutput separated, but inheriting from a generic DataOutput.
Same would go for these configs.

backend/app/environments/schema.py
andrii-ivaniuk: It is a list because it can be multiple identifiers fo e.g. s3

backend/app/environments/schema.py
andrii-ivaniuk: Fixed

stijn-janssens (approved):

stijn-janssens (commented):
I think there are still things wrong with the mapping between your Pydantic schemas and the ORM model in SQL Alchemy. Up till now, these mappings match perfectly.
Why did you not include things like platform_id etc in the pydantic schemas?

andrii-ivaniuk:
> I think there are still things wrong with the mapping between your Pydantic schemas and the ORM model in SQL Alchemy. Up till now, these mappings match perfectly. Why did you not include things like platform_id etc in the pydantic schemas?

Do we need it? If yes, I'll add it.

andrii-ivaniuk:
> I think there are still things wrong with the mapping between your Pydantic schemas and the ORM model in SQL Alchemy. Up till now, these mappings match perfectly. Why did you not include things like platform_id etc in the pydantic schemas?

Added metadata about environment, platform, and service. Added endpoint that returns all the configs for a specific environment.

PR #359, opened 2024-10-30
Feature/databricks env config

JesusGarciaRamirez:


backend/app/environment_platform_service_configurations/schemas/databricks_schema.py
```diff
@@ -0,0 +1,5 @@
+from app.shared.schema import ORMModel
+
+
+class DatabricksConfig(ORMModel):
+    pass_value: str
```
stijn-janssens, line 5: have you tried this with an empty config? Would be cleaner

integrations/.DS_Store
stijn-janssens, line 1: Can you remove all these .DS_Store files from Git please :)

stijn-janssens (changes requested):

backend/app/environment_platform_service_configurations/schemas/databricks_schema.py
JesusGarciaRamirez: Yup, it works!

backend/app/environment_platform_service_configurations/schemas/databricks_schema.py
stijn-janssens, line 1: I have not tested it, but I am afraid that this will now allow any kind of configuration to pass through as valid DatabricksConfiguration. Because we essentially allow just any kind of class.

It seems this is the pydantic way to do it, but I also have not tested it.

```
class Empty(BaseModel):
    ...

    model_config = ConfigDict(extra="forbid")
```

JesusGarciaRamirez: The pydantic way to do it seems to work (both the docker compose and the portal_read_config.py script)

stijn-janssens (approved):

PR #640, opened 2025-01-23
Feature/datasets lifecycles

stijn-janssens:
Add lifecycles to datasets
Various improvements in dataset UI after lifecycles
Fixes auto-rerendering when editing lfiecycles

### DIFF
--- a/backend/app/database/alembic/versions/2025_01_23_1117-4c49c19ee972_.py
+++ b/backend/app/database/alembic/versions/2025_01_23_1117-4c49c19ee972_.py
@@ -1,0 +1,23 @@
+"""empty message
+
+Revision ID: 4c49c19ee972
+Revises: ca11eb0735ed, 0402aaf51be7
+Create Date: 2025-01-23 11:17:39.198274
+
+"""
+
+from typing import Sequence, Union
+
+# revision identifiers, used by Alembic.
+revision: str = "4c49c19ee972"
+down_revision: Union[tuple[str, str], None] = ("ca11eb0735ed", "0402aaf51be7")
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    pass
+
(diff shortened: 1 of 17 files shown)


backend/app/database/alembic/versions/2025_01_23_1117-4c49c19ee972_.py
vlieven, line 1: Why are there two migration files that have empty actions?

stijn-janssens: when you create a new revision from alembic it needs to have a single head. Because we sometimes create different revisions from the same head in different branches we can diverge into 2 heads when merging in main. When that happens you just have to merge those 2 heads through alembic, creating a new single head. That is done with such an "empty migration". If you check the revisions you'll see it does a merge

vlieven: Huh ok, I did not know that this was a thing. Thanks for clarifying!

frontend/src/pages/datasets/components/datasets-table/datasets-table-columns.tsx
```diff
@@ -49,17 +59,21 @@ export const getDatasetTableColumns = ({ t, datasets }: { t: TFunction, datasets
...
-        //     title: t('Status'),
-        //     dataIndex: 'status',
-        //     render: (status) => {
-        //         return <TableStatusTag status={status} />;
-        //     },
-        //     ...new FilterSettings(datasets, ds => getStatusLabel(ds.status)),
-        //     sorter: sorter.stringSorter(ds => getStatusLabel(ds.status)),
-        //     width: "12%"
-        // },
+        //TODO Fix
```
vlieven, line 62: Is this TODO still relevant?
If yes: best to specify what still nees to be fixed
If no: best to remove comment

vlieven (changes requested):

Issue #627, opened 2025-01-20
OpenAPI spec issues with OnlyWithProductAccess


stijn-janssens:
## Describe the bug
API routes that depend on OnlyWithProductAccess will mess with the OpenAPI spec

### Steps to reproduce
Look at a method that depends on OnlyWithProductAccess. e.g. `link_dataset_to_data_product`

### Expected behavior
The id should be of type string, format uuid only. But instead also has the option to be nullable
There is an extra data_product_id in the schema. 

### Environment
NA

### Additional context
OnlyWithProductAccess has been coded generically to fit with multiple routes, however this messes up the openAPI spec, and the ability to generate clients from this spec. We should fix this issue asap

stijn-janssens:
Bug has been fixed for python openapi generator (openapi-python-client)
Go generators still complain about null values (oapi-codegen)

PR #654, opened 2025-01-28
Fix/dependency and openapi schema issues

stijn-janssens:
Fixes partially #627 Go codegen is still complaining about null types not existing. Python codegen now works


backend/app/data_products/schema_base_get.py
```diff
@@ -10,7 +10,7 @@ class BaseDataProductGet(ORMModel):
     id: UUID
     name: str
     description: str
-    about: Optional[str]
+    about: Optional[str] = ""
```
vlieven, line 13: I'm not sure this pattern is correct. Typically, you would say `about: Optional[str] = None`. If you let the value default to `""`, then there is no reason to label this type as `Optional` anymore (`Optional[str] == str | None`).

So I believe this should either be `about: Optional[str] = None` or `about: str = ""`.

vlieven (changes requested):

backend/app/data_products/schema_base_get.py
stijn-janssens: Good catch!

vlieven (approved):

PR #658, opened 2025-01-29
Feature/refactor settings

stijn-janssens:
- Allow deletes of lifecycles (resets back to default if in use)
- Allow empty values as default values in custom settings, used to be a required value so could not be empty.
- Add customizable settings not only for data products, but datasets as well


backend/app/dependencies.py
```diff
@@ -29,7 +29,7 @@ async def only_for_admin(authenticated_user: User = Depends(get_authenticated_us
         )
 
 
-async def only_dataset_owners(
+def only_dataset_owners(
```
vlieven, line 32: Why is this no longer async?

stijn-janssens: It should never have been async imo. Previously it was used by fastapi Dependencies, which use async on their own, regardless of whether or not you define the function as async, similar to how their routing functions work.
Now I also want to use it without the FastAPI Dependency, as a "normal" call, and it needs to return synchronously

vlieven (approved):

vlieven (approved):

PR #664, opened 2025-01-30
Part 1 of RBAC

vlieven:
This PR should more or less put the foundations for the UI in place to manage roles.
Currently the route is navigable, but not linked from anywhere. If that's better I can also temporarily remove the route.

### Preview

Screenshot 2025-01-30 at 17 05 57

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ data sources, and sharing concepts, ensuring control and oversight for business 
 Our goal is to bridge the gap between data governance, data platforms and data catalogs and provide a 360 view of all
 ongoing data initiatives that is easy to understand by everybody.
 
-To read more about it, please checkout our [announcement blogpost](https://conveyordata.com/portal-introducing-new-open-source-data-product-portal?utm_source=github-portal-readme&utm_medium=referral)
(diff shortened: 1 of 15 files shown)


stijn-janssens:
Overall: 👌 , nice UI. Clear and easy to use imo

Personal preference, feel free to disagree. I would put a disclaimer on top of the page that it is work in progress and that nothing actually happens? I am fine with the route not being in the navigation for now, you might as well add it however since it is admin only?
Some rather minor comments, don't need to be fixed for a "starter point PR":
- In other places in the UI the tooltips are question marks instead of `i`.
- Not sure if you are using a different component or what explains it, but the tabs in other places in the UI (e.g. detailed Data Product page) show animation when changed.
- The tabs should also be routable (see detailed Data Product page)

vlieven:
I've updated the page based on the feedback received today.
Regarding that tabs should be routable: I agree that this makes sense, but I think this functionality should be extracted into a custom hook. The implementation that exists now in the Data Product details page f.e. does not account for the case where an invalid key is provided (should fall back to default, but now does not render anything). Let's tackle this in a follow-up PR.

stijn-janssens:
Agreed! (And good catch on the invalid keys)

As discussed, the table could use some improved styling. e.g. smaller rows etc. But I would look at that only when we have a better view of needed permissions.

stijn-janssens (approved):

PR #683, opened 2025-02-04
Feature/webhooks

### LINKED ISSUE #615: Allow defining a hook on the "Add user" action
## Summary
Allow triggering an integration on the event of adding a user to a data product or dataset.

### Motivation
Allows to sync user information back to external systems

stijn-janssens:
First implementation of webhooks:

- Adds middleware to be triggered on all POST, PUTs and DELETE
- sends the path of the url and the response for POSTs (which is the ID of the newly created object)
- Should be enough information to allow the webhook to fetch all necessary extra information if permitted.

Closes #615

### DIFF
--- a/backend/app/core/webhooks/webhook.py
+++ b/backend/app/core/webhooks/webhook.py
@@ -1,0 +1,23 @@
+import json
+
+import httpx
+
+from app.core.logging.logger import logger
+from app.settings import settings
+
+
+async def call_webhook(content, method, url, query, body=None):
+    webhook_url = settings.WEBHOOK_URL
+    async with httpx.AsyncClient() as client:
+        content = {
+            "method": method,
+            "url": url,
+            "query": query,
+            # "body": body if body else {},
+            # Body is behaving weirdly. Not sure if needed here?
+            "response": content,
+        }
+        logger.info("Webhook triggered with content: %s", content)
+        resp = await client.post(webhook_url, json=json.dumps(content))
+        if (resp.status_code != 200) or (resp.json().get("status") != "ok"):
+            logger.warning(f"Failed to send notification to {webhook_url}")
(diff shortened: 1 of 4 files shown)


backend/app/main.py
```diff
@@ -79,6 +82,28 @@ async def log_middleware(request: Request, call_next):
...
+@app.middleware("http")
+async def send_response_to_webhook(request: Request, call_next):
+    response = await call_next(request)
+    # Gets are not logged
+    if request.method in ["POST", "PUT", "DELETE"] and not request.url.path.startswith(
+        "/api/auth/"
+    ):
+        response_body = [chunk async for chunk in response.body_iterator]
+        response.body_iterator = iterate_in_threadpool(iter(response_body))
+        body = (b"".join(response_body)).decode()
```
vlieven, line 94: Possible problem here: you're fetching the entire body and uploading it all at once in the response. For pathological cases, this might cause the memory usage of the API to spike. I don't know if it is possible to fix this though, as you're introducing a second reader, so the options I see are creating a copy (bad), or couple the readers (probably worse).

backend/app/core/webhooks/webhook.py
```diff
@@ -0,0 +1,26 @@
...
+
+import httpx
+
+from app.core.logging.logger import logger
+from app.settings import settings
+
+
+async def call_webhook(content, method, url, query, status_code):
+    webhook_url = settings.WEBHOOK_URL
+    if webhook_url:
```
vlieven, line 11: This check should be in the `send_response_to_webhook` function (too?). Now you're creating an empty task on the API for each request. This will introduce some overhead (I don't know how large this will be) even for installations that don't use the webhook function

backend/app/main.py
stijn-janssens: Perhaps a fix could be to only fetch this information on POST calls, which is also the only time I put the body in the webhook content. That way we control the information that will be fetched, right? It should only be IDs I think

backend/app/core/webhooks/webhook.py
stijn-janssens: Agree!

vlieven (approved):

Issue #605, opened 2025-01-14
Customisable Roles & Responsibilities


vlieven:
## Value

Tailor Portal behavior towards future Luminus process requirements

### Luminus

Luminus would like that this flow can be implemented through portal

```mermaid
sequenceDiagram
    title Data Product Lifecycle Process

    participant DPA as Admin
    participant DO as Domain Owner
    participant DPO as Data Product Owner
    participant DPM as Data product team member
    participant SDA as Solution Architect
    participant DS as Data Steward

    %% Data Domain Creation
    Note over DPA: Define Data Domains across the organisation
    DPA->>DPA: create("new Data Domain")
    DPA->>DPA: assign("Domain owner")

    %% Purpose Creation : Purpose can also be created by DPM but needs DPO approval extra
    Note over DPA, SDA: Purpose Creation
    DPO->>DPA: requestApproval("Tooling and Capabilities")
    DPA-->>DPO: approve("Tooling and Capabilities")
    %% Todo: add additional purpose configurations
    DPO->>DPO: set("extra purpose properties")
    DPO->>SDA: requestApproval("Purpose")
    SDA-->>DPO: approve("Purpose creation")

    %% Team Setup:
    Note over DPO, SDA: Define Purpose Engineers
    DPO->>DPM: assign("Team Members")
    DPM->>DPO: requestApproval("Purpose membership & role")
    DPO-->>DPM: approve("Purpose membership & role")

    %% Data Product Definition
    Note over DPO, SDA: Data Product Definition
    DPM->>DPM: select("Data Business capabilities")
    DPM->>DPM: select("Data domain")
    DPM->>DPM: optionalLink("Purpose (if exists)")
    DPM->>DPM: select("Data product type [SADP/CADP]")
    DPM->>DPM: define("Snowflake schemas")
    DPM->>SDA: request("Create Data Product")
    SDA-->>DPM: approve("Data Product creation")
    DPM->>DPO: request("Create Data Product")
    DPO-->>DPM: approve("Data Product creation")
  
    Note over DPO, SDA: Link data product to purpose (Data Product First)
    DPM->>DPM: link("Purpose to Data Product")
    DPM->>SDA: requestApproval("link Purpose To Data Product")
    SDA-->>DPM: approval("link Purpose To Data Product")
    DPM->>DPO: requestApproval("link Purpose To Data Product")
    DPO-->>DPM: approval("link Purpose To Data Product")

    %% Publish table definition
[...]

PR #692, opened 2025-02-06
Basic integration for casbin

vlieven:
Add a new `authz` module to enable RBAC in the future.
The module is not wired in yet, so functionally nothing should change (yet). 

In support of #605


backend/app/core/authz/authorization.py
```diff
@@ -0,0 +1,72 @@
...
+from app.users.schema import User
+from app.utils.singleton import Singleton
+
+from .actions import AuthorizationAction
+
+
+class Authorization(metaclass=Singleton):
+
+    def __init__(self, enforcer: AsyncEnforcer = None) -> None:
+        self.enforcer: AsyncEnforcer = cast(AsyncEnforcer, enforcer)
```
stijn-janssens, line 20: why do you still need to cast here? `_construct_enforcer` seems to return the correct type `AsyncEnforcer`?

backend/app/core/authz/authorization.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @classmethod
+    async def initialize(cls) -> "Authorization":
+        model_location = Path(__file__).parent / "rbac_model.conf"
```
stijn-janssens, line 24: Do we envision a scenario where we would ever want to allow to pass a custom config file here? In that case this setting needs to be more flexible

stijn-janssens (commented):
I assume you tested this code already, or is it just the scaffolding and will you tweak this if things don't work? Asking because I don't see any calls where authorization actually happens in this PR.

stijn-janssens:
In general, very nice code. Quite a lot of magic but I think I get the gist 👌

backend/app/core/authz/authorization.py
vlieven: Good question. At this point in time, I think this is an implementation detail on our side, and should be excluded from user config.
The configuration dictates how casbin works. It's not super extensive/complicated, but it does require reading the documentation thoroughly, so you understand how it works.

The rest of our authorization flows will be designed around this chosen configuration, so if you allow this to be changed, the chances of just breaking the authorization flow will be very high.

backend/app/core/authz/authorization.py
vlieven: It's an artifact of how the class is constructed (Singleton + async initializer). Ideally, you are able to call the `initialize` function inside `__init__`, but because of mixing sync and async, this is impossible.

So as workaround, __init__ needs to be called with the enforcer object in the first call (done in the FastAPI lifecycle), next calls don't need to provide something to __init__, as they won't be ran, the singleton is created at that moment and will be used instead. That does require providing a default value, which then leads to the type of `enforcer` actually becoming `AsyncEnforcer | None`.

Writing all of this down, the flow would probably be a whole lot neater by using a sync enforcer. Once the rest is wired, we can check if there is actually a benefit to using the async enforcer, and if not, we can switch it for the synchronous version?

vlieven:
> I assume you tested this code already, or is it just the scaffolding and will you tweak this if things don't work? Asking because I don't see any calls where authorization actually happens in this PR.

Yes, I've tested this. The intended usage is to use `Authorization.enforce` as a dependency in the `@router` annotations. This works, but the role creation / management needs to be in place before we can actually start wiring this in.

backend/app/core/authz/authorization.py
stijn-janssens: I can't really see how async would work in the case of authorizing route access. Unless perhaps the response can be calculated simultaneously/upfront when the authorization is checked and therefor you gain some time on that front?

If we end up with large role sets perhaps async becomes a must. I am fine with having complex logic in this single spot ;)

stijn-janssens (approved):

PR #719, opened 2025-02-12
Feature/rollup tags

stijn-janssens:
Tags will be rolled up.

Data Output => Dataset => Data Product
Currently there is no way to *not* roll up a tag. e.g. If a data output is tagged with PII that means all datasets using that DO will also be tagged as PII and so will all the Data Products that read the dataset as input.
If the latter Data Product process the PII in a way to pseudoanonimize it, then the outgoing data output should simply not be tagged as PII.
It still is a highly discipline dependent way of governance, but at least automates part of it.

Feedback on this approach very much welcome.


backend/app/data_products/service.py
```diff
@@ -79,10 +79,27 @@ def get_data_product(self, id: UUID, db: Session) -> DataProductGet:
             .first()
         )
 
+        rolled_up_tags = []
```
mathias-bongaerts, line 82: The code can be rewritten to only use a `set` by making the `Tag` schema class hashable, this can be done by adding to the `Tag` schema class:
```python
class Config:
  frozen = True
```

It should then suffice to only write:
```python
rolled_up_tags = set()
```

backend/app/data_products/service.py
```diff
@@ -79,10 +79,27 @@ def get_data_product(self, id: UUID, db: Session) -> DataProductGet:
...
+        unique_tag_values = set()
+
         if not data_product:
             raise HTTPException(
                 status_code=status.HTTP_404_NOT_FOUND, detail="Data Product not found"
             )
+
```
mathias-bongaerts, line 89: Code can then be rewritten as:

```python
for link in data_product.dataset_links:
    rolled_up_tags.update(link.dataset.tags)
    for output_link in link.dataset.data_output_links:
        rolled_up_tags.update(output_link.data_output.tags)
```

backend/app/datasets/service.py
```diff
@@ -36,6 +36,9 @@ def get_dataset(self, id: UUID, db: Session) -> DatasetGet:
             ],
         )
 
+        rolled_up_tags = []
```
mathias-bongaerts, line 39: Code can similarly be rewritten

mathias-bongaerts (approved):
The suggested approach is fine and makes sense to me.

The comments on the code might help to make the code more concise, but are no necessity. No further comments on the code.

mathias-bongaerts (approved):

mathias-bongaerts (approved):

frontend/src/pages/data-product/components/data-product-description/data-product-description.tsx
```diff
@@ -34,7 +34,7 @@ export function DataProductDescription({ lifecycle, type, description, businessA
                 </Space>
                 <Flex>
                     {tags.map((tag) => (
-                        <Tag color="success">{tag.value}</Tag>
+                        <Tag color={tag.rolled_up ? 'red' : 'success'}>{tag.value}</Tag>
```
vlieven, line 37: I don't think this should block the PR, but coloring the tag red when it's a rolled-up tag indicates a failure or problem of some kind. You could also add some sort of icon to indicate a rolled up tag. But we can also redesign this later.

vlieven (approved):

frontend/src/pages/data-product/components/data-product-description/data-product-description.tsx
stijn-janssens: 100% agree. I have not yet settled on something. Just wanted to make it stand out as different. This will for sure be a feature that gets tweaked in production :D

PR #733, opened 2025-02-17
Move from soft deletes to hard deletes, solve cascading issue with datasets

stijn-janssens:
Upon deletion of a dataset several cascades were not happening correctly
- dataset - data product link was not cascaded
- dataset - data output link was not cascaded
These have been fixed

We also removed the soft delete functionality completely. For now there is no reason to offer it as
- the links between objects are always hard deleted
- there is no way of returning a soft deleted item except for going in the database
- any infra integrations (and corresponding data e.g. Glue Tables) are always hard deleted


vlieven (approved):
LGTM

PR #735, opened 2025-02-17
Add glue suffix in detailed and overview

stijn-janssens:


frontend/src/components/data-outputs/data-output-technical-info/data-output-table-glue-columns.tsx
```diff
@@ -28,7 +28,7 @@ export const getGlueTechnicalInformationColumns = ({ t }: Props): TableColumnsTy
             render: (_, { data_output }) => {
                 const configuration: GlueDataOutputContract = data_output.configuration as GlueDataOutputContract;
                 // TODO figure out how to use product aligned databases here and get their ARN?
-                return configuration.database;
+                return configuration.database + '__' + configuration.database_suffix;
```
vlieven, line 31: Maybe it's better to use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead of summation here?

```suggestion
                return `${configuration.database}__${configuration.database_suffix}`;
```

frontend/src/components/data-outputs/data-output-technical-info/data-output-table-snowflake-columns.tsx
```diff
@@ -29,7 +29,7 @@ export const getSnowflakeTechnicalInformationColumns = ({ t }: Props): TableColu
                 const configuration: SnowflakeDataOutputContract =
                     data_output.configuration as SnowflakeDataOutputContract;
                 // TODO figure out how to use product aligned databases here and get their ARN?
-                return configuration.schema + '__' + configuration.schema_suffix;
+                return configuration.database + '__' + configuration.schema;
```
vlieven, line 32: Same here

```suggestion
                return `${configuration.database}__${configuration.schema}`;
```

vlieven (dismissed):

PR #766, opened 2025-02-24
Complete refactor of any occurence of Business Area. Renamed to Domain.

tuuuuur:
Business area -> domain
Business areas -> domains

This is the main logic used in this refactor, reflecting both frontend & backend code.


backend/app/database/alembic/versions/4c40a4ff5a7f_initialize_database.py
mathias-bongaerts, line 1: Rather than editing the original database setup script, a new migration script should be created.

mathias-bongaerts (changes requested):
Changes look good, almost done:
-  Still need to add a database migation script
- Run pre-commit hooks

backend/app/database/alembic/versions/4c40a4ff5a7f_initialize_database.py
tuuuuur: The original database script has been restored, and a new migration has been added.

stijn-janssens (approved):

PR #786, opened 2025-02-28
Chore/missing unit tests

stijn-janssens:
Add missing unit tests for all API calls
Add extra Github Action that makes sure that all fastapi routes are covered by at least one unit test

### DIFF
--- a/backend/app/core/auth/device_flows/service.py
+++ b/backend/app/core/auth/device_flows/service.py
@@ -18,7 +18,7 @@ from app.core.auth.device_flows.exceptions import (
 )
 from app.core.auth.device_flows.model import DeviceFlow as DeviceFlowModel
 from app.core.auth.device_flows.schema import DeviceFlow, DeviceFlowStatus
-from app.core.auth.jwt import oidc
+from app.core.auth.jwt import get_oidc, oidc
 from app.core.helpers.templates import render_html_template
 from app.core.logging.logger import logger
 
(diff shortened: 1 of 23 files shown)


backend/app/core/auth/device_flows/service.py
```diff
@@ -18,7 +18,7 @@
 )
 from app.core.auth.device_flows.model import DeviceFlow as DeviceFlowModel
 from app.core.auth.device_flows.schema import DeviceFlow, DeviceFlowStatus
-from app.core.auth.jwt import oidc
+from app.core.auth.jwt import get_oidc, oidc
```
vlieven, line 21: Should the `oidc` import not be dropped if you've changed everything to a `get_oidc()` call?

backend/tests/auth/test_device_router.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_get_device_token(self, client):
+        response = client.post(f"{ENDPOINT}/device_token?client_id=test")
+        assert response.status_code == 200
+        assert response.json()["oidc_redirect_uri"] == "http://test-redirect-uri"
+        assert response.json()["status"] == "authorization_pending"
+
+    def test_get_jwt_token(self, client):
+        response = client.post(f"{ENDPOINT}/device_token?client_id=test")
+        device_code = response.json()["device_code"]
+        sleep(5)
```
vlieven, line 22: Why is a sleep needed in the test? This will just slow down the test, no?

vlieven (changes requested):

backend/tests/auth/test_device_router.py
stijn-janssens: There is a built-in SlowdownException in the device flow logic, to stop people from non-stop polling. Without the sleep, the instant check throws this exception instead of executing actual logic. But yes, this does slow everything down :(

vlieven, line 22:
```suggestion
        sleep(5)  # Avoids hitting the SlowdownException
```

vlieven (approved):

tuuuuur (approved):

PR #792, opened 2025-03-03
[frontend] Add linting rule for sorting imports

vlieven:
This enforces consistency in our `.ts[x]` imports.

Uses the [simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort) plugin, instead of the builtin [sort-imports](https://eslint.org/docs/latest/rules/sort-imports) rule, as the former is `--fix`-able, while the latter is not.


stijn-janssens (approved):
❤️

PR #806, opened 2025-03-04
Settings page restyling

mathias-bongaerts:
## Changes
- Changed the sidebar to only have a single entry for the settings (rather than a menu item with a single child)
- Split up the settings page into multiple tabs for different categories
- No more inline editing of settings
- Improved rendering of data product & dataset custom settings table
- Made the data product portal name customisable from the settings


stijn-janssens:
I find it a bit weird that the edit behavior for the portal name is different from everything else. I get why you have done it, because there is no table etc. Don't have a suggestion of how to make it more uniform, but I would have preferred to have it all similar 😅

backend/app/general_settings/model.py
```diff
@@ -0,0 +1,16 @@
...
+
+from app.database.database import Base
+
+SETTINGS_ID = 1
+
+
+class GeneralSettings(Base):
+    __tablename__ = "general_settings"
+    id = Column(Integer, primary_key=True, default=SETTINGS_ID)
+    portal_name = Column(String)
```
stijn-janssens, line 12: I would make the GeneralSettings table more generic by allowing for multiple settings to be placed inside this table. i.e. make a column `value` and a column `type` or something like that. This means we will always have a single row in this table.

stijn-janssens: This could be extended with logo etc, theming, font size, all frontend choices etc

backend/tests/conftest.py
```diff
@@ -104,6 +105,11 @@ def clear_db(session: scoped_session[Session]) -> None:
     """Clear database after each test."""
     for table in reversed(Base.metadata.sorted_tables):
         session.execute(table.delete())
+
+    # Reintroduce general settings
+    settings = GeneralSettings(id=SETTINGS_ID, portal_name="Data Product Portal")
```
stijn-janssens, line 110: why does this need to be placed? I would add the general settings in the sample data file, which ensures the record is there at testing.

frontend/src/components/branding/sidebar-logo/sidebar-logo.tsx
```diff
@@ -9,10 +10,12 @@ type Props = {
 };
 
 export function SidebarLogo({ variant = 'light' }: Props) {
-    const { t } = useTranslation();
+    const { data: settings } = useGetGeneralSettingsQuery();
     return (
         <Flex vertical className={styles.logoWrapper}>
-            <Typography.Text className={clsx(styles.logoText, styles[variant])}>{t('Logo')}</Typography.Text>
+            <Typography.Text className={clsx(styles.logoText, styles[variant])}>
+                {settings?.portal_name}
```
stijn-janssens, line 17: I would work with a default name in case the setting is not set or something goes wrong with the query

stijn-janssens (changes requested):
Loving this change 🤩

backend/app/general_settings/model.py
vlieven: Not really loving the name "general settings" tbh, is this not more related to "theme/theming settings"?

mathias-bongaerts: I considered the `value` and `type` combination and chose not to do so for the following reasons:
- There will be only a _singleton_ `GeneralSettings` object that has to be returned, splitting it up into multiple entries in a table will:
  - Make it more cumbersome and complex to reconstruct the `GeneralSettings` object
  - Create a mismatch between the model and the schema
  - Create the need for an extra `Enum` that just corresponds to the schema fields
- Have better static type checking and prevent type coercion, since the `value` column would need a generic type and need to be coerced into the correct type of the schema.

I do understand the suggestion, relational tables aren't really ment to only store a singleton.

Concerning the naming `CustomisationSettings` might be more appropriate?

backend/tests/conftest.py
mathias-bongaerts: The entire database is wiped before each test but not seeded from the sample data, as such it is necessary to add the `GeneralSettings` back again, because with the current implementation the data is expected to be present from the start; or in other words you cannot create `GeneralSettings`, a default should be present which can be updated.

For that same reason there is no entry in the sample data file, instead the entry is added in the migration script.

frontend/src/components/branding/sidebar-logo/sidebar-logo.tsx
mathias-bongaerts: Resolved with latest commit.

backend/app/general_settings/model.py
vlieven: Well, settings are by definition something you can customize, but it does not quite specify what it is that is being customized no?

mathias-bongaerts: Fair point, was looking for something more generic, but in the end these settings always boil down to changing the front-end theming, renamed `GeneralSettings` to `ThemeSettings` with latest commit.

backend/tests/conftest.py
mathias-bongaerts: I removed this from the `clear_db` fixture and created a dedicated factory instead, which is called at the beginning of each test.

stijn-janssens (approved):
@vlieven are you okay with the approach of the singleton row and column for the ThemeSettings? We discussed this briefly and can't think of a cleaner solution.

vlieven:
> @vlieven are you okay with the approach of the singleton row and column for the ThemeSettings? We discussed this briefly and can't think of a cleaner solution.

I think using a [Single-Row Table](https://nolongerset.com/reader-qa-single-row-tables/) indeed makes the most sense here. If you ever need to support different themes for different users in the same installation, or some other form of multi-tenancy, this is also trivially extended. So for me this is okay :)

PR #811, opened 2025-03-05
Add crude audit logging to all but GET calls

### LINKED ISSUE #321: Audit logs
## Summary
audit logs

### Basic example
Keep audit logs of all actions in portal

### Motivation
Audit compliance, governance

stijn-janssens:
Crude auditing system
- Added a dependency on all routes that is triggered on all but GET calls
- Adds an audit log with the authenticated user, route name and subject id and tries to auto find potential target_ids

This system gives the most info with the least amount of custom coding per route. Feedback / suggestions much appreciated

Closes #321


backend/app/datasets/service.py
```diff
@@ -129,9 +145,25 @@ def remove_dataset(self, id: UUID, db: Session):
...
+        EventService().create_event(
+            db,
+            EventCreate(
+                name="Dataset removed",
+                subject_id=dataset.id,
+                subject_type=Type.DATASET,
+                actor_id=authenticated_user.id,
+                domain_id=dataset.domain_id,
+            ),
+        )
```
tuuuuur, line 157: create_event is called before & after RefreshInfrastructureLambda() in different functions, could this change the behaviour of operation between functions?

tuuuuur (commented):
Some small remarks that might not affect the functionality

stijn-janssens:
Eventing and history has been merged to main
Audit log (log all api calls) is for now not considered to be wanted functionality

PR #812, opened 2025-03-05
Rework authz model

vlieven:
- Changes the casbin model we use to one that can be expressed more succinctly (=faster)
- Add a readme to explain how the model works
- Add functions to the Authorization class to manage roles and role assignments
- Add tests for these functions


backend/tests/__init__.py
```diff
@@ -1,22 +1,13 @@
...
-    return (
-        f"postgresql://{settings.POSTGRES_USER}:"
-        f"{settings.POSTGRES_PASSWORD}@{settings.POSTGRES_SERVER}:"
-        f"{settings.POSTGRES_PORT}/{settings.POSTGRES_DB}"
-    )
-
-
-engine = create_engine(get_test_url())
-session_factory = sessionmaker(autoflush=False, bind=engine, query_cls=MyQuery)
+engine = create_engine(database.get_url())
```
stijn-janssens, line 9: why do you change this? I assume this might break somewhere in the cicd scripts

vlieven: The implementation of `get_test_url` is identical to `database.get_url()`.
My goal was to avoid confusion due to this unnecessary duplication, unless I'm missing something?

backend/app/core/authz/authorization.py
```diff
@@ -27,7 +27,9 @@ async def initialize(cls) -> "Authorization":
 
     @staticmethod
     async def _construct_enforcer(model: str) -> AsyncEnforcer:
-        adapter = sqlalchemy_adapter.Adapter(database.get_url(async_=True))
+        """Initializes the casbin table in the DB and constructs the enforcer."""
+        url = database.get_url(async_=True)
+        adapter = sqlalchemy_adapter.Adapter(url, warning=False)
```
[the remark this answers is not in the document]
vlieven: This suppresses the warning that if you're using the default rule structure (which we are), you still need to create the table (which `adapter.create_table()` below does). 

If you keep `warning=True` as default, you'll get this warning message at application startup:
`"Using default CasbinRule table, please note the use of the 'Adapter().create_table()' method to create the table, and ignore this warning if you are using a custom CasbinRule table."`

stijn-janssens (approved):
Looking good 🤩

PR #817, opened 2025-03-06
Download CLI executables from Github

mathias-bongaerts:
Made it possible to download the CLI executables from the front-end by introducing a small download button on the user bar.

Download links refer to release assets on Github and takes into account Portal version. The list is filtered based on OS by making use of `navigator.userAgent` with a fallback showing all available executables.


stijn-janssens:
Couple of remarks:
- change icon to terminal icon
- add tooltip
- smaller padding between icons

stijn-janssens (approved):

PR #819, opened 2025-03-06
Remove unconfigured data outputs from the UI

stijn-janssens:


frontend/src/components/data-products/data-output-form/data-output-form.component.tsx
```diff
@@ -277,17 +277,27 @@ export function DataOutputForm({ mode, formRef, dataProductId, modalCallbackOnSu
...
-                            isLoading={isLoading}
-                            onTileClick={onConfigurationClick}
-                        />
-                    ))}
+                    {selectedDataPlatform?.children
+                        ?.filter((platform) => {
+                            return platformConfig?.some(
+                                (configObj) =>
+                                    configObj.platform.name == platform.label ||
+                                    configObj.service.name == platform.label,
```
vlieven, line 285: We should use `===` instead of `==`, but the linter should be able to check this too...

```suggestion
                                    configObj.platform.name === platform.label ||
                                    configObj.service.name === platform.label,
```

vlieven (approved):

PR #844, opened 2025-03-13
Chore/add field descriptions

stijn-janssens:


vlieven:
What purpose do these changes serve? (It does not help a lot with clarity IMO)

stijn-janssens:
I fully agree. Changes will be completed by @kristofmartens and only serve as purpose to add extra documentation to the openapi in order to get better performance from potential AI agentic systems

PR #884, opened 2025-03-24
First UI authz check implemented

stijn-janssens:
Initial implementation of 1 authz check in frontend


frontend/src/components/data-products/data-product-settings/data-product-settings.component.tsx
```diff
@@ -43,6 +44,24 @@ export function DataProductSettings({ dataProductId, scope }: Props) {
...
+    const { data: product_access } = useCheckAccessQuery(
+        {
+            object_id: dataProductId,
+            action: 302,
+        },
+        { skip: !dataProductId || scope !== 'dataproduct' },
+    );
+    const { data: dataset_access } = useCheckAccessQuery(
+        {
+            object_id: dataProductId,
```
vlieven, line 56: Is this the correct ID? I think this is checking the ID of a data product against an action scoped to a dataset, so this can (should) never evaluate to true

vlieven (commented):
Maybe we should re-define the enum for the actions again in the frontend? I find that just having the raw integer values does not make for very readable code...

frontend/src/components/data-products/data-product-settings/data-product-settings.component.tsx
stijn-janssens: it is, but I shall maybe refactor it to a generic object_id or just id. It started out as data product only code and now the ID swaps between data product / dataset id but the naming does not 😬

vlieven (approved):

vlieven (dismissed):

PR #885, opened 2025-03-24
Create API for role assignments

vlieven:
Endpoints added for both data product and dataset scopes.

### DIFF
--- a/backend/app/database/alembic/versions/2025_02_28_1505-269e6dbd565c_add_general_settings.py
+++ b/backend/app/database/alembic/versions/2025_02_28_1505-269e6dbd565c_add_general_settings.py
@@ -1,7 +1,7 @@
 """Add general settings
 
 Revision ID: 269e6dbd565c
-Revises: d98efbda5e7c
+Revises: 2a58b7ce3cea
 Create Date: 2025-02-28 15:05:41.442418
 
 """
@@ -16,7 +16,7 @@ from app.theme_settings.model import SETTINGS_ID, ThemeSettings
 
 # revision identifiers, used by Alembic.
 revision: str = "269e6dbd565c"
-down_revision: Union[str, None] = "d98efbda5e7c"
+down_revision: Union[str, None] = "2a58b7ce3cea"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/backend/app/database/alembic/versions/2025_03_21_1631-3b18d71bcaba_add_role_assignments.py
+++ b/backend/app/database/alembic/versions/2025_03_21_1631-3b18d71bcaba_add_role_assignments.py
@@ -1,0 +1,59 @@
+"""Add role assignments
+
+Revision ID: 3b18d71bcaba
+Revises: 269e6dbd565c
+Create Date: 2025-03-21 16:31:41.325704
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+from app.role_assignments.enums import DecisionStatus
+from app.shared.model import utcnow
+
+# revision identifiers, used by Alembic.
+revision: str = "3b18d71bcaba"
+down_revision: Union[str, None] = "269e6dbd565c"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "role_assignments_data_product",
+        sa.Column("id", sa.UUID, primary_key=True),
+        sa.Column("data_product_id", sa.UUID, sa.ForeignKey("data_products.id")),
+        sa.Column("user_id", sa.UUID, sa.ForeignKey("users.id")),
+        sa.Column("role_id", sa.UUID, sa.ForeignKey("roles.id")),
+        sa.Column("decision", sa.Enum(DecisionStatus), default=DecisionStatus.PENDING),
+        sa.Column("requested_by_id", sa.UUID, sa.ForeignKey("users.id")),
+        sa.Column("requested_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("decided_by_id", sa.UUID, sa.ForeignKey("users.id")),
+        sa.Column("decided_on", sa.DateTime(timezone=False)),
+        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
+        sa.Column("deleted_at", sa.DateTime),
+    )
+    op.create_table(
+        "role_assignments_dataset",
+        sa.Column("id", sa.UUID, primary_key=True),
+        sa.Column("dataset_id", sa.UUID, sa.ForeignKey("datasets.id")),
+        sa.Column("user_id", sa.UUID, sa.ForeignKey("users.id")),
+        sa.Column("role_id", sa.UUID, sa.ForeignKey("roles.id")),
+        sa.Column("decision", sa.Enum(DecisionStatus), default=DecisionStatus.PENDING),
+        sa.Column("requested_by_id", sa.UUID, sa.ForeignKey("users.id")),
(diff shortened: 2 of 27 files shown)


backend/app/database/alembic/versions/2025_02_28_1505-269e6dbd565c_add_general_settings.py
```diff
@@ -16,7 +16,7 @@
 
 # revision identifiers, used by Alembic.
 revision: str = "269e6dbd565c"
-down_revision: Union[str, None] = "d98efbda5e7c"
+down_revision: Union[str, None] = "2a58b7ce3cea"
```
stijn-janssens, line 19: why do you need to change this? This is a migration that is already for quite some time in main and has been migrated already. Changing this might mess with it, no? Resulting in migrations to happen twice etc?

backend/app/database/alembic/versions/2025_03_21_1631-3b18d71bcaba_add_role_assignments.py
```diff
@@ -0,0 +1,59 @@
...
+        sa.Column("decision", sa.Enum(DecisionStatus), default=DecisionStatus.PENDING),
+        sa.Column("requested_by_id", sa.UUID, sa.ForeignKey("users.id")),
+        sa.Column("requested_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("decided_by_id", sa.UUID, sa.ForeignKey("users.id")),
+        sa.Column("decided_on", sa.DateTime(timezone=False)),
+        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
+        sa.Column("deleted_at", sa.DateTime),
+    )
+
```
stijn-janssens, line 55: would you already create the domain one now, or not?

backend/app/role_assignments/data_product/router.py
```diff
@@ -0,0 +1,103 @@
...
+)
+from app.role_assignments.data_product.service import RoleAssignmentService
+from app.role_assignments.enums import DecisionStatus
+from app.role_assignments.tasks import AuthAssignment
+from app.users.model import User
+
+router = APIRouter(prefix="/data_product")
+
+
+@router.get("/")
```
stijn-janssens, line 25: You can't have trailing slashes in fastapi routes. It results in a lot of redirect errors that are unneeded. Just use ""

backend/app/role_assignments/data_product/router.py
```diff
@@ -0,0 +1,103 @@
...
+@router.delete("/{id}")
+def delete_assignment(
+    id: UUID,
+    background_tasks: BackgroundTasks,
+    db: Session = Depends(get_db_session),
+    user: User = Depends(get_authenticated_user),
+) -> None:
+    assignment = RoleAssignmentService(db=db, user=user).delete_assignment(id)
+
+    if assignment.decision is DecisionStatus.APPROVED:
```
stijn-janssens, line 55: I am afraid that doing this asynchronously might interact weirdly with the refreshes of the Redux cache in the frontend.
If you approve a pending request, you trigger a refresh of said role, which refreshes the rights of that user, but because the actually rights have not been put you don't see your new rights. And because the cache is already refreshed you have to do an explicit refresh.

I don't think it is an actual issue, since you will be approving for other users, not yourself. But just pointing the possibility out

backend/app/role_assignments/data_product/schema.py
```diff
@@ -0,0 +1,51 @@
...
+
+class RoleAssignment(RoleAssignmentResponse):
+    data_product_id: UUID
+    user_id: UUID
+    role_id: UUID
+    requested_by_id: Optional[UUID]
+    decided_by_id: Optional[UUID]
+
+
+class UpdateRoleAssignment(ORMModel):
```
stijn-janssens, line 48: If you use the ORMModel you should specify the orm model in the schema. I would expect more of these schema's to link to the actual ORM model?

stijn-janssens (commented):
Really cool logic

backend/app/database/alembic/versions/2025_02_28_1505-269e6dbd565c_add_general_settings.py
vlieven: This avoids having to create a 'merge' migration for two DB migrations that don't interact. This change pretends that the migration was created based on `n-1` instead of `n-2`. I believe this is fine as the tests keep working, but if you think this is dangerous, I can also introduce an empty migration to merge them.

backend/app/database/alembic/versions/2025_03_21_1631-3b18d71bcaba_add_role_assignments.py
vlieven: Domain is not represented yet as an authz scope outside of casbin, so I think the priority for doing this is less.
We should tackle this eventually, but this can happen after the V1 release of RBAC I think.

We do still need to add a table for global roles I realized yesterday, but I'm no longer sure if I allowed for that in the casbin policy outside of the admin role. Something to still check

backend/app/role_assignments/data_product/router.py
vlieven: Are you sure? It's literally their first example: https://fastapi.tiangolo.com/tutorial/first-steps/#first-steps ?

backend/app/role_assignments/data_product/router.py
vlieven: I see what you mean, but this is +- by design. We could indeed make everything synchronous, but in the worst case, this will delay responses of the API. In this way, the API can perform the action the user has requested, but the role syncing happens in the background. In practice, this should not matter (at least I hope so - having noticeable delays introduced by casbin would not be a good thing). Architecturally, I think this setup makes the most sense. I'm open to outside input on this though, if you think syncing the permissions should happen before sending the response?

backend/app/role_assignments/data_product/schema.py
vlieven: Maybe we should have a call on how the ORMModel class should be used, I think I'm missing some info here 😅

backend/app/database/alembic/versions/2025_02_28_1505-269e6dbd565c_add_general_settings.py
stijn-janssens: Tests keep working since they always start from a fresh db. I am only scared for actual migrated databases. I think it is fine, but I am certain that the empty merges are fine since they just introduce a fresh (albeit useless) migration. We can try but it might mess something up

backend/app/role_assignments/data_product/router.py
stijn-janssens: Might be related to how we call the API routes from the frontend. I remember having issues with it previously in this project. Also for consistency, everywhere else we use ""

backend/app/role_assignments/data_product/router.py
stijn-janssens: I think performance is indeed the more important factor here, especially since my example is purely theoretical and should not really happen in practice.

backend/app/role_assignments/data_product/schema.py
stijn-janssens: It is still confusing for me as well, and I have noticed I omitted the meta setting in various places as well.
How I understand it:
The orm model tries to solve an issue with the interaction between SQLAlchemy and nested pydantic models. They are not parsed / saved correctly with the default implementation so we explicitly link SQLAlchemy models to pydantic schemas to allow for nested parsing to happen correctly. This is a nasty workaround and is only used in some places in the code I believe. In other places we update these nested fields explicitly which is an equally good / nasty solution.

backend/app/database/alembic/versions/2025_02_28_1505-269e6dbd565c_add_general_settings.py
vlieven: Changed to using a merge migration

backend/app/role_assignments/data_product/schema.py
vlieven: Ok, still not sure that I understand, but I added the `Meta` classes

stijn-janssens (approved):

PR #895, opened 2025-03-25
Feature/rbac UI authz check

stijn-janssens:
Continuation of UI refactor with RBAC


frontend/src/components/layout/navbar/breadcrumbs/breadcrumbs.component.tsx
```diff
@@ -277,6 +280,7 @@ export const Breadcrumbs = () => {
                                         path.split('/').length == 4
                                     ) {
                                         Object.assign(breadcrumbItem, {
+                                            path: path + '#' + DataOutputTabKeys.Datasets,
```
vlieven, line 283: I would prefer to use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to construct strings instead of concatenation through addition.

frontend/src/pages/dataset/components/dataset-actions/dataset-actions.tsx
```diff
@@ -10,6 +10,8 @@ import { DataPlatform, DataPlatforms } from '@/types/data-platform';
 import { CustomDropdownItemProps } from '@/types/shared';
 
 import styles from './dataset-actions.module.scss';
+// import { useCheckAccessQuery } from '@/store/features/authorization/authorization-api-slice';
```
vlieven, line 13: Should this be commented?

vlieven (changes requested):

frontend/src/pages/dataset/components/dataset-actions/dataset-actions.tsx
stijn-janssens: Currently no integrations are enabled/can be enabled so it makes no real sense to check for access, however I wanted to already make the code. Fine or not for you?

vlieven: Clear!

vlieven (approved):

PR #902, opened 2025-03-26
Replace iterable with set because iterable can only be consumed once

stijn-janssens:
Fixes bug with data output explorer generation


vlieven (approved):
Beautiful

PR #904, opened 2025-03-26
Feature/authz check api

stijn-janssens:
initial implementation of dependency to check access in api calls

### DIFF
--- a/backend/app/core/authz/authorization.py
+++ b/backend/app/core/authz/authorization.py
@@ -1,6 +1,7 @@
+from abc import ABC, abstractmethod
 from collections.abc import Callable
 from pathlib import Path
-from typing import Self, Sequence, Type, TypeAlias, Union, cast
+from typing import Sequence, Type, TypeAlias, Union, cast
 
 import casbin_async_sqlalchemy_adapter as sqlalchemy_adapter
 from cachetools import Cache, LRUCache, cachedmethod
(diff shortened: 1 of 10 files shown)


backend/app/core/authz/authorization.py
```diff
@@ -20,7 +23,53 @@
...
 from .actions import AuthorizationAction
 
-Model: TypeAlias = Union[Type[DataProduct], Type[Dataset], None]
+Model: TypeAlias = Union[Type[DataProduct], Type[Dataset], Type[DataOutput], None]
+
+
+class SubjectResolver(ABC):
+    DEFAULT: str = "*"
+
+    @abstractmethod
```
vlieven, line 32: This can/should be decorated with `@classmethod` as well.
Reference: https://docs.python.org/3/library/abc.html#abc.abstractmethod

```suggestion
    @classmethod
    @abstractmethod
```

backend/app/core/authz/authorization.py
```diff
@@ -47,6 +96,7 @@ async def _construct_enforcer(model: str) -> AsyncEnforcer:
     def enforce(
         cls,
         action: AuthorizationAction,
+        object_fetcher: type[SubjectResolver] = DefaultResolver,
```
vlieven, line 99: We have many parameter with default values. Might be good to make these keyword only, to avoid confusion?

```suggestion
        action: AuthorizationAction,
        *,
        object_fetcher: type[SubjectResolver] = DefaultResolver,
```

vlieven, line 99:
```suggestion
        action: AuthorizationAction,
        resolver: type[SubjectResolver] = DefaultResolver,
```

backend/app/core/authz/authorization.py
```diff
@@ -55,7 +105,9 @@ def inner(
             user: User = Depends(get_authenticated_user),
             db: Session = Depends(get_db_session),
         ) -> None:
-            obj = cls.resolve_parameter(request, object_id)
+            if not settings.AUTHORIZER_ENABLED:
+                return
+            obj = object_fetcher.resolve(request, object_id, db)
```
vlieven, line 110:
```suggestion
            obj = resolver.resolve(request, object_id, db)
```

backend/app/core/authz/authorization.py
stijn-janssens: nice, I tried it the other way around and that did not work. Was not smart enough to turn the order around :P

vlieven (approved):

PR #920, opened 2025-03-31
Feature/authz check api

stijn-janssens:
Further protection of backend API calls with authz


backend/app/core/authz/actions.py
```diff
@@ -15,6 +15,8 @@ class AuthorizationAction(IntEnum):
     GLOBAL__CREATE_DATASET = 103
     GLOBAL__REQUEST_DATAPRODUCT_ACCESS = 104
     GLOBAL__REQUEST_DATASET_ACCESS = 105
+    GLOBAL__DELETE_USER = 106
+    GLOBAL__CREATE_USER = 107
```
vlieven, line 19: Dumb remark, but can we still swap the order while the PR is not merged? Usually we've put the DELETE after the corresponding CREATE action.

```suggestion
    GLOBAL__CREATE_USER = 106
    GLOBAL__DELETE_USER = 107
```

backend/app/core/authz/authorization.py
```diff
@@ -62,6 +63,10 @@ def resolve(cls, request: Request, key: str, db: Session = Depends(get_db_sessio
         return cls.DEFAULT
 
 
+class DatasetResolver(DataProductResolver):
```
vlieven, line 66: I don't think this inheritance hierarchy is modeled properly, I don't think Dataset should be a subclass of DataProduct?

backend/app/datasets/service.py
```diff
@@ -113,14 +122,45 @@ def _fetch_tags(self, db: Session, tag_ids: Iterable[UUID] = ()) -> list[TagMode
...
+                detail="Owner role not found",
+            )
         db.add(model)
         db.commit()
+        for owner in new_dataset.owners:
+            resp = create_assignment(
+                CreateRoleAssignment(
+                    dataset_id=model.id, user_id=owner.id, role_id=owner_role.id
+                ),
+                db,
```
vlieven, line 154: General remark for this function, but we might make the same design mistake elsewhere too.
I think that we're mixing responsibilities here. I think the logic should be structured more like this:

- Service manages it's DB table, and is responsible for creating the Dataset in the DB.
- The `create_dataset` function should return the full object, instead of only the ID.
- The router can then use the information in this Dataset object to call the `create_assignment` and `approve_assignment` calls. Afterwards, `RefreshInfrastructureLambda().trigger()` can be called from the router as well.

WDYT?

backend/app/core/authz/authorization.py
stijn-janssens: I agree, however I think the DataProductResolver is actually the DefaultResolver, of which the DatasetResolver just changes the model. I'll subclass DataProductResolver from the Default one as well and maybe make that an ABC as well

backend/app/datasets/service.py
stijn-janssens: That would make some drastic changes to the codebase (not saying I'm against it). For now all create functions only return the id. Also all logic currently resides in the services, with the routers being as dumb as possible. I'm not sure what the best practice is in this case to be honest.

vlieven: Well that split up sort of defeats the point of splitting the router and service IMO. Then your router is the same as your service, but you put the annotations in a separate file. By assigning concrete responsibilities to services, you can isolate their behavior, and better test those in isolation. The router can then be used to compose service calls into the action chain that is required when an API call is made.

stijn-janssens: Fair point. I will refactor it for the code I edited in this PR and see what it gives. I would not refactor the whole codebase already though.

backend/app/core/authz/actions.py
```diff
@@ -15,6 +15,8 @@ class AuthorizationAction(IntEnum):
     GLOBAL__CREATE_DATASET = 103
     GLOBAL__REQUEST_DATAPRODUCT_ACCESS = 104
     GLOBAL__REQUEST_DATASET_ACCESS = 105
+    GLOBAL__CREATE_USER = 106
+    GLOBAL__DELETE_USER = 107
```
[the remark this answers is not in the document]
vlieven: Not a code smell, but it should be used in the frontend as well. If you're defining these actions, it should also be possible to assign them to a role, that was the entire point of a configurable RBAC?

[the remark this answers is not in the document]
stijn-janssens: Ah right, my bad. Indeed it should be at least used on the configuration page of the roles. Not sure if all actions are already present there at current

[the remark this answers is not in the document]
vlieven: The ones that were there initially mirrored the state of the backend enum at the time. Hence my remark that a process to keep these two from drifting would be nice 😃 Working across two languages does not make this easy though...

vlieven (approved):

PR #922, opened 2025-03-31
No email needs to be sent on public dataset link requests

stijn-janssens:


vlieven (changes requested):
Would it not be cleaner to wrap this in a new function? So you could have a structure like:

```
if dataset.access_type != DatasetAccessType.PUBLIC:
    send_email_for_dataset_link()
```

vlieven (approved):

PR #939, opened 2025-04-02
Add assignments for global roles

### LINKED ISSUE #870: Add role assignment service + tables
Design:

- Data Product Assignments table
- Datasets Assignments table
- Global Assignments table

Sync table contents to Casbin

vlieven:
- Add service and table for global role assignments
- Modify the casbin policy to support global roles
- Make the admin role a special case of a global role
- Update the role configuration page in the frontend to keep everything ordered

Closes #870

### DIFF
--- a/backend/app/role_assignments/global_/router.py
+++ b/backend/app/role_assignments/global_/router.py
@@ -79,12 +79,6 @@ def decide_assignment(
             detail="This assignment was already decided",
         )
 
-    if request.decision is DecisionStatus.APPROVED and original.role_id is None:
-        raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail="Cannot approve a request that does not have a role assignment",
-        )
-
     assignment = service.update_assignment(
         UpdateRoleAssignment(id=id, decision=request.decision)
     )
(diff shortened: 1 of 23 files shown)


backend/app/role_assignments/global_/router.py
```diff
@@ -0,0 +1,118 @@
...
+    service = RoleAssignmentService(db=db, user=user)
+    original_role = service.get_assignment(id).role_id
+
+    if (role_id := request.role_id) == "admin":
+        role_id = ADMIN_UUID
+    assignment = service.update_assignment(UpdateRoleAssignment(id=id, role_id=role_id))
+
+    if assignment.decision is DecisionStatus.APPROVED:
+        assert (
+            original_role is not None
```
stijn-janssens, line 111: I'm not following with this logic? Why does the original_role have to be not None? Doesn't this need to be the new role?

stijn-janssens: also, modifying to an unexisting role sounds like removing? Maybe I'm completely misinterpreting this

stijn-janssens (dismissed):

stijn-janssens (approved):

PR #991, opened 2025-04-14
PR incorperating feedback from Mathias

tuuuuur:
Changes:
-> Usage of Factory for notification creation
-> Improved polymorphism
-> Distinction between pending actions & notification (= pending actions are not fetched via notification object)
-> Notification details specifics about a moment in time, not about the current state of the referenced object
-> Service logic moved to model helper functions


backend/app/notifications/model.py
```diff
@@ -1,36 +1,174 @@
...
+        "DataProductMembership",
+        back_populates="notifications",
+        passive_deletes=True,
+    )
+    __mapper_args__ = {
+        "polymorphic_identity": "DataProductMembershipNotification",
+    }
+
+
+class NotificationFactory:
```
stijn-janssens, line 91: why did you move so much logic to the model files themself? This seems like it should be in a service? Same with the removal methods? Can you explain the reasoning? :)

tuuuuur: I feel like the delete methods being in the model makes quite some sense, it allows for easy access when deleting the parent. Ideally these methods would not even be needed, but SQLAlchemy seems to struggle with the polymorphic keys for CASCADE delete.

The factories were a suggestion from Mathias, and make the code pretty clean. I can see the argument however about the code seeming like it belongs in the service class. Moving the logic shouldn't take long if preffered.

Am I missing something basic for the CASCADE? It seems kind-off odd to me that I have to manually handle it 🤔.

backend/app/notification_interactions/service.py
```diff
@@ -1,222 +1,15 @@
...
-from app.datasets.model import ensure_dataset_exists
 from app.notification_interactions.model import NotificationInteraction
 from app.notification_interactions.schema_get import NotificationInteractionGet
-from app.notifications.model import Notification
-from app.notifications.notification_types import NotificationTypes
-from app.notifications.service import NotificationService
 from app.users.schema import User
 
 
 class NotificationInteractionService:
```
vlieven, line 12: What does a "NotificationInteraction" represent?

backend/app/notifications/model.py
```diff
@@ -1,36 +1,174 @@
...
+    @staticmethod
+    def createDataProductDatasetNotification(
```
vlieven, line 93: In this function, you have an `approved` boolean, that gets mapped to a "Status" enum, that gets assigned to a field called "origin". Problems:
- It is unclear to me what "origin" represents here.
- I don't like the translation of a boolean to an enum. If need to pass something that represents the enum, just pass the enum directly as parameter, and don't bother with the `approved` boolean.
- You create a new enum type for each notification type. I think this is redundant, and can be a single enum.
We already have a "DecisionStatus" enum that you can use, which can be `PENDING`, `APPROVED` or `DENIED`.
- In my mental model, it makes sense to have an entity be "pending" on creation, and can then be "approved" or "denied" in an update step. I don't understand why a notification is created in an "approved" or "denied" state. This probably links back to me not understanding what "origin" represents here.

backend/app/notifications/model.py
```diff
@@ -1,36 +1,174 @@
...
+            }.values()
+        )
+        notification.notification_interactions = [
+            NotificationInteraction(user=receiver, notification=notification)
+            for receiver in receivers
+        ]
+        db.add(notification)
+
+    @staticmethod
+    def createDataOutputDatasetNotification(
```
vlieven, line 121: This function is nearly identical to the previous one, which leads me to believe that this probably should be a single function instead of three copies.

backend/app/notifications/model.py
```diff
@@ -1,36 +1,174 @@
...
+        )
+        receivers = list(
+            {
+                owner.id: owner
+                for owner in (
+                    data_output_dataset.dataset.owners
+                    + [data_output_dataset.requested_by]
+                )
+            }.values()
+        )
```
vlieven, line 141: What I believe you are trying to achieve here is to create a collection of unique receivers? You map the user field in your `NotificationInteraction` object below, but I believe you can achieve the same by assigning the correct `user_id`. You can simplify your code to:

```suggestion
        import itertools
        
        receivers = set(
                owner.id for owner in (
                    itertools.chain(*data_output_dataset.dataset.owners, data_output_dataset.requested_by)
                )
            )
```

backend/app/notifications/model.py
```diff
@@ -1,36 +1,174 @@
...
+        notification.notification_interactions = [
+            NotificationInteraction(user=receiver, notification=notification)
```
vlieven, line 143: This then becomes

```suggestion
            NotificationInteraction(user_id=receiver, notification=notification)
```

backend/app/notifications/model.py
vlieven: This should not be called `Factory` in my opinion. This is copying 2008 Java naming style to Python, and we already use `Factory` in the codebase to refer to the mocks used in testing.

vlieven: About the manual CASCADE, I indeed suspect that you are missing something here. The entire point of adding a CASCADE rule is that the database takes care of linked deletes for you. Having to do this manually means that the cascade is not set up correctly.

vlieven (changes requested):
In general, I find it very hard to understand the design of the code. This can mean two things:
- The solution is as complex as it needs to be in order to solve the problem. In this case, I would like to have some sort of design document that explains the design of the code, and why it should be this way.
- The solution is more complex than it needs to be. In this case, we should get rid of the accidental complexity and simplify the design.

backend/app/notification_interactions/service.py
tuuuuur: A NotificationInteraction links a Notification to a user, making it possible for each user to interact with a notification differently. Eg: you no longer want to see a notification & delete it, then this should only delete the notification for that user.

backend/app/notifications/model.py
tuuuuur: Thank you for the suggestion, that is a lot cleaner indeed

backend/app/notifications/model.py
tuuuuur: I have narrowed the scope to only include notifications in the event of the related action being resolved (= the action gets accepted or denied). This removes the need for the origin & solves the remarks stated.

backend/app/notifications/model.py
tuuuuur: CASCADE was fixed in latest commit

PR #1009, opened 2025-04-17
Draft PR examining possible Enum refactor

tuuuuur:
Draft PR examining possible Enum refactor based on Lieven's feedback:

DataProductDatasetLinkStatus, DataOutputDatasetLinkStatus, DataProductMembershipStatus & NotificationOrigin are replaced with one central enum, as they contain the same values anyway.

This does change does warrant some changes to alembic migration files, as some imports in old files are no longer resolving.


frontend/src/pages/data-output/components/data-output-tabs/dataset-tab/components/dataset-table/dataset-table-columns.tsx
```diff
@@ -67,20 +67,18 @@ export const getDataOutputDatasetsColumns = ({
             title: t('Actions'),
             key: 'action',
             render: (_, { dataset, dataset_id, status }) => {
-                const buttonText = status === 'pending_approval' ? t('Cancel') : t('Remove');
-                const popupTitle = status === 'pending_approval' ? t('Cancel Request') : t('Unlink Dataset');
+                const buttonText = status === 'pending' ? t('Cancel') : t('Remove');
```
stijn-janssens, line 70: is it not possible to use the enum here instead of the hardcoded value?

stijn-janssens: same remark in other files in frontend

tuuuuur: Now uses enum instead

Issue #317, opened 2024-10-23
Full explorer view


stijn-janssens:
## Summary
New tab: "Explorer" that shows a full map of all data products, outputs and sets.

### Basic example
Graph with all products, outputs and sets linked

### Motivation
Better overview what is present in the portal

PR #1017, opened 2025-04-18
Feature/full explorer

RobbertDM:
Bonjour tout le monde,

All data products, datasets, and data outputs in 1 huge graph on its own page.
https://github.com/conveyordata/data-product-portal/issues/317

image


stijn-janssens:
There are still some issues with the graph itself as well. Mainly direction related. Reading from or writing to a dataset are now considered equal. Instead in the first case the dataset should be the source and in the latter the target.

stijn-janssens (dismissed):
very nice! 🤩

stijn-janssens (approved):

PR #1028, opened 2025-04-23
Performance improvements for fetching data product & dataset lists

mathias-bongaerts:
## Summary

Improved performance by focusing on the interplay between SQL Alchemy, FastAPI and Pydantic.

Other improvements will follow in a later PR, most pressing issue with fetching data products and datasets is alleviated.

## Detail

By default SQL Alchemy lazily loads any relationship field defined on models, this means that on access of such fields that an additional SELECT statement is executed to fetch the necessary data. In converting any SQL Alchemy model to a Pydantic schema, Pydantic will access all fields defined on the schema in turn loading any lazily loaded relationship fields (even when fields are excluded from serialization with `Field(exclude=True)`). This results in several extra SELECT statements due to the high inter-connectivity of our data model, this effect is amplified by listing all objects, which causes numerous queries to be executed sequentially causing the slow down.

Solution to the issue is maximizing the use of eager loading and avoiding unnecessary references in Pydantic schemas to avoid loading unnecessary relations.

## Resolution

The following actions have been taken:
- Define joined loading strategies on:
    - Select queries to load the necessary relations
    - Relationships defined at the model level (only for some "default" master data fields, e.g. tags, domain, ...)
- Moved calculated fields to the model level as properties (this avoids the need to have extra references in the Pydantic schema and thus avoids loading unnecessary relations)
- Polymorphic eager loading for `DataOutputConfiguration` (defined as mapper arguments on each sub-class)
- Made the `DatasetsGet` schema more lean by making the `DataOutputLink` schema less bloated

Additionally:
- Added default loading on several models
    - `joined` loading for eager loading
    - `raise` loading to raise an error when accessing a lazily loaded field by accident
- Converted old `session.query` API to newer API

## Questions
I added the script which I used to fill up the database using the factories to the tests folder
- Should we keep this script?
- Is this the right place for the script?


tuuuuur:
Just curious, how drastic is the effect of this change on the performance?

mathias-bongaerts:
Fetching the list of data products is about 10 times faster and fetching the list of datasets is about 15 times faster.

backend/app/data_outputs/service.py
```diff
@@ -95,11 +96,25 @@ def _get_tags(self, db: Session, tag_ids: list[UUID]) -> list[TagModel]:
...
+                )
+            )
+            .unique()
+            .all()
+        )
         return data_outputs
 
     def get_data_output(self, id: UUID, db: Session) -> DataOutput:
-        return db.query(DataOutputModel).filter(DataOutputModel.id == id).first()
+        return db.get(
```
stijn-janssens, line 111: it seems like we always need this joinedload? It's not possible to that any other way? If not, then I would at least extract it to a method so we can easily reuse it

backend/app/domains/model.py
```diff
@@ -15,12 +15,23 @@
...
     name = Column(String)
     description = Column(String)
-    datasets: Mapped[list["Dataset"]] = relationship(lazy="select")
-    data_products: Mapped[list["DataProduct"]] = relationship(lazy="select")
+
+    # Relationships
+    datasets: Mapped[list["Dataset"]] = relationship(lazy="raise")
+    data_products: Mapped[list["DataProduct"]] = relationship(lazy="raise")
+
+    @property
```
stijn-janssens, line 27: this helps in computing before the query executes I assume?

stijn-janssens: where pydantic first fetches all the results?

backend/tests/seed_random_data.py
```diff
@@ -0,0 +1,63 @@
...
+    # Get the S3 service from the database for data output creation
+    service = test_session.scalar(
+        select(PlatformService).where(PlatformService.name == "S3")
+    )
+
+    # Create data products and data outputs and associate them with each other
+    for _ in range(nr_of_data_products):
+        data_product = DataProductFactory()
+
+        for _ in range(random.randint(1, 3)):
```
stijn-janssens, line 33: for sake of edge case testing I would prefer to also pick range 0 here

stijn-janssens (changes requested):
Very big improvement, love the PR.
Curious: do you have a reasoning behind which relationships are 'joined' and which 'raise' when lazy loaded? So we can follow that pattern in the future? Or is it more trial and error to spot cycles?

mathias-bongaerts:
> Very big improvement, love the PR. Curious: do you have a reasoning behind which relationships are 'joined' and which 'raise' when lazy loaded? So we can follow that pattern in the future? Or is it more trial and error to spot cycles?

My general idea was to have as a general guideline:
- Have all metadata fields related to an object load eagerly (e.g. tags, domain, type, ...)
- Have all links to other objects (e.g. data product memberships, data output dataset associations,...) _raise_ when lazily loaded to create hard boundaries between the big concepts in portal and make inclusion in queries/schemas more explicit
- Pragmatism, some hard boundaries do not make sense (e.g. no use to not load the `User` for a `DataProductMembership`) as some fields just should always be included. In general just keep a critical attitude when setting eager loading as a default on a model, but also be wary of adding the same `joinedload` on queries everywhere.

However creating the hard boundaries still proved pretty hard as our schemas are still too bloated, having too much nested levels and links to other schemas (and underlying models). Only after tackling our schemas can we actually think about (enforcing) a guideline, as otherwise it really is more trial and error to spot cycles.

backend/app/data_outputs/service.py
mathias-bongaerts: Good catch, since we use this nearly everywhere its better to just have this loaded eagerly by default.

mathias-bongaerts: Made `DataOuput.dataset_links` eagerly loaded with my latest commit

backend/tests/seed_random_data.py
mathias-bongaerts: I wanted highly interconnected data products, sets and outputs, but indeed better to have also the edge cases.

Changed with latest commit.

backend/app/data_outputs_datasets/service.py
```diff
@@ -90,23 +94,26 @@ def get_user_pending_actions(
...
+                )
+                .filter(
+                    DataOutputDatasetAssociationModel.dataset.has(
+                        DatasetModel.owners.any(UserModel.id == authenticated_user.id)
+                    )
                 )
+                .order_by(asc(DataOutputDatasetAssociationModel.requested_on))
             )
-            .order_by(asc(DataOutputDatasetAssociationModel.requested_on))
+            .unique()
```
vlieven, line 117: Where is this generating duplicates?

backend/app/data_product_memberships/service.py
```diff
@@ -247,27 +247,27 @@ def update_data_product_membership_role(
...
+                        DataProductModel.memberships.any(
+                            user_id=authenticated_user.id,
+                            role=DataProductUserRole.OWNER,
+                        )
                     )
                 )
+                .order_by(asc(DataProductMembership.requested_on))
             )
-            .order_by(asc(DataProductMembership.requested_on))
+            .unique()
```
vlieven, line 267: Same question here, why do we need to remove duplicates?

backend/app/data_products/service.py
```diff
@@ -110,47 +116,64 @@ def get_data_product(self, id: UUID, db: Session) -> DataProductGet:
...
+        )
         user_ids = [
             membership.user_id
             for membership in data_product.memberships
             if membership.role == DataProductUserRole.OWNER
         ]
-        return db.query(UserModel).filter(UserModel.id.in_(user_ids)).all()
+        return (
+            db.scalars(select(UserModel).filter(UserModel.id.in_(user_ids)))
+            .unique()
```
vlieven, line 154: Same here

backend/app/data_products/service.py
```diff
@@ -110,47 +116,64 @@ def get_data_product(self, id: UUID, db: Session) -> DataProductGet:
...
+                )
+                .filter(
+                    DataProductModel.memberships.any(
+                        user_id=user_id, status=DataProductMembershipStatus.APPROVED
+                    )
                 )
+                .order_by(asc(DataProductModel.name))
             )
-            .order_by(asc(DataProductModel.name))
+            .unique()
```
vlieven, line 176: Same here

backend/app/data_products/service.py
```diff
@@ -473,7 +514,11 @@ def get_conveyor_ide_url(self, id: UUID, db: Session) -> str:
         return CONVEYOR_SERVICE.generate_ide_url(data_product.namespace)
 
     def get_data_outputs(self, id: UUID, db: Session) -> list[DataOutputGet]:
-        return db.query(DataOutputModel).filter(DataOutputModel.owner_id == id).all()
+        return (
+            db.scalars(select(DataOutputModel).filter(DataOutputModel.owner_id == id))
+            .unique()
```
vlieven, line 519: Same here

backend/app/domains/model.py
mathias-bongaerts: Previously we calculated these fields on the schema side:

```python
class DomainsGet(Domain):
    data_products: Annotated[list[DataProduct], Field(exclude=True)]
    datasets: Annotated[list[Dataset], Field(exclude=True)]

    @computed_field
    def data_product_count(self) -> int:
        return len(self.data_products)

    @computed_field
    def dataset_count(self) -> int:
        return len(self.datasets)
```

To do this you needed the `data_products` and `datasets` fields on your schema even though they are excluded from serialization. The problem with having these fields on your schema, is that they reference another schema which might contain again several nested schemas, which, due to the interplay of Pydantic and SQL Alchemy, can result in a cascade of lazily loaded fields (even though these are completely excluded from serialization!).

By defining these computed fields on the model you purposefully limit the calculation to only load that one relation and not accidentally a ton of other relations defined on the schema.

backend/app/data_outputs_datasets/service.py
mathias-bongaerts: It's a bit of a quirk of SQL Alchemy when using eager loading, when using eager loading on one-to-many or many-to-many associations it expects you to add `unique()`.

Check out the tip under eager loading in the [documentation](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#joined-eager-loading)

SQL Alchemy even throws errors if you don't add it:
```
sqlalchemy.exc.InvalidRequestError: The unique() method must be invoked on this Result, as it contains results that include joined eager loads against collections
```

backend/app/data_products/service.py
mathias-bongaerts: This one is indeed obsolete

stijn-janssens (approved):

PR #1029, opened 2025-04-23
Pagination updates for data products & datasets

tuuuuur:
Changes to 'Data Products overview page' & 'Data Product details page':
-> FIX: Pagination now reflects the table filtering and will update accordingly
-> NEW: Added pagination where previously unused


stijn-janssens:
I like the improvement, but doesn't it make more sense to put the pagination on top of the tables? Both in the detailed as in the overview? Right now pagination can fall off the screen which results in people not knowing there is an extra page.
Also, it would be nice if we can show how many out of how many we are showing. (and even better if that is adjustable :D )

tuuuuur:
Post feedback:

->Item count on page added
->Pagination is now positioned on top 
->switched to other pagination look to prevent clutter at high amount of pages

stijn-janssens (approved):

PR #1043, opened 2025-04-28
Homepage rework

tuuuuur:
Features:
-> Filtering pending requests based on type
-> Accepting/denying pending request from the homepage is now possible
-> Small list pagination improvements
-> Added typing for pending actions in frontend
-> UI redesigned based on portal team feedback iterations


stijn-janssens (commented):
very nice PR, clean solution with the `usePendingActionHandlers`.
Only request: I see the pointers on the cards below the pending request (data products etc) are different from the pointers on the pending request. While both are clickable. I would change the pointers on the pending request to the finger pointer, unless you have an alternative in mind?

frontend/src/pages/home/components/pending-requests-inbox/pending-requests-inbox.module.scss
```diff
@@ -20,60 +21,60 @@ $item-card-width: 360px;
...
+}
+
+.width {
+    width: 100%;
+}
+
+.resolve-button {
+    margin: 0;
+    padding: 0;
+    height: auto !important;
```
stijn-janssens, line 75: out of interest, why is the `!important` necessary?

tuuuuur: I just removed it, it was not needed in this particular case. !Important is still used in other places to override antd default layout specs.

stijn-janssens (approved):

PR #1061, opened 2025-05-02
Integration test to test latest migration of Alembic

stijn-janssens:
Currently our integration test always start from a freshly initialized database, running all of the migrations at once.
I notice that the behaviour is often different if you only need to run the latest migration on a previously existing database. So I want to automate that test to prevent breaking a release

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -2,6 +2,8 @@ name: Nightly Docker Compose Test
 on:
   schedule:
     - cron: '0 2 * * *'
+  pull_request:
+    branches: ['*']
 permissions:
   contents: read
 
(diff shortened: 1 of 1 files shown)


.github/workflows/alembic-workflow.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: docker compose up -d postgresql backend
+
+      - name: Wait for Postgres
+        run: |
+          for i in {1..10}; do
+            if pg_isready -h localhost -p 5432; then
+              break
+            fi
+            sleep 3
+          done
```
vlieven, line 28: Did you check if we can do this cleaner when using the [docker compose action](https://github.com/marketplace/actions/docker-compose-action)?
(Note: I don't know if this would actually fix it, just asking the question)

.github/workflows/alembic-workflow.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Downgrade to previous migration
+        run: |
+          sleep 10
```
vlieven, line 32: Also here: The sleeps are not exactly indicative of bullet-proof flows

.github/workflows/alembic-workflow.yml
```diff
@@ -0,0 +1,47 @@
...
+          docker compose exec backend alembic downgrade -1
+
+      - name: Upgrade only latest migration
+        run: |
+          docker compose exec backend alembic upgrade head
+
+      - name: Start remaining services
+        run: docker compose up -d frontend nginx  # or your app service name
+
+      - name: Curl Test
```
vlieven, line 42: "Curl Test" is not very descriptive - maybe change it so "Check API Availability" or something along those lines?

vlieven (changes requested):
I think the initiative is very good, but I have some questions about how the setup is being done

vlieven (approved):

PR #1065, opened 2025-05-05
enable cascade delete for users

moenesbs:


stijn-janssens:
Thanks for your PR!
Some issues regarding the database cascading:
- dataset owners are not cascaded properly
- cascading the requested by and decided by means deleting a user can cause other users to lose their memberships.
- the data product memberships also have requested by and decided by that need to be handled properly, you have currently ommitted them from the migration.

My proposal would be to cascade delete when the user itself is the subject of the database entry (e.g. product membership, dataset ownership, ...) but to not cascade when the user is referenced for auditing purposes, instead set to null.

A more complicated approach could be to implement soft deletes for users instead. Allowing the requested and decided by fields to still work properly, referencing archived users instead. But that sounds like overengineering to me for now.

PR #1069, opened 2025-05-05
Added CASCADE delete to fix delete bug on main

tuuuuur:
Added CASCADE in alembic script to ensure delete behaviour.

### DIFF
--- a/backend/app/database/alembic/versions/2025_03_25_1542-c02e4638423f_add_role_assignments.py
+++ b/backend/app/database/alembic/versions/2025_03_25_1542-c02e4638423f_add_role_assignments.py
@@ -28,7 +28,7 @@ def upgrade() -> None:
         sa.Column(
             "data_product_id",
             sa.UUID,
-            sa.ForeignKey("data_products.id"),
+            sa.ForeignKey("data_products.id", ondelete="CASCADE"),
             nullable=False,
         ),
         sa.Column("user_id", sa.UUID, sa.ForeignKey("users.id"), nullable=False),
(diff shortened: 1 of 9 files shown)


tuuuuur:
Dataset & DataProduct delete fixed.
Issues fixed:

-> CASCADE not properly configured on alembic script for tables
-> 'casbin_rule' table is now also correctly affected by the delete

Manual testing done, running the following queries before & after deletion:

for dataset:
`SELECT * FROM casbin_rule WHERE v2 = 'dataset_id';`
`SELECT * FROM role_assignments_dataset rad WHERE dataset_id ='dataset_id';`

for data_product:
`SELECT * FROM casbin_rule WHERE v2 = 'data_product_id';`
`SELECT * FROM role_assignments_data_product radp WHERE data_product_id ='data_product_id';`

backend/app/database/alembic/versions/2025_03_25_1542-c02e4638423f_add_role_assignments.py
stijn-janssens, line 1: This migration has already been executed. Please create a new revision to add the missing cascades to the foreign keys instead.

backend/app/datasets/service.py
```diff
@@ -180,6 +183,11 @@ def remove_dataset(self, id: UUID, db: Session) -> None:
         db.commit()
         RefreshInfrastructureLambda().trigger()
 
+        background_tasks.add_task(
+            Authorization().clear_assignments_for_resource,
+            resource_id=str(id),
+        )
```
vlieven, line 189: Also here, this should be in the router, not the service.

PR #1070, opened 2025-05-05
Visualize domains in the Full Explorer

RobbertDM:
Draft PR, most changes kept local for now


stijn-janssens:
PR Ready for review:
Contains some code for domain nodes, this logic is hardcoded to "off" in the frontend as it is still bugged.
All other filtering should work.

backend/app/graph/router.py
```diff
@@ -1,18 +1,24 @@
...
 
 
 @router.get("")
 def get_graph_data(
     db: Session = Depends(get_db_session),
-    user: User = Depends(get_authenticated_user),
+    domain_nodes_enabled: bool = True,
+    data_product_nodes_enabled: bool = True,
+    dataset_nodes_enabled: bool = True,
+    data_output_nodes_enabled: bool = True,
```
vlieven, line 17: If I understand things correctly, this will expose these options through the OpenAPI spec. Is that the behavior that we want?

stijn-janssens: Well, it's one of those calls that belongs to the internal rather than the external API as it does not really makes sense to call it outside of the UI. However if you call it outside of it, these parameters do makes sense. They change the content of nodes and edges depending on which filter you enable or disable. Why would you not put it in the open API spec?

vlieven: Given that it's a list of booleans, this looks to me like something leftover from debugging / feature flags. But fine by me if it's the intended interface

stijn-janssens: It is a one on one mapping of toggles on the UI side. Changing a toggle triggers a new (cached) API call because we want to keep the logic in the backend. Makes more sense? Bad design?

PR #1108, opened 2025-05-12
History added for dataset, data product, data output, domain, and users (comments resolved)

tuuuuur:
Features:
-Events are now created in database for about 40 database endpoints
-An event remains when the object it refers to is deleted, allowing to see past trail of operations
-Events are displayed in frontend under the name "History"

Events are currently only deleted when the related domain gets deleted. Domains can only be deleted if there are no longer any dataset/data outputs/data products under that domain, an operation only possible for admins.

As for users, since user deletion is currently not supported: deleting a user that is referenced by an event is currently not possible.


backend/app/data_product_memberships/service.py
```diff
@@ -41,10 +43,21 @@ def request_user_access_to_data_product(
         data_product_membership = DataProductMembership(
             user_id=user.id,
             role=DataProductUserRole.MEMBER,
-            requested_by_id=authenticated_user.id,
+            requested_by_id=user_id,
```
stijn-janssens, line 46: this change does not seem correct? It is possible to execute this call for another user. Which means `user_id=user.id` is correct, but it is requested by the `authenticated_user.id` and that is not necessarily == to `user_id`

backend/app/events/schema.py
stijn-janssens, line 1: Please follow the new naming schemes for the pydantic schemas proposed by @mathias-bongaerts

backend/app/events/service.py
```diff
@@ -0,0 +1,132 @@
...
+
+from app.data_outputs.model import DataOutput
+from app.data_products.model import DataProduct
+from app.datasets.model import Dataset
+from app.events.enum import Type
+from app.events.model import Event as EventModel
+from app.users.model import User
+
+
+@event.listens_for(User, "before_delete")
```
stijn-janssens, line 14: how are these events linked to the current database? Does sqlalchemy follow up on them? Do we need to create an Alembic migration to add the event listeners to an existing database?

stijn-janssens: does it make sense to put the listeners in the service? Open question

stijn-janssens (changes requested):
Scherm­afbeelding 2025-05-15 om 12 27 14
Wondering if all of this is useful information to add in the Event table. Is there ever a use case where the dataset involved is not the dataset you are currently watching? This feels like it is double information.
The look and feel of this table is also a bit too cluttered imo.
Same with the email address, it is clear that this is a user, no?

stijn-janssens:
(Have not really started reviewing frontend yet except for look and feel)

frontend/src/utils/history.helper.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const displayType = getTypeDisplayName(t, target_type);
+
+    if (deleted_target_identifier) {
+        return deleted_target_identifier + ' ' + displayType;
+    }
+
+    switch (target_type) {
+        case EventObject.User:
+            return record.user.email + ' ' + displayType;
```
stijn-janssens, line 54: please use proper string formatting instead of `+`

stijn-janssens:
Scherm­afbeelding 2025-05-15 om 12 45 12

(Continuation of previous comment)
I would like to have the targets "linkable".
I am not a fan of having the type (dataset, user, ...) explicitly worded here, but I don't know how to make it clear otherwise.
I think having the subjects repeated here, seldomly makes sense?

backend/app/data_product_memberships/service.py
tuuuuur: Good catch

frontend/src/utils/history.helper.ts
tuuuuur: Updated in commit

backend/app/events/service.py
tuuuuur: Service seemed like the most logical place for me to put them, as they require quite some imports that would be annoying to call from the model for example. I have however opted to put them in the service file, but outside the service class itself. For this I do not have a strong reasoning, I put them outside the class because the events do not really need to be callable.

tuuuuur: I also did some research and found that the events are application-level hooks – not part of the database schema. By default they do not create database triggers or alter tables.

vlieven:
> Scherm­afbeelding 2025-05-15 om 12 45 12
> 
> (Continuation of previous comment) I would like to have the targets "linkable". I am not a fan of having the type (dataset, user, ...) explicitly worded here, but I don't know how to make it clear otherwise. I think having the subjects repeated here, seldomly makes sense?

Can we not simply merge the `Event name` and `Involved entities` columns? E.g. "Digital Marketing Data Ingestion Data Product requested access to Employee Data Dataset", and call that `Description`?

backend/app/events/enum.py
```diff
@@ -0,0 +1,9 @@
+from enum import UNIQUE, Enum, verify
+
+
+@verify(UNIQUE)
+class Type(str, Enum):
```
vlieven, line 5: `Type` is a way to generic name for this concept. The best I can come up with is `EventEntity`, although that's also not great.

backend/app/events/model.py
```diff
@@ -0,0 +1,64 @@
...
+    from app.datasets.model import Dataset
+    from app.domains.model import Domain
+    from app.users.model import User
+    from app.data_outputs.model import DataOutput
+    from app.data_products.model import DataProduct
+
+from app.shared.model import BaseORM
+
+
+class Event(Base, BaseORM):
```
vlieven, line 21: Is this model based on some kind of theoretic foundation?
It looks both very generic (actor, subject, target), and very restrictive (I can never have more than one subject or target, an event cannot exist across domains). I'm not 100% sure this is a good representation of an event.

vlieven (changes requested):

backend/app/data_outputs/service.py
```diff
@@ -141,10 +148,22 @@ def create_data_output(
         model = DataOutputModel(**data_output_schema, tags=tags, owner_id=id)
 
         db.add(model)
+        db.flush()
+        db.add(
+            EventModel(
+                name="Data output created",
```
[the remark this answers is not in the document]
tuuuuur: In the current version there are translations available in the frontend for each event, removing the need for explicit handling as is needed for enum values. I went this way mainly because of the issues alembic migrations have with enums, and any new event for future operations would need to be included in the alembic migration. This is of course possible but did not really seem worth the hassle to me, as just adding it to the translation is far easier.

backend/app/events/enum.py
tuuuuur: Very true, I'll looking into it

backend/app/events/model.py
tuuuuur: I experimented with Multiple subjects (list of subjects) but for the current database model it did not really seem worth the hassle. The only time an operation really has more than an actor & subject in the current state is on data product/ dataset edits where you edit the memberships, but in the current approach this just fires multiple events. I agree about the domains part being lacking tho, although I think it might be worth revisiting when domains become more prominent?

tuuuuur:
Screenshot 2025-05-16 at 16 18 28
UI update

backend/app/events/schema.py
tuuuuur: Schemas updated in latest commit

backend/app/database/alembic/versions/2025_05_06_1628-f967668dbf54_event_history.py
```diff
@@ -0,0 +1,59 @@
...
+
+def upgrade() -> None:
+    op.create_table(
+        "events",
+        sa.Column("id", sa.UUID, primary_key=True, nullable=False),
+        sa.Column("name", sa.String, nullable=False),
+        sa.Column("subject_id", sa.UUID),
+        sa.Column("target_id", sa.UUID, nullable=True),
+        sa.Column("subject_type", sa.Enum(Type)),
+        sa.Column("target_type", sa.Enum(Type), nullable=True),
```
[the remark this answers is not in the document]
tuuuuur: Switched to table containing string keys instead of enum in latest commit

backend/app/events/model.py
vlieven: If you model multiple relations as multiple events, I can live with that.
About the domains: if you think this should be revisited later, should we not just leave it out entirely out of the event right now?

vlieven:
> Screenshot 2025-05-16 at 16 18 28 UI update
> 
> (disabled links are for referenced objects that are deleted)

Are you still planning to merge the "Event name" and "External involved entity" columns?

backend/app/events/model.py
tuuuuur: That makes sense indeed, I'll update the model 👍

tuuuuur:
Screenshot 2025-05-19 at 14 30 56

Event name & Involved entity merged

mathias-bongaerts (changes requested):
Have yet to review the actual front end.

backend/app/events/service.py
```diff
@@ -0,0 +1,136 @@
...
+            and_(
+                EventModel.target_id == target.id,
+                EventModel.target_type == EventReferenceEntity.USER,
+            )
+        )
+        .values(deleted_target_identifier=target.email)
+    )
+
+
+@event.listens_for(Dataset, "before_delete")
```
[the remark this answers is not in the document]
tuuuuur: I opted for the event listeners in this case for the sake of simplicity, as any operation triggering a cascade delete on the objects would have to be taken into account otherwise (Only User is not prone to CASCADE delete if i remember correctly). Using events seemed like the more readable solution with less duplicate code.

tuuuuur:
Update 21/05: Ready for review, all comments should be resolved.

PR #1116, opened 2025-05-14
Authorizer is now enabled. This breaks most unit tests

stijn-janssens:
As the setting of the new authorizer is now by default enabled, most unit tests need to have the appropriate rights to be executed correctly


vlieven:
Shouldn't we update the tests to take the authorizer into account? This flag will be removed eventually right?

vlieven (commented):
All in all I think this looks good.
I'm wondering if we can automate the role syncing by overriding `_build` in the RoleFactory? (reference: https://factoryboy.readthedocs.io/en/stable/reference.html#factory.Factory._build)

backend/tests/app/role_assignments/data_product/test_auth.py
```diff
@@ -20,17 +20,13 @@ def test_add(self, authorizer: Authorization):
         user: User = UserFactory()
         role: Role = RoleFactory(scope=Scope.DATA_PRODUCT)
 
-        assignment: RoleAssignment = DataProductRoleAssignmentFactory(
+        DataProductRoleAssignmentFactory(
             data_product_id=data_product.id,
             user_id=user.id,
             role_id=role.id,
             decision=DecisionStatus.APPROVED,
         )
```
vlieven, line 28:
```suggestion
        assert not authorizer.has_resource_role(
            user_id=str(user.id), role_id=str(role.id), resource_id=str(data_product.id)
        )
        
        DataProductRoleAssignmentFactory(
            data_product_id=data_product.id,
            user_id=user.id,
            role_id=role.id,
            decision=DecisionStatus.APPROVED,
        )
```

backend/tests/app/role_assignments/data_product/test_auth.py
```diff
@@ -47,10 +43,6 @@ def test_remove(self, authorizer: Authorization):
...
 
-        assert not authorizer.has_resource_role(
-            user_id=str(user.id), role_id=str(role.id), resource_id=str(data_product.id)
-        )
```
vlieven, line 52: Move this assertion above the call to DataProductRoleAssignmentFactory

backend/tests/app/role_assignments/dataset/test_auth.py
```diff
@@ -20,17 +20,13 @@ def test_add(self, authorizer: Authorization):
...
+        DatasetRoleAssignmentFactory(
             dataset_id=dataset.id,
             user_id=user.id,
             role_id=role.id,
             decision=DecisionStatus.APPROVED,
         )
 
-        assert not authorizer.has_resource_role(
-            user_id=str(user.id), role_id=str(role.id), resource_id=str(dataset.id)
-        )
```
vlieven, line 32: Move this assert above the call to DatasetRoleAssignmentFactory

backend/tests/app/role_assignments/dataset/test_auth.py
```diff
@@ -47,10 +43,10 @@ def test_remove(self, authorizer: Authorization):
...
-        DatasetAuthAssignment(assignment).add()
+        # assert not authorizer.has_resource_role(
+        #     user_id=str(user.id), role_id=str(role.id), resource_id=str(dataset.id)
+        # )
```
vlieven, line 48: Move this assertion above the DatasetRoleAssignmentFactory call

backend/tests/conftest.py
```diff
@@ -117,16 +123,16 @@ def clear_db(session: scoped_session[Session]) -> None:
 
 @pytest.fixture
 def admin() -> UserFactory:
-    return UserFactory(external_id="sub", is_admin=True)
+    role = RoleFactory(scope=Scope.GLOBAL, prototype=Prototype.ADMIN, id=ADMIN_UUID)
+    user = UserFactory(external_id="sub", is_admin=True)
+    assignment: RoleAssignment = GlobalRoleAssignmentFactory(
+        user_id=user.id, role_id=role.id
+    )
+    GlobalAuthAssignment(assignment).add()
```
vlieven, line 131: I don't think the add() should be executed if you set up the GlobalRoleAssignmentFactory correctly

backend/tests/factories/role_assignment_global.py
```diff
@@ -12,3 +14,11 @@ class Meta:
...
     role_id = factory.Faker("uuid4")
     decision = DecisionStatus.PENDING
+
+    @factory.post_generation
+    def sync_role(self, create, extracted, **kwargs):
+        authorizer = Authorization()
+        authorizer.assign_global_role(
+            role_id=str(self.role_id), user_id=str(self.user_id)
+        )
+        test_session.commit()
```
vlieven, line 24: This should check if an admin role is being created and call the correct `assign_...()` accordingly.

vlieven (commented):
I have not put the same comments on every test_auth, but the idea should be clear I think.
Also, you have some lines commented out instead of removed, just get rid of them :)

PR #1125, opened 2025-05-15
SQL Alchemy loading strategy improvements

mathias-bongaerts:
Changes:
- Finished assigning default loading strategies to the different models with a focus on avoiding lazy loading
- Added the necessary loading options to all queries (removed some if unnecessary)
- Converted legacy `session.query` API while going over each query

Reasoning behind loading strategies:
- Eagerly load as much as possible using the joined load strategy
- Deliberately cut eager loading as a default in between data products, datasets and data outputs to avoid long (recursive) query chains. By using the default `raise` loading strategy, SQL Alchemy will throw an error when these links between objects are accessed without assigning a loading strategy at query time. This avoids accidental lazy loading.
- The relationships from metadata (e.g. tags, domain, ...) to data products, datasets and data outputs also use the `raise` loading strategy.

Excluded:
- Platform & services as this will still need a general rework
- Role & role assignments to avoid adding extra changes (and possible issues) on top

### DIFF
--- a/backend/app/users/model.py
+++ b/backend/app/users/model.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
 
 class User(Base, BaseORM):
     __tablename__ = "users"
+
     id = mapped_column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     email = Column(String, unique=True)
     external_id = Column(String)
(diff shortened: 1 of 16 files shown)


backend/app/users/model.py
```diff
@@ -19,68 +19,95 @@
...
     first_name = Column(String)
     last_name = Column(String)
     is_admin = Column(Boolean, server_default="false", nullable=False)
 
+    # Relationships - Data Products
     data_product_memberships: Mapped[list["DataProductMembership"]] = relationship(
         "DataProductMembership",
         foreign_keys="DataProductMembership.user_id",
         back_populates="user",
+        # Deliberately lazy:
```
stijn-janssens, line 35: If these complicate things. Do they still make sense after RBAC? Are they still used somewhere?

stijn-janssens (approved):
Overall very nice PR, bit of magic going on for me :P
Any idea on performance gains? Or is this purely code / db cleanup?

backend/app/users/model.py
mathias-bongaerts: With the implementation of RBAC there will be a need to review the private dataset concept and then a similar mechanism will be necessary to determine the transitive access to a private dataset (members of a data product that reads a private dataset can access the private dataset).

mathias-bongaerts:
> Overall very nice PR, bit of magic going on for me :P Any idea on performance gains? Or is this purely code / db cleanup?

The real performance improvements were in the very first PR, this PR focuses mostly on preventing the same issues from popping up again.

vlieven (approved):

PR #1126, opened 2025-05-15
Cleanup/Debug Data Product Memberships & Data Product Role Assignments part 1

tuuuuur:
This Pull request aims to cleanup data product memberships & data product role assignments. The two overlap & their interactions are pretty bugged in the current state. My recommendation is to refactor Data Product Memberships out of portal where possible.

Changes:

1) Central pending requests fetch point
2) Removed remaining Data Product Membership router endpoints (these ones were related to pending requests)
3) Introduced Router endpoints in the Data Product Role Assignment router, to make up for the ones deleted.

Frontend:
The data Product page is one of the more bugged pages after the rbac changes. Mainly editing a data product is buggy:
This operation edits the memberships, while working via the team tab changes the role assignments. The two are out of sync and behave undesirably.

Changes made on this page are to the request handling, mainly to requesting acces on a data product.

Other frontend pages have also been refactored to use the singular pending requests endpoint. (Notifications, Homepage pending requests)

The data products page still has issues after this PR. A follow up PR will be needed for this. Main focus of this follow up PR would be: 
Sync changes made via data product gear edit icon & the team tab changes


backend/app/role_assignments/data_product/service.py
```diff
@@ -21,6 +29,27 @@ def __init__(self, db: Session, user: User) -> None:
         self.db = db
         self.user = user
 
+    def _get_data_product_request_resolvers(self, data_product_id: UUID) -> List[User]:
```
stijn-janssens, line 32: what does this logic fetch exactly? I am not following why the permission needs to be present

backend/sample_data.sql
```diff
@@ -121,7 +121,7 @@ begin
...
     INSERT INTO public.users (email, id, external_id, first_name, last_name, created_on, updated_on, deleted_at) VALUES ('jane.doe@dataminded.com', 'd9f3aae2-391e-46c1-aec6-a7ae1114a7da', 'jane.doe@dataminded.com', 'Jane', 'Doe', timezone('utc'::text, CURRENT_TIMESTAMP), NULL, NULL) returning id INTO jane_id;
-    INSERT INTO public.users (email, id, external_id, first_name, last_name, created_on, updated_on, deleted_at, is_admin) VALUES ('john.doe@dataminded.com', 'b72fca38-17ff-4259-a075-5aaa5973343c', 'john.doe@dataminded.com', 'John', 'Doe', timezone('utc'::text, CURRENT_TIMESTAMP), NULL, NULL, true) returning id INTO john_id;
+    INSERT INTO public.users (email, id, external_id, first_name, last_name, created_on, updated_on, deleted_at, is_admin) VALUES ('john.doe@dataminded.com', 'b72fca38-17ff-4259-a075-5aaa5973343c', 'john.doe@dataminded.com', 'John', 'Doe', timezone('utc'::text, CURRENT_TIMESTAMP), NULL, NULL, false) returning id INTO john_id;
```
stijn-janssens, line 124: Assume this was for testing? Please revert if that is the case

backend/app/role_assignments/data_product/service.py
stijn-janssens: edit: These are requests that you are allowed to process I think? But the logic seems duplicated in the file

stijn-janssens:
Does this also fix the Data Product Owners column on the right of the Data Products Page?

vlieven (changes requested):

backend/app/role_assignments/data_product/service.py
```diff
@@ -71,3 +100,92 @@ def update_assignment(self, request: UpdateRoleAssignment) -> RoleAssignment:
...
+                .filter(DataProductRoleAssignment.decision == DecisionStatus.PENDING)
+                .filter(DataProductRoleAssignment.data_product_id.in_(data_product_ids))
+                .order_by(asc(DataProductRoleAssignment.requested_on))
+            )
+            .unique()
+            .all()
+        )
+        return actions
+
+    def request_role_assignment(
```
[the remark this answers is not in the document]
tuuuuur: Pre rbac, requesting a role would send an email to the owners. If sending the email is still desirable, it would probably be best to split the 2 endpoints. If the same endpoint is used for both operations, any time an owner adds someone to the data product team each owner will get an email to accept the person the owner is adding himself.

backend/app/role_assignments/data_product/service.py
tuuuuur: This function fetches any data product member that has permission to accept requests. This is used instead of the old deprecated way of just fetching the owners of the data product, as there can now be multiple roles that have the permission to accept requests.

backend/app/role_assignments/data_product/service.py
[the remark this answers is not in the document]
tuuuuur: Small adjustment, now only a duplicate router endpoint exists, one for requests with email & one without email

tuuuuur:
> Does this also fix the Data Product Owners column on the right of the Data Products Page?

Currently still untouched in this PR. Is displaying only the owners still the way to go for this tab? Or do new roles need to be included in the display somehow?

backend/app/role_assignments/data_product/service.py
tuuuuur: About it being duplicate, the difference is: 
Function 1) fetches all the users that have permission to accept requests for a given data product, 
Function 2) fetches all data products the authenticated user has permission on to accept requests 
(and retrieves the pending requests for these data products)

I feel like the functionality is a bit too distinct to group them together via some sort of refactoring

backend/app/role_assignments/data_product/service.py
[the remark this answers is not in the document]
stijn-janssens: This is very tricky though. Because with RBAC we allow users to have add rights for users, but not approve rights for the role. Which would mean that everyone with approve rights needs to get an email. However if adding and approving are both rights granted then sending an email is considered spam ... :D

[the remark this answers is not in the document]
tuuuuur: Merged both endpoints into one, email behaving as follows:

If the user that creates the membership assigment has permissions to accept memberships, the pending stage is bypassed & the assignment is immediatly accepted ( no email sent)

If the user cannot accept the request he is making himself, the request is made as before, but now also sends the email to the members of the data product that do have accept permissions.

vlieven:
State should be clean for merging now, however I noticed that membership removal does not work. Will investigate tomorrow

vlieven (dismissed):

backend/app/role_assignments/data_product/service.py
```diff
@@ -67,7 +109,7 @@ def delete_assignment(self, id_: UUID) -> RoleAssignment:
             and self.count_owners(assignment.data_product_id) <= 1
         ):
             raise HTTPException(
-                HTTPStatus.FORBIDDEN,
+                status.HTTP_403_FORBIDDEN,
                 "A data product must always be owned by at least one user",
```
tuuuuur, line 113: This check can be bypassed using the update assignment:
Update last owner to role that is not an owner -> bug
When removing the last owner this way, all users can also be removed from a data product -> bug

backend/app/data_products/router.py
```diff
@@ -127,6 +128,7 @@ def create_data_product(
             ),
             db=db,
             user=authenticated_user,
+            background_tasks=background_tasks,
```
tuuuuur, line 131: Does it make sense to use create_assignment from the router here? Feels kind off odd to pass the background tasks via the data product router if it will never be used? Does it make sense to create a seperate function to use in create_data_product instead of the existing create_assignment? Just thinking aloud

(I wrote this line but I'm not too sure about it being to optimal way to resolve things)

backend/app/role_assignments/data_product/service.py
vlieven: Good catch, fixed in new commit

backend/app/data_products/router.py
tuuuuur: I just mean that the email sending component of the function can never be triggered when creating a data_product, so having to provide background_tasks for the function to work is kinda odd if the background tasks will never be used. But you are right, it does not seem worth to create an extra function that does exactly the same.

frontend/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "dev": "npm run copy-config && vite",
+    "dev": "npm run copy-config && vite --force",
```
stijn-janssens, line 7: why is this needed?

frontend/src/store/features/pending-actions/pending-actions-api-slice.ts
```diff
@@ -0,0 +1,34 @@
...
+import { PendingAction } from '@/types/pending-actions/pending-actions';
+
+export const pendingActionsTags: string[] = [
+    TagTypes.DataProduct,
+    TagTypes.Dataset,
+    TagTypes.UserDataProducts,
+    TagTypes.UserDatasets,
+    TagTypes.Role,
+];
+export const dataProductsDatasetsApiSlice = baseApiSlice
```
stijn-janssens, line 13: this feels like incorrect naming?

frontend/package.json
vlieven: Not needed, but this will 're-optimize' dependencies. Thought this might be related to the padding bug I discussed with you earlier this week (but it's not). Anyway, this prevents some very strange bugs after dependency upgrades, at the cost of being a bit slower to start. If you want, I can remove it again.

stijn-janssens: if it solves strange bugs then definitely leave it in ;)

stijn-janssens (approved):

PR #1141, opened 2025-05-20
Remove memberships from the data product object

vlieven:


backend/app/data_products/model.py
```diff
@@ -59,6 +60,13 @@ class DataProduct(Base, BaseORM):
         "DataProductMembership.role",
```
stijn-janssens, line 60: too soon to remove membership from here?

backend/app/data_products/router.py
```diff
@@ -447,7 +439,6 @@ def get_graph_data(
 @router.post(
     "/{id}/settings/{setting_id}",
     dependencies=[
-        Depends(OnlyWithProductAccessID([DataProductUserRole.OWNER])),
```
stijn-janssens, line 450: Is it smart to remove the old authentication partially? I would just remove all of the dependencies in one big PR maybe? (Not this one of course)

backend/app/data_products/schema_request.py
```diff
@@ -16,22 +13,11 @@ class DataProductCreate(ORMModel):
...
     about: Optional[str] = None
-    memberships: Annotated[list[DataProductMembershipCreate], MinLen(1)]
+    owners: Annotated[list[UUID], MinLen(1)]
     domain_id: UUID
     tag_ids: list[UUID]
     lifecycle_id: UUID
 
-    @field_validator("memberships", mode="after")
-    @classmethod
-    def contains_owner(
```
stijn-janssens, line 26: doesn't this remain a requirement? At least one owner?

stijn-janssens: never mind, solved with the MinLen(1)

frontend/src/pages/data-product/components/data-product-tabs/about-tab/about-tab.tsx
```diff
@@ -23,28 +22,25 @@ export function AboutTab({ dataProductId }: Props) {
...
-    const canEditNew = edit_access?.allowed;
     if (isFetching) {
         return <LoadingSpinner />;
     }
-
     if (!dataProduct || !currentUser) {
         return <EmptyList />;
     }
 
-    const canEdit =
```
stijn-janssens, line 43: same question as on the backend: Did you remove all of the old authentication checks? Or only partially?

frontend/src/pages/data-product/components/data-product-tabs/dataset-tab/components/dataset-table/dataset-action-button.component.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { AuthorizationAction } from '@/types/authorization/rbac-actions.ts';
+import type { DatasetContract } from '@/types/dataset';
+import { DecisionStatus } from '@/types/roles';
+
+type Props = {
+    dataset: DatasetContract;
+    dataProductId: string;
+    status: DecisionStatus;
+};
+export function DatasetActionButton({ dataset, dataProductId, status }: Props) {
```
stijn-janssens, line 17: Do we want this to be separate from the data output dataset revoke button? Or would reusing code be beneficial in this case?

frontend/src/store/features/role-assignments/data-product-roles-api-slice.ts
stijn-janssens, line 1: do we want to make the other assignment api slices also already?

stijn-janssens: should we have separate / new Tags for this?

frontend/src/utils/data-product-user-role.helper.ts
```diff
@@ -1,61 +1,23 @@
...
-    userId: string,
-): DataProductMembershipRoleType {
-    return (
-        dataProduct?.memberships?.find((membership) => membership?.user_id === userId)?.role ||
-        DataProductMembershipRole.NonMember
+import { Prototype } from '@/types/roles/role.contract';
+import type { UserContract } from '@/types/users';
+
+export function useGetDataProductOwners(dataProductId: string | undefined): UserContract[] {
+    const { data: roleAssignments } = useGetRoleAssignmentQuery(
```
stijn-janssens, line 7: this is not limited to only owners, or am I wrong?

stijn-janssens: nm, Github diffing was just messing with me

stijn-janssens (commented):
Overall, great PR 🚀

backend/app/data_products/model.py
vlieven: Tried that, but that breaks the migrations... I decided to park this for now, and look into how we can fix this cleanest once these references are removed from the API layer.

backend/app/data_products/schema_request.py
vlieven: Indeed, it remains the requirement, but I've now disabled updating the owners though the form after creation.
Just implemented a check in the assignments service that should prevent the last owner being removed

backend/app/data_products/router.py
vlieven: Not sure, why do you believe it to be a problem to gradually remove these? They should not be used any more since the switch right?

frontend/src/pages/data-product/components/data-product-tabs/about-tab/about-tab.tsx
vlieven: Partially, as part of the files touched during this cleanup. I think removing all old authentication checks in one big PR will be a lot more annoying.

frontend/src/pages/data-product/components/data-product-tabs/dataset-tab/components/dataset-table/dataset-action-button.component.tsx
vlieven: Not sure. They are subtly different, generalizing this to a single component might not be more clear

backend/app/data_products/model.py
tuuuuur: If the goal is to deprecate data product memberships completely in the long run, might it not be a good idea to map them over to data product role assignments via an alembic version file in this PR?
If the existing data product memberships get replaced with this PR & removed from the models entirely, this ensures no lingering memberships remain I think no? While also ensuring no data loss? 

I think going half way in this commit & keeping the memberships in the model might cause there to be lingering memberships, given that the 'request data product membership' function etc are currently on the main, and have probably been used after the rbac migration. This would mean that for users of portal new memberships can have been created since rbac.

Am I maybe missing something?

frontend/src/store/features/role-assignments/data-product-roles-api-slice.ts
vlieven: The other assignment slices should be implemented as well yes, but I don't think it needs to be part of this PR.

As for the tags: tbh I don't quite understand the current tagging strategy, so I don't know the right answer. But probably?

backend/app/data_products/model.py
tuuuuur: I think gergely just posted about an issue caused by the 'request data product membership' currently remaining in use

vlieven: You are right Tuur, we are now in the state where old state is migrated to the new one, but we still allow old state to be created. So we should replace the old endpoints as soon as we can

stijn-janssens (approved):

PR #1144, opened 2025-05-21
API User needs to have admin rights for the API key to function properly

stijn-janssens:


backend/app/database/alembic/versions/2025_05_21_1055-6441623a586b_grant_api_user_admin.py
```diff
@@ -0,0 +1,94 @@
...
+                user=api_bot,
+            )
+            decide_assignment(
+                id=assignment.id,
+                request=DecideRoleAssignment(decision=DecisionStatus.APPROVED),
+                db=session,
+                user=api_bot,
+            )
+        else:
+            print(f"User {api_bot.email} already has the admin role.")
```
vlieven, line 69: Do we use prints in the migrations?

vlieven (approved):

PR #1145, opened 2025-05-21
Fixed an error when seeding the database

mathias-bongaerts:
Introduced when switching authorization from sync to async #1132

Seeding still happened, but it threw an error message just before the end

```sh
Something went wrong when seeding a coroutine was expected, got None
```


vlieven (approved):

vlieven:
Good catch!

PR #1147, opened 2025-05-21
Rbac/add authz to role assignments

stijn-janssens:


backend/app/core/authz/actions.py
```diff
@@ -47,3 +56,7 @@ class AuthorizationAction(IntEnum):
     DATASET__APPROVE_DATAPRODUCT_ACCESS_REQUEST = 411
     DATASET__REVOKE_DATAPRODUCT_ACCESS = 412
     DATASET__READ_INTEGRATIONS = 413
+    DATASET__ASSIGN_ROLE = 414
+    DATASET__UPDATE_ROLE_ASSIGNMENT = 415
+    DATASET__DECIDE_ROLE_ASSIGNMENT = 416
+    DATASET__DELETE_ROLE_ASSIGNMENT = 417
```
vlieven, line 62: Same remark as for data product actions

backend/app/core/authz/actions.py
```diff
@@ -17,6 +17,11 @@ class AuthorizationAction(IntEnum):
     GLOBAL__REQUEST_DATASET_ACCESS = 105
     GLOBAL__CREATE_USER = 106
     GLOBAL__DELETE_USER = 107
+    GLOBAL__UPDATE_ROLE_CONFIGURATION = 108
+    GLOBAL__ASSIGN_ROLE = 109
+    GLOBAL__UPDATE_ROLE_ASSIGNMENT = 110
+    GLOBAL__DECIDE_ROLE_ASSIGNMENT = 111
+    GLOBAL__DELETE_ROLE_ASSIGNMENT = 112
```
vlieven, line 24: Same remark as for data product actions

backend/app/core/authz/actions.py
```diff
@@ -17,6 +17,11 @@ class AuthorizationAction(IntEnum):
     GLOBAL__REQUEST_DATASET_ACCESS = 105
     GLOBAL__CREATE_USER = 106
     GLOBAL__DELETE_USER = 107
+    GLOBAL__UPDATE_ROLE_CONFIGURATION = 108
```
vlieven, line 20: The goal is to allow management of roles here, right? I wonder if we need this action, given that this basically gives users the option to elevate their rights to admin once they gain this permission. So I think it makes sense to only allow this if you have the admin role. We could check with the PO?

backend/app/core/authz/actions.py
```diff
@@ -33,6 +38,10 @@ class AuthorizationAction(IntEnum):
     DATA_PRODUCT__REQUEST_DATASET_ACCESS = 313
     DATA_PRODUCT__REVOKE_DATASET_ACCESS = 314
     DATA_PRODUCT__READ_INTEGRATIONS = 315
+    DATA_PRODUCT__ASSIGN_ROLE = 316
+    DATA_PRODUCT__UPDATE_ROLE_ASSIGNMENT = 317
+    DATA_PRODUCT__DECIDE_ROLE_ASSIGNMENT = 318
+    DATA_PRODUCT__DELETE_ROLE_ASSIGNMENT = 319
```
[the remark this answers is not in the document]
stijn-janssens: Makes sense, did not think enough :P

backend/app/role_assignments/global_/router.py
```diff
@@ -32,7 +34,16 @@ def list_assignments(
     return RoleAssignmentService(db=db, user=user).list_assignments(user_id=user_id)
 
 
-@router.post("")
+@router.post(
+    "",
+    dependencies=[
+        Depends(
+            Authorization.enforce(
+                Action.GLOBAL__ASSIGN_ROLE, resolver=DataProductResolver
```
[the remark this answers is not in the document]
stijn-janssens: I believe all GLOBAL actions also need a resolver currently which does nothing. Maybe we should either replace it with a dummy resolver or make the resolver optional. Although on a previous PR you (correctly) commented that it is risky to have the resolver, an important argument, as optional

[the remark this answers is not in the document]
vlieven: Good thing that you remember the stuff I say ^^
Maybe we should have a `GlobalResolver` that does not do anything?

backend/app/core/authz/actions.py
stijn-janssens: Yes indeed management of roles was my idea.
I get where you are coming from, it makes sense. Question is: If only admins can create admins, how do we create the first admin :D but that is an issue that won't be solved with this action anyway. So I will maybe revert it indeed. Good catch

stijn-janssens: Is there currently an "only for admin" option, that bypasses the need for an action? I think not?
You could also state that we let this action exist, and it is the end users fault if they enable it for other non-admin roles? Because I can't admin protect a route with the new RBAC yet I think

backend/app/core/authz/actions.py
stijn-janssens: I don't agree here. You have create and delete user actions, which for me really mean creating and deleting users.
As a separate set of actions you have the global role assignment. These are not the same for me?

backend/app/core/authz/actions.py
vlieven: Right now we don't have it, but we can invent something :)

backend/app/core/authz/actions.py
vlieven: Hmm, I see what you mean. That does mean we should definitely rename the actions on the data product and dataset groups regarding users. As "creation" of a user does not make sense there, only role assignments.

stijn-janssens: We can rename them. But it does make somewhat sense? Creating a user on a dataproduct means assigning him a role. Creating a user in portal is not the same as assigning a global role to a user?

I don't see why we would not distinguish them? We can always group them in the UI and make it seem like they should be enabled together, but there is no strict rule to be that way I think?

backend/app/core/authz/actions.py
stijn-janssens: We can make it "admin only" by not having the action show up in the UI to assign to other roles? That way it can be avoided only by making an explicit api call to add the action to a non-admin role? Which gives flexiblity if absolutely needed, but also protects end users from accidentally elevating all privileges?

vlieven: I think we can do it by having an action with the value "*"

backend/app/core/authz/actions.py
vlieven: Well, there are multiple reasons:
- more actions = more load on casbin
- more actions = more decisions for users to take
- more actions = more checks to implement in the UI

All of which are bad. So if we think there is no realistic scenario where we want to separate the creation of a user from assigning that user a role, we should not do it

backend/app/role_assignments/dataset/schema.py
```diff
@@ -13,7 +13,6 @@
 
 
 class CreateRoleAssignment(BaseModel):
-    dataset_id: UUID
```
vlieven, line 16: This is needed, no?

vlieven (approved):

PR #1156, opened 2025-05-23
Chore/update colour scheme

stijn-janssens:
Default Dataminded colour scheme


vlieven (dismissed):
I will assume it looks ok 😄

tuuuuur:
fancy 💜

stijn-janssens:
Scherm­afbeelding 2025-05-23 om 15 48 43

vlieven (approved):
Not loving the font TBH, but ok

frontend/src/theme/antd-theme.ts
```diff
@@ -114,6 +114,9 @@ const datamindedThemeConfig: ThemeConfig = {
         Input: {
             activeShadow: 'none',
         },
+        Badge: {
+            colorInfo: '#E6B14B',
```
vlieven, line 118: I don't think this should be hardcoded?

stijn-janssens: it is used to show the yellow pending color in the badge status. Otherwise pending gets the blue color which is somewhat weird imo

PR #1183, opened 2025-05-30
RBAC migration for admin concept

vlieven:


backend/app/database/alembic/versions/2025_05_28_1746-eb57869c9840_remove_admin_field.py
```diff
@@ -0,0 +1,29 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "eb57869c9840"
+down_revision: Union[str, None] = "93e7f3907ca8"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.drop_column("users", "is_admin")
```
stijn-janssens, line 22: open question: Do we simply want to drop this column or do we want to make a complicated query checking that all users who have is_admin are also migrated to the global role assignments table as admin?

backend/tests/app/authorization/test_router.py
```diff
@@ -39,3 +39,16 @@ def test_check_access_authorized(
 
         access = AccessResponse(**response.json())
         assert access.allowed is True
+
+    def test_is_admin(self, client: TestClient):
+        response = client.get(f"{ENDPOINT}/admin")
+        assert response.status_code == 200
+        assert response.json() is False
+
+    def test_is_admin_authorized(self, client: TestClient):
```
stijn-janssens: should this also not work with the admin fixture?

backend/app/database/alembic/versions/2025_05_28_1746-eb57869c9840_remove_admin_field.py
vlieven: Theoretically, the migration should have happened, so I would not write yet additional logic here.

backend/tests/app/authorization/test_router.py
vlieven: Yes, good point

PR #1208, opened 2025-06-05
Enable recommended biome rules

vlieven:


stijn-janssens (approved):
Nice, all of these were code smells detected by biome?

PR #1278, opened 2025-06-18
add soda logo

RobbertDM:


RobbertDM:
* `onTileClick={handleTileClick}` was added because handlePlatformClick did not work.
* Any advice on the os.env ? Do we need to set it in the helm chart? 
* Where do we enable/disable soda altogether?

stijn-janssens:
From the top of my head (not checking the code)
The difference between `onTileClick` and `onDataPlatformClick` is whether or not the dropdown with environments should toggle or you click the tile directly (i.e. difference between Conveyor and AWS) so that is a question you need to ask upfront, whether Soda will have environment specific URLs or not

stijn-janssens:
Lastly, you need to think about RBAC as well. Does everybody always have access to click on the Soda button? Or is it linked to membership on a dataset?

robbertluminus:
* Yes logo looks good.
* Yes environment specific URLs.
* Access to Soda button, if they can see the dataset they should be able to click it. It is displayed on that dataset page so I guess if you get there, you have access?

PR #1313, opened 2025-06-27
Fix empty explorer if no filters selected

stijn-janssens:


vlieven (approved):
I do have the feeling that the `includes=[]` pattern would have prevented this bug :)

PR #1332, opened 2025-07-01
Add management options to Role page

vlieven:
Allow users to modify and delete roles.


frontend/src/pages/roles/components/delete-role-modal.component.tsx
```diff
@@ -0,0 +1,189 @@
...
+                        })}
+                    />
+                    <Text>
+                        {t(
+                            'There are no users that have this role assigned. The role can be safely removed, but be aware that this cannot be undone.',
+                        )}
+                    </Text>
+                </Space>
+            ) : (
+                <Space direction="vertical" size={'middle'}>
```
stijn-janssens, line 187: In which case is it possible to have this visible? The button is disabled if count > 0

stijn-janssens: Never mind, I was only looking at prototype roles

stijn-janssens (approved):
Nice, very clean UX, love it

stijn-janssens (changes requested):
Change remaining assignments to only show uniques?

PR #1338, opened 2025-07-02
Feature/posthog analytics

Yukekbyte:
Added posthog data tracking for local testing.

a POSTHOG_KEY (phc_your_api_key) and POSTHOG_HOST (e.g. 'https://eu.i.posthog.com') should be defined in the config.local.js so that events can be tracked locally.

Added a switch in the general settings to enable/disable tracking. The code in these files (data-tracking.module and data-tracking.component) is mainly based on the files for the theme-form setting just above it.


frontend/src/pages/settings/components/settings-tabs/components/data-tracking/data-tracking.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTranslation } from 'react-i18next';
+import posthog from '@/config/posthog-config.ts';
+
+import { dispatchMessage } from '@/store/features/feedback/utils/dispatch-feedback.ts';
+
+import styles from './data-tracking.module.scss';
+
+export function DataTracking() {
+    const { t } = useTranslation();
+    const [trackingEnabled, setTrackingEnabled] = useState(() => {
```
stijn-janssens: I think this is a setting we need to save to the DB. Current behavior is (correct me if I'm wrong)
Every time a new session starts you need to disable the tracking? This is unwanted on different levels
- Only admins can toggle and view this setting, which means non-admins always have tracking enabled
- Each time the session resets they need to remember to turn this off.
Better approach would be to globally set the tracking on / off for the whole installation of portal -> a setting in the db which can be changed with this toggle

My advise would be to remove the toggle from this PR, since the behavior is not what we want and implement the correct behavior in a second PR.

stijn-janssens: (You can base yourself of the way the "Data Product Portal" name is settable and changeable through the general settings)

stijn-janssens: You will need to figure out a way to fetch the value of this setting at frontend startup time, so you can always check whether to enable / disable tracking

frontend/src/config/app-config.ts
```diff
@@ -46,6 +46,13 @@ interface Config {
      * @description This is used to customize the appearance of your application.
      */
     THEME_CONFIGURATION: string;
+
+    /**
+     * The PostHog key and host for tracking events.
+     * @description This is used to send events to PostHog for analytics.
+     */
+    POSTHOG_KEY: string;
```
stijn-janssens, line 54: If you want to see Posthog in action on the actual demo environments (capture some real data)
This setting needs to be adjustable in the helm charts. Have a look at `helm/templates/configmap_frontend.yaml` and `helm/values.yaml`

Yukekbyte:
New changes since review:

- Reverted switch to enable/disable tracking (to be revisited in another PR).
- Added POSTHOG KEY en HOST in yaml files to go beyond local.
- Hardcoded Posthog API key.

frontend/config.docker.js
```diff
@@ -41,5 +41,11 @@ const config = (() => {
          * @description This is used to customize the appearance of your application.
          */
         THEME_CONFIGURATION: 'datamindedthemeconfig',
+        /**
+         * The PostHog key and host for tracking events.
+         * @description This is used to send events to PostHog for analytics.
+         */
+        POSTHOG_KEY: 'phc_NDxOG0gXQtkPItPFJXLOAQhLmbZw7v0SbIQesSWO4gc',
```
stijn-janssens, line 48: to be seen if we want this or not. For now I'll approve it, but this makes it that all of the compose events will also be send. Potentially skewing our data

PR #1344, opened 2025-07-03
RBAC improvements

vlieven:
Improvements related to RBAC

- In the delete assignment modal, a counter is added to indicate how many assignments exist within a given dataproduct/set
- Rename user-page and related assets to people-page to be consistent with the implementation
- Add check to prevent removing the last admin from your installation


backend/app/role_assignments/global_/service.py
```diff
@@ -85,3 +80,32 @@ def update_assignment(
...
+
+    def ensure_is_global_scope(self, role_id: UUID) -> None:
+        role = self.db.get(Role, role_id)
+        if role and role.scope != Scope.GLOBAL:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Role not found for this scope",
+            )
+
+    def _guard_against_illegal_admin_removal(self, assignment: RoleAssignment) -> None:
```
stijn-janssens, line 94: I'm not against having this check, but you will need to do tricky things to remove the last admin since the bot user also counts as an admin

stijn-janssens: (and it is not shown on the people page) so you would need to dive in the db to get the ID, to then go back in the API and remove that users admin priviliges, using the api key :P
Technically possible, but I think if they want to do it they would do it from the db

stijn-janssens (changes requested):
While testing this PR I discovered an issue that is probably not caused by this PR but fits in it quite well.

In the team tab: Updating a role to the role it currently has gives an error, perhaps something you can also fix?

backend/app/role_assignments/global_/service.py
vlieven: I've tested this locally and this check works as I expected it to (not taking any bot user into account).
Or is the bot user not present on a dev environment?

stijn-janssens: Could be that it doesn't have the admin powers locally because of bad seeding script?

vlieven: Possibly, good thinking to flag this issue. I've added a filter to remove the bot user from the admin count.

stijn-janssens (approved):

PR #1354, opened 2025-07-04
Feature/posthog extra events

Yukekbyte:
Added capturing of some events with posthog.

5 type of events are captured, these are enumerated in the PosthogEvents enum in `posthog.constants.ts `.


vlieven:
I think you forgot to update your lockfile in the last commit. Can you still add this?

frontend/src/pages/datasets/components/datasets-table/datasets-table.component.tsx
```diff
@@ -49,6 +50,31 @@ export function DatasetsTable() {
...
+
+    function containsPII(searchTerm: string) {
+        // placeholder
+        return (
+            searchTerm.includes('John') ||
+            searchTerm.includes('Alice') ||
+            searchTerm.includes('Bob') ||
+            searchTerm.includes('Jane')
+        );
+    }
```
vlieven: How are you planning on implementing this function eventually?

Yukekbyte: I talked about it with one of my supervisors, and will be bringing it up tomorrow during my check-up meeting. But in my opinion, I don't think it is possible to capture any search input without risking PII as half a name, or even the first 3 letters could be classified as potential 'PII'. So I could actually delete that function for now, and add it later back in (properly implemented) if necessary.

vlieven (approved):

PR #1372, opened 2025-07-09
added homepage event tracking + linking hashed user email with events

Yukekbyte:
3 coherent changes:

- added more event captures on the homepage, more specifically:
1) Added switch on path name changes for 7 explicit events to circumvent the problem that you can't breakdown an event by property in path insights in Posthog. Although not the cleanest code (duplicate events), it is minimal and effective.
2) Grouped the UUID path name changes together with an asterisk.
3) Added capturing of the Requests Inbox of the homepage (accepting, rejecting and tabchanges).
4) Homepage Dataset and Data product tabchanges events captured (between owned and last viewed).

- Disabled Session Replays
- Fixed a posthog import that had a wrong path in `frontend/src/components/data-products/data-product-form/data-product-form.component.tsx`

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -55,6 +55,7 @@
     },
     "devDependencies": {
         "@biomejs/biome": "2.1.1",
+        "@biomejs/cli-linux-x64": "^2.1.1",
         "@types/node": "^24.0.12",
         "@types/react": "^19.1.8",
         "@types/react-dom": "^19.1.6",
@@ -65,7 +66,7 @@
         "msw": "^2.10.4",
         "sass": "^1.89.2",
         "typescript": "^5.8.3",
-        "vite": "^6.0.0",
+        "vite": "^7.0.3",
         "vite-plugin-svgr": "^4.3.0"
     },
     "optionalDependencies": {
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -48,6 +48,7 @@
             },
             "devDependencies": {
                 "@biomejs/biome": "2.1.1",
+                "@biomejs/cli-linux-x64": "^2.1.1",
                 "@types/node": "^24.0.12",
                 "@types/react": "^19.1.8",
                 "@types/react-dom": "^19.1.6",
@@ -58,7 +59,7 @@
                 "msw": "^2.10.4",
                 "sass": "^1.89.2",
                 "typescript": "^5.8.3",
-                "vite": "^6.0.0",
+                "vite": "^7.0.3",
                 "vite-plugin-svgr": "^4.3.0"
             },
             "engines": {
@@ -578,9 +579,9 @@
                 "x64"
             ],
             "dev": true,
-            "optional": true,
+            "license": "MIT OR Apache-2.0",
             "os": [
-                "darwin"
+                "linux"
             ],
             "engines": {
                 "node": ">=14.21.3"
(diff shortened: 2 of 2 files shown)


frontend/src/components/layout/root/root.layout.tsx
```diff
@@ -18,7 +18,43 @@ export default function RootLayout() {
...
-        posthog.capture(PosthogEvents.PATHNAME_CHANGED, { pathname: pathname });
+        // general capture
+        let abstract_pathname = pathname;
+
+        // if the pathname ends with a UUID, replace it with an asterisk.
+        const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-5][0-9a-f]{3}-[089ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+        if (pathname.length >= 36) {
+            const potentialUUID = pathname.slice(-36);
+            if (UUID_REGEX.test(potentialUUID)) abstract_pathname = pathname.slice(0, -36) + '*';
+        }
```
vlieven: A UUID can also exist in the middle of a path (e.g. /dataproducts/{uuid}/settings). So you will need to replace this at any point in the URL, not only at the end.  You can use [replace](https://www.freecodecamp.org/news/javascript-string-replace-example-with-regex/) in order to replace the regex by an asterisk, instead of slicing.

You can also extract this logic into a small helper function, which should be a [callback](https://react.dev/reference/react/useCallback).

frontend/src/components/layout/root/root.layout.tsx
```diff
@@ -18,7 +18,43 @@ export default function RootLayout() {
     });
 
     useEffect(() => {
-        posthog.capture(PosthogEvents.PATHNAME_CHANGED, { pathname: pathname });
+        // general capture
+        let abstract_pathname = pathname;
```
vlieven: "abstract pathname" feels a bit too... abstract. Maybe "captured pathname" is clearer?

```suggestion
        let captured_pathname = pathname;
```

frontend/src/components/layout/root/root.layout.tsx
```diff
@@ -18,7 +18,43 @@ export default function RootLayout() {
...
+            case ApplicationPaths.People:
+                posthog.capture(PosthogEvents.PATHNAME_CHANGED_PEOPLE);
+                break;
+            case ApplicationPaths.AuditLogs:
+                posthog.capture(PosthogEvents.PATHNAME_CHANGED_AUDIT_LOGS);
+                break;
+            case ApplicationPaths.Settings:
+                posthog.capture(PosthogEvents.PATHNAME_CHANGED_SETTINGS);
+                break;
+            default:
```
vlieven: You could also write this in the following way,
you first map the pathname to the event, and then you capture the event.
This avoids all the `break;` noise, and makes it clear that the goal is capturing the event.

```suggestion
        const changed_event = (() => {
            switch (pathname) {
                case ApplicationPaths.Home:
                    return PosthogEvents.PATHNAME_CHANGED_HOMEPAGE;
                case ApplicationPaths.DataProducts:
                    return PosthogEvents.PATHNAME_CHANGED_DATA_PRODUCTS;
                case ApplicationPaths.Datasets:
                    return PosthogEvents.PATHNAME_CHANGED_MARKETPLACE;
                case ApplicationPaths.Explorer:
                    return PosthogEvents.PATHNAME_CHANGED_EXPLORER;
                case ApplicationPaths.People:
                    return PosthogEvents.PATHNAME_CHANGED_PEOPLE;
                case ApplicationPaths.AuditLogs:
                    return PosthogEvents.PATHNAME_CHANGED_AUDIT_LOGS;
                case ApplicationPaths.Settings:
                    return PosthogEvents.PATHNAME_CHANGED_SETTINGS;
                default:
                    return undefined;
            }
        })();
        
        if (changed_event) {
            posthog.capture(changed_event);
        }
```

vlieven (changes requested):
Some potential code improvements

frontend/src/components/layout/root/root.layout.tsx
Yukekbyte: Great, I didn't know that function existed. That made the code way cleaner. It actually ended up being almost a one-liner so I wasn't sure if the callback was still needed or if that would be boilerplate. I could still add that if you think that's better.

frontend/src/components/layout/root/root.layout.tsx
Yukekbyte: yes, I left it out completely now (see previous comment).

frontend/src/components/layout/root/root.layout.tsx
```diff
@@ -18,7 +18,36 @@ export default function RootLayout() {
     });
 
     useEffect(() => {
-        posthog.capture(PosthogEvents.PATHNAME_CHANGED, { pathname: pathname });
+        // general capture
```
vlieven:
```suggestion
```

vlieven (approved):

frontend/src/components/layout/auth/auth.layout.tsx
```diff
@@ -95,6 +96,17 @@ export const AuthLayout = () => {
         });
     }, [events, handleAuthorizeUser, signinSilent, signoutRedirect]);
 
+    useEffect(() => {
+        if (user?.email) {
+            const hashed_id = sha256(user.email);
+            console.log(`User email: ${user.email} hashed to :${hashed_id}`);
```
stijn-janssens: Please remove the console log

frontend/src/components/layout/auth/auth.layout.tsx
```diff
@@ -95,6 +96,17 @@ export const AuthLayout = () => {
...
+            posthog.identify(hashed_id);
+        } else {
+            // Is the user logged out here??
```
stijn-janssens: If user is null then indeed the user is not logged in properly

stijn-janssens (changes requested):

frontend/package.json
```diff
@@ -64,7 +65,7 @@
         "msw": "^2.10.4",
         "sass": "^1.89.2",
         "typescript": "^5.8.3",
-        "vite": "^7.0.3",
+        "vite": "^6.0.0",
```
stijn-janssens: vite has been upgraded in main :)

PR #1388, opened 2025-07-16
Chore/do not ignore scripts to fix biome

stijn-janssens:


stijn-janssens:
Going to close this PR for now as I have not seen the issue return

PR #1397, opened 2025-07-17
Fix migration to tiptap v3

Yukekbyte:
questionable actions I have been asked to take and in no way bear responsibility for.
I refer to stijn-janssens for this PR.


stijn-janssens:
sad life, didn't hack the system :P

PR #1421, opened 2025-07-23
Feature/mcp server

stijn-janssens:
Absolute bare minimum mcp server for experimenting.

Things to figure out:
- testing
- production deploy
- correct user / client id logging to posthog

Preliminary tests with Claude Code and Claude Desktop show very promising results


backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+from app.domains.service import DomainService
+from app.settings import settings
+
+# from app.core.auth.jwt import JWTToken, get_oidc
+
+
+if settings.POSTHOG_ENABLED:
+    posthog_client = Posthog(
+        project_api_key=settings.POSTHOG_API_KEY, host=settings.POSTHOG_HOST
+    )
```
vlieven: I suppose we want to import posthog from a central location instead of redefining it here?

backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+            "source": context.source,
+            "type": context.type,
+            "method": context.method,
+        }
+
+        if hasattr(context.message, "__dict__"):
+            try:
+                entry["payload"] = context.message.__dict__
+            except (TypeError, ValueError):
+                entry["payload"] = "<non-serializable>"
```
vlieven: This looks very unpythonic - is this the logic you want?

backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+
+        try:
+            access_token: AccessToken = get_access_token()
+            user_id = get_authenticated_user(
+                token=JWTToken(sub="", token=f"Bearer {access_token.token}"),
+                db=next(get_db_session()),
+            ).id
+        except Exception:
+            user_id = None
```
vlieven: Do you want to call posthog with a `null` user id?

backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+# from app.core.auth.jwt import JWTToken, get_oidc
+
+
+if settings.POSTHOG_ENABLED:
+    posthog_client = Posthog(
+        project_api_key=settings.POSTHOG_API_KEY, host=settings.POSTHOG_HOST
+    )
+
+
+class LoggingMiddleware(Middleware):
```
vlieven: I think this class belongs in a separate file

backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+initialize_models()
+
+auth = None
+if settings.OIDC_ENABLED:
+    auth = BearerAuthProvider(
+        issuer=get_oidc().authority,
+        jwks_uri=get_oidc().jwks_uri,
+    )
+
+mcp = FastMCP(name="DataProductPortalMCP", auth=auth)
```
vlieven: This should probably be encapsulated somehow instead of floating around freely in the file

backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+        "id": user.id,
+        "external_id": user.external_id,
+        "first_name": user.first_name,
+        "email": user.email,
+        "last_name": user.last_name,
+    }
+
+
+@mcp.tool
+async def get_my_data(ctx: Context) -> dict[str, Any]:
```
vlieven: Why is this called `get_my_data`?

backend/app/mcp/mcp.py
```diff
@@ -0,0 +1,675 @@
...
+            }
+            total_count = 0
+            search_types = entity_types or [
+                "data_products",
+                "datasets",
+                "data_outputs",
+                "domains",
+            ]
+            query_results = {}
+            # Search Data Products - get all and filter manually
```
vlieven, line 108: I don't think fetching everything and then filtering manually is a sound strategy. Why not push the filtering into the database?

backend/app/mcp/router.py
```diff
@@ -0,0 +1,107 @@
...
+            "jwks_uri": get_oidc().jwks_uri,
+            "response_types_supported": ["code"],
+            "subject_types_supported": ["public"],
+            "id_token_signing_alg_values_supported": ["RS256"],
+        }
+    )
+
+
+# OAuth Protected Resource Metadata
+@router.get("/.well-known/oauth-protected-resource/mcp/mcp")
```
vlieven: Why is this router path needed?

vlieven (changes requested):

backend/app/mcp/mcp.py
stijn-janssens: if you set distinct_id to None posthog will use a random id, which I believe is what we want. If we don't have an authenticated user random id is fine?

backend/app/mcp/router.py
stijn-janssens: FastMCP library uses this to discover oauth endpoints (at least it does in the "debug" runs) I'm not sure it calls the path also in the production runs. It does not really hurt to have it I think :D

backend/app/mcp/mcp.py
stijn-janssens: agree, but wanted to leave the changes to the core app minimal for this POC

backend/app/mcp/router.py
vlieven: I mean why do we have two endpoints for a single routine

vlieven (dismissed):

backend/app/mcp/router.py
stijn-janssens: From the limited information the logs give me it looks like it uses the first path in the debug mode and the second in the actual deployed mode. I can get rid of the first one if you want I think

vlieven (approved):

PR #1464, opened 2025-08-06
Feature/explorer domain nodes

Yukekbyte:
Added domain nodes in the explorer page:

Summary of changes:

- domain toggle in explorer page (= full explorer)
- straight edges instead of bezier curves (both in full explorer and in explorer tab in dataset/product pages)
- switched from dagre to elkjs as layout algorithm
- removed node-editor.helper
- added node-parser.helper to clean up parseFullNodes() in full-explorer.tsx

Comments:
- hard coded a list of colors for domain nodes in node-parser.helper.
- hard coded the size of the regular nodes (i.e. non-domain nodes) in base-node.module.scss (not ideal...). The use-node-editor file also fetches these values with getNodeDimensions() from the document style (to not hard code it twice). 
- the handles of the nodes (where the edges start and end) have been centered for the straight edges logic but all the bezier edges logic has been preserved. Only [default-handle.module.scss](https://github.com/conveyordata/data-product-portal/compare/feature/explorer-domain-nodes?expand=1#diff-a8e6f87d06cb34508059efc6b9e3fedde74c47939b07f9eb78691c415b67e2d4) and base-node.tsx have to be reverted (Position.Top to Position.Left / Position.Right).

- Formatting for biome check ~still has to be done~ is done.


frontend/src/components/charts/custom-nodes/domain-node/domain-node.module.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.node-box {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    text-align: center;
+    border-radius: 8% !important; // 50% forces circular shape
+    //width: 350px; // set dymanically in use-node-editor.tsx by the layout algorithm
+    //height: 350px;
```
vlieven: You can just remove these instead of putting them in a comment

frontend/src/components/charts/custom-edges/straight-edge.tsx
```diff
@@ -0,0 +1,48 @@
+import { BaseEdge, type EdgeProps, getStraightPath } from '@xyflow/react';
+
+export function StraightEdge({ id, sourceX, sourceY, targetX, targetY, style }: EdgeProps) {
```
vlieven, line 3: I'm just wondering if we need to implement this ourselves? It seems like this would be the most basic thing to export for a diagram library?

Yukekbyte: Hm, there is a straight edge type yes, but I'm not sure how to integrate the margin (the early stopping of drawing the edges such that they don't collide with the node icon) that way. It seemed easier to make it custom.

backend/pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Stijn Janssens <stijn.janssens@dataminded.com>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.13"
+python = "^3.12"
```
stijn-janssens: This should not be downgraded

backend/poetry.lock
stijn-janssens, line 1: This file should not have changed

frontend/src/components/explorer/explorer.tsx
```diff
@@ -60,31 +60,23 @@ function parseNodes(nodes: NodeContract[], defaultNodePosition: XYPosition): Nod
                     };
                     break;
                 default:
-                    break;
+                    throw new Error(`Unknown node type: ${node.type}`);
```
stijn-janssens: ideally we don't throw Errors in the frontend as this hardcrashes your application. This should dispatch an error message instead and show an empty node or no node

stijn-janssens:
Couple of issues when testing:
- jumping to nodes no longer works
- auto zoom does not always work
- I liked the inactive edges better without color (just grey) than with the primary color, that makes them more distinguishable
- Sometimes the layouting algorithm still overlaps but overall it looks okay

vlieven:
Maybe it's best to have separate components for the (large) explorer view, and the (small) detail views shown on data product / dataset level.

My reasoning is that these two components share a technical foundation (show similar concepts, use `xyflow`),
but the use cases are very different.

For the detail view, a left-to-right layout makes a lot of sense, as you want to inspect up- and downstream dependencies for a single entity.

For the explorer view, the layout will be a lot less structured by the nature of it. The layout is a quite crucial part of a component that has visualising relationships as a goal. Hence my recommendation is to split these two components for now and evolve them separately. This avoids having to find a structure that sort-of works for both use cases, and allows you to focus on doing the best separate thing for each component.

Caveat: The look-and-feel should still match, so extracting and sharing some shared visual components is likely a good idea.

vlieven (dismissed):

vlieven (approved):

PR #1557, opened 2025-09-02
Use selectinload for more efficient querying

stijn-janssens:
Locally improved performance of both get products and get dataset calls from 800ms to 200ms because of the reduction of row explosion.
This changes the resulting queries from JOINS to SELECT WHERE IN

Calls like explorer go from 20 sec! to 4 sec (still bad) during performance testing (100 DPs, 100 DS)
All calls seem to benefit from this change in general.


stijn-janssens:
I agree with the remark that we should not do this without performance testing. The reddit comment is not a right heuristic to follow though.
Looking at the get datasets call we should write

```
select(DatasetModel)
                .options(
                    selectinload(DatasetModel.data_product_links)
                    .joinedload(DataProductDatasetAssociationModel.data_product)
                    .raiseload("*"),
                    selectinload(DatasetModel.data_output_links)
                    .joinedload(DataOutputDatasetAssociationModel.data_output)
                    .options(
                        joinedload(DataOutput.configuration),
                        joinedload(DataOutput.owner),
                        raiseload("*"),
                    ),
                )
```               
instead of
```
select(DatasetModel)
                .options(
                    selectinload(DatasetModel.data_product_links)
                    .selectinload(DataProductDatasetAssociationModel.data_product)
                    .raiseload("*"),
                    selectinload(DatasetModel.data_output_links)
                    .selectinload(DataOutputDatasetAssociationModel.data_output)
                    .options(
                        selectinload(DataOutput.configuration),
                        selectinload(DataOutput.owner),
                        raiseload("*"),
                    ),
                )
```

However the first query takes 5 secs and the second 0.15
Purely using it on a one-to-one or not case is not the right approach. I don't know if we can easily performance test all of the calls separately?

stijn-janssens:
It could be that the heuristic works if you consider chains to be part of the same relationship. I.e. the first selectinload is crucial because it is a one-to-many relationship, therefor everything that follows should also be selectinload?

PR #1596, opened 2025-09-08
Feature/mcp server

stijn-janssens:
MCP server broke down
For some reason the authentication stopped working (but not for localhost) when using the updated libraries.
Next to that we make the proxy file less dependent on our libraries so it can be a standalone file.
Docs to follow


docs/docs/user-guide/experimental-features.md
```diff
@@ -0,0 +1,110 @@
...
+
+```bash
+pip install uv
+````
+
+---
+
+#### 2. Download the MCP proxy script
+
+Download the MCP proxy script from the [following location](https://raw.githubusercontent.com/conveyordata/data-product-portal/refs/heads/main/backend/app/remote_proxy.py).
```
nclaeys: should we move the logic to a dedicated /mcp folder instead of in backend?

docs/docs/user-guide/experimental-features.md
```diff
@@ -0,0 +1,110 @@
...
+}
+```
+
+* Replace `/absolute/path/to/remote_proxy.py` with the full path to the downloaded `remote_proxy.py` file.
+* Replace `https://<your-portal-host>` with the actual endpoint of your Data Product Portal or `http:localhost:5050` if you are running locally.
+
+Configuration for Claude Desktop is located at: `Settings > Developer > Edit config`
+For Cursor you can find it at: `Cursor Settings > MCP & Integrations > New MCP Server`
+
+---
```
nclaeys, line 78: Maybe also add a short section on how to check whether everything works correct? Check the logs and verify it found the necessary tools?

stijn-janssens: I have placed that under ### Available tools but not too happy how it turned out. Any suggestions?

PR #1646, opened 2025-09-16
Improve filtering of datasets and data products to include your own roles

stijn-janssens:


frontend/src/components/filters/role-filter.component.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const { t } = useTranslation();
+    const currentUser = useSelector(selectCurrentUser);
+    const { data: userDatasetRoles = [] } = useGetDatasetRoleAssignmentsQuery(
+        { user_id: currentUser?.id || '' },
+        { skip: !currentUser },
+    );
+    const { data: userDataProductRoles = [] } = useGetDataProductRoleAssignmentsQuery(
+        { user_id: currentUser?.id || '' },
+        { skip: !currentUser },
+    );
```
vlieven, line 31: You can also skip one of these based on the `mode` - avoids a request for data you won't use

frontend/src/components/filters/role-filter.component.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <Typography.Text>
+                {t(
+                    mode === 'data_products'
+                        ? 'Show only data products where I am a:'
+                        : 'Show only datasets where I am a:',
+                )}
+            </Typography.Text>
+            <Select
+                placeholder={t('Select roles')}
+                value={selectedRole || undefined}
```
vlieven: I assume that you want to convert `selectedRole=''` to `undefined` instead. I'm just wondering why the type of `selectedRole` should be `string` then? Would it not be clearer to have that as `string | undefined`?

frontend/src/components/filters/role-filter.component.tsx
```diff
@@ -0,0 +1,92 @@
...
+            }
+            const roleName = assignment.role.name;
+            if (!uniqueRoles.has(roleId)) {
+                uniqueRoles.set(roleId, { label: roleName, value: roleId, productIds: [productId] });
+            } else {
+                uniqueRoles.get(roleId)!.productIds.push(productId);
+            }
+        });
+
+        return Array.from(uniqueRoles.values());
```
vlieven: This array should be sorted alphabetically by the `label` property still. That way your options will be stable in your selector, and easier to use.

frontend/src/components/filters/role-filter.component.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const currentUser = useSelector(selectCurrentUser);
+    const { data: userDatasetRoles = [] } = useGetDatasetRoleAssignmentsQuery(
+        { user_id: currentUser?.id || '' },
+        { skip: !currentUser },
+    );
+    const { data: userDataProductRoles = [] } = useGetDataProductRoleAssignmentsQuery(
+        { user_id: currentUser?.id || '' },
+        { skip: !currentUser },
+    );
+    const userRoles = mode === 'data_products' ? userDataProductRoles : userDatasetRoles;
```
vlieven: This value can be inside the `useMemo` below

frontend/src/components/filters/role-filter.component.tsx
```diff
@@ -0,0 +1,92 @@
...
+        userRoles.forEach((assignment) => {
+            const roleId = assignment.role.id;
+            let productId: string;
+            if (mode === 'data_products' && 'data_product' in assignment) {
+                productId = assignment.data_product.id;
+            } else if (mode === 'datasets' && 'dataset' in assignment) {
+                productId = assignment.dataset.id;
+            } else {
+                return;
+            }
```
vlieven: Maybe it's better to select an extractor function based on the mode instead of checking at every iteration?
e.g.

```
if mode === 'data_products' {
    get_id = (a) => a.data_product.id;
} else {
    get_id = (a) => a.dataset.id;
}

userRoles.forEach((a) => {
    const id = get_id(a);
}
```

stijn-janssens:
**Old**
Scherm­afbeelding 2025-09-16 om 10 40 10

**New**
Scherm­afbeelding 2025-09-16 om 10 39 42

frontend/src/components/filters/role-filter.component.tsx
stijn-janssens: I agree with the typing however the `|| undefined` is needed to show the placeholder after clearing a previous selected value. Otherwise it would simply use the value `""` and show an empty selection

PR #1770, opened 2025-10-08
if circular dependency exists in graph, lazy loading crashes

stijn-janssens:


vlieven:
Have you tested the performance impact of this change?
An alternative would be to not allow the state to appear in the DB?

PR #1787, opened 2025-10-10
Feature/rework datasets

stijn-janssens:
- Adds migration towards tightly coupled datasets with data products
- Blocks bad database states from API and UI
- Fixes parts of frontend

### DIFF
--- a/frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
+++ b/frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
@@ -345,7 +345,7 @@ export function DatasetForm({ mode, datasetId, dataProductId, dataOutputId }: Pr
                 rules={[
                     {
                         required: true,
-                        message: t('Please select the data product parent of the dataset'),
+                        message: t('Please select the data product to which this dataset belongs'),
                     },
                 ]}
             >


frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -318,6 +339,25 @@ export function DatasetForm({ mode, datasetId }: Props) {
                     allowClear
                 />
             </Form.Item>
+            <Form.Item<DatasetCreateFormSchema>
+                name={'data_product_id'}
+                label={t('Data Product Parent')}
+                rules={[
+                    {
+                        required: true,
+                        message: t('Please select the data product parent of the dataset'),
```
vlieven: Would you not just call this "the data product to which this dataset belongs"?

backend/app/data_products/service.py
```diff
@@ -118,6 +119,7 @@ def get_data_products(self) -> Sequence[DataProductsGet]:
                 .options(
                     selectinload(DataProductModel.dataset_links).raiseload("*"),
                     selectinload(DataProductModel.assignments).raiseload("*"),
+                    selectinload(DataProductModel.datasets).raiseload("*"),
```
nclaeys: Are these needed here? You do not return them anyway?

backend/app/data_products/service.py
```diff
@@ -154,6 +156,7 @@ def get_user_data_products(self, user_id: UUID) -> Sequence[DataProductsGet]:
...
```
nclaeys, line 158: Same as previous comment.

backend/app/data_products/service.py
stijn-janssens: Yes, for performance reasons we change the loading in SQLAlchemy to always raise an exception if implicit joins are happening. Because we return the datasets in the pydantic schema (schema_response.py l33) we need to make the join explicit.

backend/app/datasets/schema_response.py
```diff
@@ -53,4 +54,5 @@ class DatasetGet(BaseDatasetGet):
 
 
 class DatasetsGet(BaseDatasetGet):
+    data_product: DataProduct
```
nclaeys: Do we always want to return the full data_product when querying datasets, just the id is enough no? I would not add this yet. I find it more logic to add it with the datasetGet no?

backend/app/data_products/service.py
nclaeys: Ah, I would only add it to the dataProductGet and not dataProductsGet to reduce the load. Is it needed to return all datasets of a data product when listing the data products?

backend/app/datasets/schema_response.py
stijn-janssens: Currently returning the full data product to show the name of the "producing" DP in the DS overview. 
I might solve it in a cleaner way by only returning the name, or returning the ID and fetching the name

nclaeys (approved):

PR #1825, opened 2025-10-16
Update concepts in docs

vlieven:
This updates the documentation to reflect the new naming convention and how they relate to each other.

### DIFF
--- a/docs/docs/concepts/concept-linking.md
+++ b/docs/docs/concepts/concept-linking.md
@@ -16,33 +16,41 @@ config:
   layout: elk
 ---
 flowchart TD
- subgraph op1["Redshift Output Port"]
+ subgraph op1["Output Port: Realtime Equipment Metrics"]
         ta1(["Redshift Table X"])
         ta2(["Redshift Table Y"])
   end
- subgraph op2["S3 Output Port"]
+ subgraph op2["Output Port: Daily Aggregated Metrics"]
         ta3(["S3 export"])
   end
- subgraph dp1["Data Product A"]
+ subgraph op3["Output Port: Maintenance schedule"]
+        ta4(["Iceberg Table X"])
+  end
+ subgraph op4["Output Port: Maintenance alerts"]
+        ta5(["Iceberg Table Y"])
+  end
+ subgraph dp1["Data Product: Equipment Metrics"]
         op1
         op2
   end
- subgraph dp2["Data Product B"]
-        ip1["Input Port: Redshift"]
-        ip2["Input Port: S3"]
+ subgraph dp2["Data Product: Maintenance Planning"]
+        ip1["Input Port: Realtime Equipment Metrics"]
+        op3
+        op4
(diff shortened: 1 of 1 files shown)


docs/docs/concepts/concept-linking.md
```diff
@@ -1,32 +1,48 @@
...
 
 ## 🔁 Example Walkthrough
 
-1. **Data Product A** creates two outputs: a Redshift table and an S3 export.
-2. These outputs are bundled into **Dataset 1**.
-3. **Data Product B** wants to use that dataset as input.
-4. B requests access ➡️ A approves it ➡️ Access is granted.
-5. Now, B can use the data from A as input in its own data processing.
+1. **Data Product A** produces three technical assets: two Redshift tables and an S3 export.
+2. These technical assets are grouped into two **Output Ports**: one port providing access to the two Redshift tables and another port providing access to the S3 export.
```
stijn-janssens: I assume that output ports are not technology specific? If that is the case I would opt to not use a single technology in each port, because that gives the feeling that it is technology dependent

vlieven: There is no reason to enforce that, but in practice I imagine that output ports will often produce a set of data that lives on the same storage (data products that are processing related data on heterogeneous storage does not seem like a default pattern). So constructing that as an example seems less realistic?

stijn-janssens: Fair, maybe my concern can be solved by explicitly stating that output ports are technologically independent where technical assets are not? Keeping this same example regardless?

docs/docs/concepts/concept-linking.md
```diff
@@ -1,32 +1,48 @@
...
+3. **Data Product B** wants access to both Output Ports. A separate request is made for each Output Port.
+4. Data Product A can separately evaluate and approve these requests. When approved, these requests are recorded in the Data Agreement.
```
nclaeys: Do we have a data agreement concept already? If not I would not (yet) mention it here.

docs/docs/concepts/concept-linking.md
nclaeys: Maybe it helps to make it a bit less abstract to show an example and then you can easier go away from technology dependency.

I thought something like this:

DP1: equipment metrics
- Ouptput port 1: realtime equipment metrics
- Output port 2: Daily aggregated metrics

DP2: maintenance planner
- Input: daily aggregated metrics
- Ouput port 1: maintenance schedule table
- Output port 2: alerts table

What do you think?

vlieven: Good suggestion, thanks Niels :) I've updated the documentation according to your example

stijn-janssens (approved):

docs/docs/concepts/concept-linking.md
nclaeys: I like it!

PR #1858, opened 2025-10-21
Allow configuring the default user while developing.

nclaeys:


.env.docker
```diff
@@ -11,6 +11,7 @@ AWS_DEFAULT_REGION=eu-west-1
 
 # OIDC Configuration
 OIDC_ENABLED=false
+DEFAULT_USERNAME=john.scientist@pharma.com
```
stijn-janssens: In the current sample script this is john.doe@dataminded.com. 
This user is also the one that is made admin

backend/app/settings.py
```diff
@@ -30,6 +30,9 @@ class Settings(BaseSettings):
     OIDC_REDIRECT_URI: Optional[str] = None
     OIDC_AUDIENCE: Optional[str] = None
 
+    # Default username when oidc is disabled
+    DEFAULT_USERNAME: str = 'john.doe@pharma.com'
```
stijn-janssens: same remark here about the email address. It is also not consistent with the one used in .env.docker

backend/app/core/auth/auth.py
stijn-janssens, line 1: I think there are quite some unit tests who expect the current user to be "sub", so we should think of a way of keeping those tests working while also allowing you to switch users

.env.docker
stijn-janssens: scratch that, I did not push my change apparently :P

nclaeys: yeah everything is pharma atm :sweat_smile:

backend/app/core/auth/auth.py
nclaeys: Yeah noticed, reworked all of them.

backend/app/settings.py
nclaeys: I will use this as a default for now.

stijn-janssens (approved):

PR #1879, opened 2025-10-23
Initial rework docs

nclaeys:
Reworked the docs to make it more structured and remove confusion.

Main tasks:
- remove empty pages
- remove amount of pages
- moved content of user guides separated across pages into 1 clear guide

Once we finished the data product creation flow, we need to update the user guide further.

Screenshot from 2025-10-23 13-46-15

### DIFF
--- a/docs/docs/getting-started/quickstart.md
+++ b/docs/docs/getting-started/quickstart.md
@@ -18,6 +18,12 @@ Before you begin, ensure you have the following installed:
 
 ---
 
+## 🧑‍🎓 Learn the basic concepts
+
+Read through the [following pages](/docs/category/concepts) to understand the Portal concepts.
+
+---
+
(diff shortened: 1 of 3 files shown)


docs/docs/user-guide/overview.md
stijn-janssens, line 1: by removing this file you broke the "Get started button" on the homepage. You should also update that link

vlieven (approved):

docs/docs/user-guide/overview.md
nclaeys: Yes forgot that one as it is in a weird place. Fixed it for now.

nclaeys:
> when running `npm run build` there are quite a lot of broken links

Not sure where the broken links came from, only issue that I found was the getting started, roadmap one. Now I do not see any errors anymore.

PR #1882, opened 2025-10-23
Added an export of the openapi spec

stijndehaes:
This makes it easier to generate clients later
on, and it makes it easier to check for breaking
changes when changing the API.

I ensured in tests that we always keep it up to date

### DIFF
--- a/backend/app/open_api_export.py
+++ b/backend/app/open_api_export.py
@@ -1,0 +1,27 @@
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import typer
+
+from app.main import app as f_app
+
+app = typer.Typer(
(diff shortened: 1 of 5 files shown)


backend/app/open_api_export.py
```diff
@@ -0,0 +1,25 @@
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import typer
+
+from app.main import app as f_app
+
+app = typer.Typer(help="Database migration toolkit for the Data product portal.")
```
stijn-janssens: Minor comment: Please update the copied help statement to something relevant

stijndehaes: Done thanks!

PR #1892, opened 2025-10-24
Rename data outputs to technical assets and datasets to output ports

stijn-janssens:


frontend/src/pages/dataset/components/dataset-tabs/data-output-tab/components/data-output-table/data-output-table-columns.tsx
```diff
@@ -102,8 +102,8 @@ export const getDatasetDataProductsColumns = ({
                     return (
                         <Flex>
                             <Popconfirm
-                                title={t('Allow Data Output Access')}
-                                description={t('Are you sure you want to allow access to data output {{name}}?', {
+                                title={t('Allow Technical Asset Access')}
```
stijndehaes, line 105: I assume this is something we need to eventually change? People should only request access to output ports right?

stijn-janssens: If you want to input data to a dataset, now the owner of the dataset (output port) would have to accept that link. This is the place where that happens.
I am not sure how the RBAC model will evolve in the new concept and whether or not that acceptance flow will still be relevant. For now it is implemented in a way that auto-approves if you have the right to do so. If not, this approval is still necessary from this place.

stijndehaes: Thanks for the explanation :)

PR #1904, opened 2025-10-27
Fix input ports

stijn-janssens:
- do not show output ports owned by the product to use as input ports
- also block this in API

### DIFF
--- a/backend/app/data_products/service.py
+++ b/backend/app/data_products/service.py
@@ -279,6 +279,10 @@ class DataProductService:
         *,
         actor: User,
     ) -> DataProductDatasetModel:
+        """
+        Links an output port to a data product to be used as input port.
+        """
+
         dataset = ensure_dataset_exists(
             dataset_id,
             self.db,


backend/app/data_products/service.py
```diff
@@ -304,6 +304,11 @@ def link_dataset_to_data_product(
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail=f"Dataset {dataset_id} already exists in data product {id}",
             )
+        if dataset.data_product_id == data_product.id:
```
stijndehaes, line 311: Should the medthod `link_dataset_to_data_product` even still exist?

stijn-janssens: I believe it should, also in the future. This is how you add input ports to your data product after creation time.
Or am I missing something?

stijndehaes: I don't think you are missing something :) 
I assumed this was to link data output ports to data products, and there the link should be created at creation time of the output port.
So I was thinking with that in mind, using this to create input ports makes total sense!
Maybe we can add some a simple docline saying that this method is to create input ports, not output ports?

stijn-janssens: It is confusing naming, I'll acknowledge that. Linking output ports (previously datasets) is indeed "using them as input". The creation link happens automatically if you have the rights to do so. Otherwise it needs to be approved as well

PR #1906, opened 2025-10-27
Rework documentation

nclaeys:


stijn-janssens:
I have one question on how we will keep the versioned docs up to date. Right now I created a 0.4.x build which will as of now remain static. Although the latest version will continuously up date.
Is that what we want? Or should the tagged version also continuously up date until we go to a next minor bump?
In that case the release process might become (way) more complicated. Check the Release 0.4.x PR for details

README.md
```diff
@@ -9,7 +9,7 @@
 
 # Data Product Portal
 
-[![Product Name Screen Shot][product-screenshot]](https://raw.githubusercontent.com/conveyordata/data-product-portal/main/images/screenshot.png)
+[![Product screenshot][product-screenshot]](https://raw.githubusercontent.com/conveyordata/data-product-portal/rework-docs-naming/images/homepage.png)
```
stijn-janssens: this will break if we merge this branch, no?

nclaeys: Good catch, will change it

stijn-janssens (approved):

PR #1934, opened 2025-10-29
Feature/datasets domain and modal

stijn-janssens:
Fixes a couple of issues with the new output ports
- Removed create new page
- Moved form to modal to give same feel as technical asset
- Fixed invalidating of tags to automatically reload list on create and remove
- Added domain by default and uneditable

### DIFF
--- a/frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
+++ b/frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
@@ -113,8 +113,6 @@ export function DatasetForm({ mode, modalCallbackOnSubmit, formRef, datasetId, d
     const { data: dataProduct, isFetching: isFetchingDataProduct } = useGetDataProductByIdQuery(dataProductId || '', {
         skip: mode === 'edit' || !dataProductId,
     });
-    const { data: domains = [], isFetching: isFetchingDomains } = useGetAllDomainsQuery();
-    const { data: dataProducts = [], isFetching: isFetchingDataProducts } = useGetAllDataProductsQuery();
     const { data: lifecycles = [], isFetching: isFetchingLifecycles } = useGetAllDataProductLifecyclesQuery();
     const { data: users = [], isFetching: isFetchingUsers } = useGetAllUsersQuery();
     const { data: availableTags, isFetching: isFetchingTags } = useGetAllTagsQuery();
@@ -162,9 +160,7 @@ export function DatasetForm({ mode, modalCallbackOnSubmit, formRef, datasetId, d
         isFetchingTags;
 
     const accessTypeOptions: CheckboxOptionType<DatasetAccess>[] = useMemo(() => getAccessTypeOptions(t), [t]);
-    const domainSelectOptions = domains.map((domain) => ({ label: domain.name, value: domain.id }));
 
-    const dataProductSelectOptions = dataProducts.map((dp) => ({ label: dp.name, value: dp.id }));
     const userSelectOptions = users.map((owner) => ({
         label: `${owner.first_name} ${owner.last_name} (${owner.email})`,
         value: owner.id,
@@ -173,15 +169,15 @@ export function DatasetForm({ mode, modalCallbackOnSubmit, formRef, datasetId, d
 
     const onFinish: FormProps<DatasetCreateFormSchema>['onFinish'] = async (values) => {
         try {
-            if (mode === 'create') {
+            if (mode === 'create' && dataProduct) {
                 const request: DatasetCreateRequest = {
                     name: values.name,
                     namespace: values.namespace,
-                    data_product_id: values.data_product_id,
+                    data_product_id: dataProduct.id,
                     description: values.description,
                     owners: values.owners,
                     tag_ids: values.tag_ids ?? [],
-                    domain_id: values.domain_id,
+                    domain_id: dataProduct.domain.id,
                     lifecycle_id: values.lifecycle_id,
                     access_type: values.access_type,
                 };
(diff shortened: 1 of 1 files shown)


frontend/src/components/data-products/data-product-data-output-link-popup/data-product-data-output-link-popup.component.tsx
```diff
@@ -3,19 +3,23 @@ import type { ReactNode, RefObject } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { FormModal } from '@/components/modal/form-modal/form-modal.component.tsx';
-import type { DataOutputConfiguration, DataOutputCreateFormSchema } from '@/types/data-output';
-
 import styles from './data-product-data-output-link-popup.module.scss';
 
-type Props = {
+type Props<T = Record<string, unknown>> = {
```
stijndehaes: I would remove the `= Record<string, unknown>`, it's not needed in this case, because I think the code only cares about the `RefObject<FormInstance<>>` part
```suggestion
type Props<T> = {
```

frontend/src/components/data-products/data-product-data-output-link-popup/data-product-data-output-link-popup.component.tsx
```diff
@@ -3,19 +3,23 @@ import type { ReactNode, RefObject } from 'react';
...
     onClose: () => void;
     isOpen: boolean;
     title: ReactNode;
     children: ReactNode;
-    formRef: RefObject<FormInstance<DataOutputCreateFormSchema & DataOutputConfiguration> | null>;
+    formRef: RefObject<FormInstance<T> | null>;
 };
 
-export function DataProductDataOutputLinkPopup({ onClose, isOpen, title, formRef, children }: Props) {
+export function DataProductDataOutputLinkPopup<T = Record<string, unknown>>({
```
stijndehaes: Same as above
```suggestion
export function DataProductDataOutputLinkPopup<T>({
```

frontend/src/components/datasets/dataset-card/dataset-card.component.tsx
```diff
@@ -48,7 +48,7 @@ export function DatasetCard({ datasetId, draggedDataOutputId }: Props) {
     const handleRemoveDataset = useCallback(async () => {
         if (!dataset) return;
         try {
-            await removeDataset(dataset.id).unwrap();
+            await removeDataset(dataset).unwrap();
```
stijndehaes: No unwrap needed when you don't use the answer
```suggestion
            await removeDataset(dataset);
```

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -235,7 +251,7 @@ export function DatasetForm({ mode, datasetId, dataProductId, dataOutputId }: Pr
     const handleDeleteDataset = async () => {
         if (canDelete && currentDataset) {
             try {
-                await deleteDataset(currentDataset?.id).unwrap();
+                await deleteDataset(currentDataset).unwrap();
```
stijndehaes:
```suggestion
                await deleteDataset(currentDataset);
```

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -361,7 +379,7 @@ export function DatasetForm({ mode, datasetId, dataProductId, dataOutputId }: Pr
                 ]}
             >
                 <Select
-                    disabled={mode !== 'create' || !!dataProductId}
+                    disabled={true}
```
stijndehaes: Why is disabled always true here?

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -381,6 +399,7 @@ export function DatasetForm({ mode, datasetId, dataProductId, dataOutputId }: Pr
             >
                 <Select
                     loading={isFetchingDomains}
+                    disabled={true}
```
stijndehaes: Disabled always true? Why?

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -465,32 +484,32 @@ export function DatasetForm({ mode, datasetId, dataProductId, dataOutputId }: Pr
...
+                                    {t('Cancel')}
+                                </Button>
+                                <Button
+                                    className={styles.formButton}
+                                    type="primary"
+                                    htmlType={'submit'}
+                                    loading={isCreating || isUpdating}
+                                    disabled={isLoading || !canSubmit}
+                                >
+                                    {mode === 'edit' ? t('Save') : t('Create')}
```
stijndehaes: You can never be in mode edit here, since the whole block is removed if we are not in create

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -305,6 +322,7 @@ export function DatasetForm({ mode, datasetId, dataProductId, dataOutputId }: Pr
             requiredMark={'optional'}
             disabled={isLoading || !canSubmit}
             initialValues={initialValues}
+            className={mode === 'create' ? styles.form : ''}
```
stijndehaes: Why does mode create need extra padding?
I think it makes more sence to fix this in the outer component

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
stijn-janssens: To be seen whether we want to show this or not. But the data product parent and domain are always fetched from the parent product where you click "Add output port" it can never be changed. Also not in edit mode

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
stijn-janssens: See above

nclaeys (approved):

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
stijndehaes: We had a conversation about this to remove it no?

stijn-janssens: you are not on the latest commit I think :P

stijndehaes: Correct

Issue #1917, opened 2025-10-28
Shopping cart experience


stijndehaes:
## Summary

We want to offer a shopping cart like experience for requesting access to data.

- [x] Allow people to remove items from the cart in the marketplace, now they can only add
- [x] Add support for justification
- [ ] ~Think about cleaning up the API, currently actions are spread between data product router, and data product data set router for approvals etc. Maybe rename to input port router, and migrate everything there?~ won't do
- [x] Make it possible to create a new data product in the marketplace flow
- [x] ~Create new data product in modal inside of cart?~ Won't do for now
- [x] Currently checkout with 0 items is possible, both at the top and in the cart page, fix this
- [ ] Integrate posthog
- [x] Save justfication and data product id in local storage, when you go away from the page currently all is lost

PR #1954, opened 2025-10-31
Added shopping experience

stijndehaes:
This is the initial implementation that allows
people to start shopping for output ports.

In support of: #1917

View video for overview of features:

https://www.youtube.com/watch?v=YrMeHtXh10I


backend/app/data_products/router.py
```diff
@@ -414,7 +419,7 @@ def update_data_product_usage(
 
 
 @router.post(
-    "/{id}/dataset/{dataset_id}",
```
nclaeys, line 417: Do we know whether this breaks functionality for our users? Or we do not care anyway.

backend/app/data_products/schema_request.py
```diff
@@ -32,3 +32,7 @@ class DataProductStatusUpdate(ORMModel):
 
 class DataProductUsageUpdate(ORMModel):
     usage: str
+
+
+class LinkDatasetToDataProduct(ORMModel):
```
nclaeys: I would rename this to LinkDatasetsToDataProduct .

frontend/src/pages/cart/cart-output-port.component.tsx
```diff
@@ -0,0 +1,64 @@
...
+                            key: '1',
+                            label: <Typography.Text strong>Data product</Typography.Text>,
+                            children: (
+                                <Typography.Text ellipsis={{ tooltip: true }}>
+                                    {dataset.data_product_name}
+                                </Typography.Text>
+                            ),
+                            span: 2,
+                        },
+                        //TODO refactor common code with DataProductDescription
```
nclaeys: maybe extract this in a common function?

stijndehaes: This component can be removed, my mistake

backend/app/data_products/router.py
stijndehaes: It shouldn't requests are currently either done completely of platform or via the UI

backend/app/data_products/schema_request.py
stijndehaes: I also renamed `LinkDatasetToDataProductPost` to `LinkDatasetsToDataProductPost`

backend/app/data_products/router.py
stijn-janssens: This will break things though? Everyone who uses this endpoint will need to change it. That is fine for me, we have not yet been backwards compatible. But I don't agree that this has no impact on functionality?

frontend/src/pages/data-product/components/data-product-tabs/dataset-tab/dataset-tab.tsx
```diff
@@ -57,19 +54,15 @@ export function DatasetTab({ dataProductId }: Props) {
                     formItemProps={{ initialValue: '', className: styles.marginBottomLarge }}
                     form={searchForm}
                     actionButton={
-                        <Button
```
stijn-janssens, line 60: Do we really want to remove this button? Is the shopping cart the only way to request access?
I had in my mind that the shopping cart would be for new products primarily, but it is implemented right now to only work with existing products?
Just a question

frontend/src/pages/data-product/components/data-product-tabs/dataset-tab/dataset-tab.tsx
```diff
@@ -57,19 +54,15 @@ export function DatasetTab({ dataProductId }: Props) {
...
-                            disabled={!canCreateDataset}
-                            type={'primary'}
-                            className={styles.formButton}
-                            onClick={handleOpen}
-                        >
-                            {t('Link Output Port')}
-                        </Button>
+                        <Link to={ApplicationPaths.Datasets}>
```
stijn-janssens, line 57: Sure, makes sense. Never mind my comment above. Might feel like it is an extra step but the feeling and flow will probably be nicer

stijn-janssens (approved):
Not much to remark, clean PR, movie helps a ton

backend/app/data_products/router.py
stijndehaes: I can also deprecate the old call, and just implement is by using the call with a single ID?

frontend/src/pages/data-product/components/data-product-tabs/dataset-tab/dataset-tab.tsx
stijndehaes: We can always revert things after user feedback :)

backend/app/data_products/router.py
stijndehaes: I added the old call as a deprecated call, using the new one.

nclaeys (approved):

Issue #1944, opened 2025-10-30
Improve marketplace cards


stijn-janssens:
## Summary
- add data product parent to detail view (clickable)
- data product in card clickable
- tooltip when hovering over linked products and make them clickable

### Motivation
Why do we need this improvement? What are the benefits?

PR #1955, opened 2025-10-31
Feature/card improvements

stijn-janssens:
Solves #1944

### DIFF
--- a/frontend/src/pages/dataset/components/dataset-tabs/data-product-tab/data-product-tab.tsx
+++ b/frontend/src/pages/dataset/components/dataset-tabs/data-product-tab/data-product-tab.tsx
@@ -39,19 +39,14 @@ export function DataProductTab({ datasetId }: Props) {
     }, [datasetDataProducts, searchTerm]);
 
     return (
-        <>
-            <Flex
-                vertical
-                className={`${styles.container} ${filteredDataProducts.length === 0 && styles.paginationGap}`}
-            >
-                <Searchbar
-                    placeholder={t('Search data products by name')}
-                    formItemProps={{ initialValue: '', className: styles.marginBottomLarge }}
-                    form={searchForm}
-                />
+        <Flex vertical className={`${styles.container} ${filteredDataProducts.length === 0 && styles.paginationGap}`}>
+            <Searchbar
+                placeholder={t('Search data products by name')}
+                formItemProps={{ initialValue: '', className: styles.marginBottomLarge }}
+                form={searchForm}
+            />
 
-                <DataProductTable datasetId={datasetId} dataProducts={filteredDataProducts} isLoading={isLoading} />
-            </Flex>
-        </>
+            <DataProductTable datasetId={datasetId} dataProducts={filteredDataProducts} isLoading={isLoading} />
+        </Flex>
     );
 }


frontend/src/components/datasets/dataset-marketplace-card/dataset-marketplace-card.component.tsx
```diff
@@ -0,0 +1,171 @@
+import {
```
stijndehaes: Since this component isn't reused, I would rather have it in the markpetplace folder, maybe add it to a components folder there?

frontend/src/components/datasets/dataset-marketplace-card/dataset-marketplace-card.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { useGetDatasetByIdQuery } from '@/store/features/datasets/datasets-api-slice.ts';
+import type { DatasetContract } from '@/types/dataset';
+import { createDataProductIdPath, createDatasetIdPath } from '@/types/navigation.ts';
+import styles from './dataset-marketplace-card.module.scss';
+
+type Props = {
+    datasetId: string;
+};
+
+export function DatasetMarketplaceCard({ datasetId }: Props) {
```
nclaeys: clearly you did not agree with my reasoning for keeping the card in the marketplace page. 
I want to avoid this becoming a habbit. Either we agree and let it go or we change it immediately :wink:.

stijn-janssens: not that I did not agree. I had to move it because I needed the hooks on the card level. Which is not allowed in a map.
Whether or not this needs to be in the components or the folder I have no opinion about

stijn-janssens:
Scherm­afbeelding 2025-10-31 om 14 47 37

Scherm­afbeelding 2025-10-31 om 14 47 55
Scherm­afbeelding 2025-10-31 om 14 48 09

frontend/src/pages/dataset/components/dataset-description/dataset-description.tsx
```diff
@@ -29,6 +37,10 @@ export function DatasetDescription({ lifecycle, accessType, description, domain,
                     <Typography.Text strong>{t('Namespace')}</Typography.Text>
                     <Typography.Text>{namespace}</Typography.Text>
                 </Flex>
+                <Flex className={styles.statusBadge}>
+                    <Typography.Text strong>{t('Data Product')}</Typography.Text>
+                    <Typography.Text>{data_product.name}</Typography.Text>
```
stijndehaes: Maybe make this a link to the data product?

frontend/src/pages/marketplace/dataset-marketplace-card/dataset-marketplace-card.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+import styles from './dataset-marketplace-card.module.scss';
+
+type Props = {
+    datasetId: string;
+};
+
+export function DatasetMarketplaceCard({ datasetId }: Props) {
+    const { t } = useTranslation();
+    const navigate = useNavigate();
+    const { data: dataset, isLoading } = useGetDatasetByIdQuery(datasetId, { skip: !datasetId });
```
stijndehaes: I would rather pass the dataset from the marketplace, instead of doing a query here again?
Because this puts unneeded pressure in the API.
Similar with the next call to get the data product.

After discussion we decided to extend the search for datasets to give it what we need here, except we don't want to add the `data_product_links` in the search call.
We want to load these when the user hovers over the tooltip instead

frontend/src/pages/marketplace/dataset-marketplace-card/dataset-card-tooltip.component.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { createDataProductIdPath } from '@/types/navigation';
+import styles from './dataset-marketplace-card.module.scss';
+
+type Props = {
+    dataset_id: string;
+};
+
+export function DatasetCardTooltip({ dataset_id }: Props) {
+    const { t } = useTranslation();
+    const { data: dataset } = useGetDatasetByIdQuery(dataset_id, { skip: !dataset_id });
```
stijndehaes, line 15: Can you quickly test out if this actually does what we want? It might still load this query without people needing to over hover the item.

frontend/src/pages/marketplace/dataset-marketplace-card/dataset-card-tooltip.component.tsx
```diff
@@ -0,0 +1,37 @@
...
+    if (!dataset) return <Tooltip title={<LoadingSpinner />} />;
+    return (
+        <Tooltip
+            placement="bottom"
+            color="white"
+            title={
+                <List
```
stijndehaes, line 22: I think the solution is simply to extract this in a seperate component, it will only be loaded when the user hovers over the tooltip.
Just ensure you have a loading spinner inside this component :)

stijndehaes: So this doesn't work, let's write it down as an improvement for the future

stijn-janssens: #1958

PR #1957, opened 2025-10-31
Create an adr for the search implementation.

nclaeys:


docs/adr/0002-search-output-ports-implementation.md
```diff
@@ -0,0 +1,57 @@
...
+### Ranking function
+Use ts_rank_cd for scoring, as it considers term proximity (cover density).
+Alternatives like ts_rank lack this normalization, and a custom ranking function would be overkill at this stage.
+
+### Fields included
+Index only title and description fields of the output port and technical assets.
+These contain the most relevant data; excluding product and tag info keeps indexing simpler.
+
+### Query transformation
+Use websearch_to_tsquery for flexible multi-word queries at the cost of making query syntaxt more complex for users.
```
stijndehaes, line 37: Maybe just add some examples to the ADR?

docs/adr/0002-search-output-ports-implementation.md
```diff
@@ -0,0 +1,57 @@
...
+- B (=0.4): descriptions (output port + technical asset)
+
+### Indexing strategy
+Use a GIN index on a specific search column of type tsvector.
+Reindex automatically when any indexed field changes.
+Periodic reindexing is an alternative but unnecessary for our current dataset size.
+I would prefer to have the most up-to-date search experience.
+
+### Fuzzy search
+Not included in this version. May be added later to handle typos and near matches.
```
stijndehaes, line 61: How hard is this to add in postgres?

nclaeys: It requires a custom extension (pgtrgm). Not sure about the performance yet and how to incorporate it in the search query but it is possible. Main work is in testing it, not the amount of code.

nclaeys:
> LGTM, only thing I'm thinking about while reading is if it would make sense to index the domain as well

Sure that is possible. Would you want me to index the name and description for the domain with the datasets?

docs/adr/0002-search-output-ports-implementation.md
nclaeys: I can do that: I will add a link to the postgres docs and write the rules down.

The rules are:
- unquoted text: text not inside quote marks will be converted to terms separated by & operators, as if processed by plainto_tsquery.
- "quoted text": text inside quote marks will be converted to terms separated by <-> operators, as if processed by phraseto_tsquery.
- OR: the word “or” will be converted to the | operator.
- -: a dash will be converted to the ! operator.

PR #1959, opened 2025-10-31
Add simple search vector

nclaeys:


backend/app/data_outputs/service.py
```diff
@@ -121,14 +122,21 @@ def remove_data_output(self, id: UUID) -> DataOutputModel:
         result = copy.deepcopy(data_output)
         self.db.delete(data_output)
         self.db.commit()
+        self.update_search_vector_associated_datasets(result)
         return result
 
+    def update_search_vector_associated_datasets(self, result: DataOutputModel):
+        dataset_service = DatasetService(self.db)
+        for dataset_link in result.dataset_links:
+            dataset_service.recalculate_search_vector_for(dataset_link.dataset_id)
```
stijndehaes, line 137: Maybe don't db.commit() in this function, but only in the end after everything?

backend/app/data_outputs/service.py
```diff
@@ -121,14 +122,21 @@ def remove_data_output(self, id: UUID) -> DataOutputModel:
         result = copy.deepcopy(data_output)
         self.db.delete(data_output)
         self.db.commit()
```
stijndehaes: We could remove the self.db.commit(), and handle that in the `update_search_vector_associated_datasets` function?

backend/app/datasets/router.py
```diff
@@ -52,6 +52,16 @@ def get_datasets(
     return DatasetService(db).get_datasets(user)
 
 
+@router.get("/search")
+def search_datasets(
```
stijndehaes, line 56: We already had a :

```
@router.get("")
def get_datasets(
```

Why not expand that one?

docs/adr/0002-search-output-ports-implementation.md
```diff
@@ -0,0 +1,58 @@
+# Postgres full text search (FTS) for searching output ports
```
stijndehaes, line 1: ADR is part of other PR right?

stijndehaes (changes requested):

backend/app/data_outputs/service.py
```diff
@@ -111,7 +112,7 @@ def create_data_output(
         return model
 
     def remove_data_output(self, id: UUID) -> DataOutputModel:
-        data_output = self.db.get(DataOutputModel, id)
+        data_output: DataOutputModel | None = self.db.get(DataOutputModel, id)
```
stijn-janssens: this could be wrapped in the `ensure_exists` call

stijn-janssens: don't you need to explicitly load the `dataset_links` to avoid the raises for not explicit joins? (Did not test this myself)

frontend/src/types/dataset/datasets-search.contract.ts
```diff
@@ -0,0 +1,10 @@
+import type { DatasetContract } from '@/types/dataset';
+import type { TagContract } from '@/types/tag';
+
+export type DatasetsSearchContract = (Omit<DatasetContract, 'data_product_links' | 'owners' | 'tags'> & {
```
stijn-janssens: Shouldn't this be the DatasetsGetContractSingle with added rank?

docs/adr/0002-search-output-ports-implementation.md
nclaeys: Yes, splitted them up to first discuss the adr.

backend/app/data_outputs/service.py
nclaeys: I prefer to keep them separate. This is actual functionality, the update of the search vector is just something to run after all the rest was successful.

backend/app/datasets/router.py
nclaeys: Just following REST best practices to have list and search calls separate. They are not doing the same thing.

backend/app/data_outputs/service.py
nclaeys: I didn't because it makes it more difficult to reason about when to do the commit. Now I just do it after updating every index. The recalculate_search_vector_for is used in multiple places. What is your reasoning for moving it out?

backend/app/data_outputs/service.py
nclaeys: I updated to use ensure_exists. I did not get that error, let me try again.

backend/app/data_outputs/service.py
stijndehaes: Efficiency, I believe that FastAPI even automatically commits at the end of your call.
Every call in FastAPI already runs in a transaction automatically, that is commited at the end. So probably it's not even needed.
You could use a flush instead, if you need the result in another query

backend/app/datasets/router.py
stijndehaes: Well REST best practices are very vague, and in my opinion searching is just filtering items from a list.
So it's just applying a filter on your list, that doesn't really need a seperate call. 

I would rather only have one way to fetch datasets to just make the API usage easier.

backend/app/data_outputs/service.py
stijn-janssens: Update: I moved this commit out of this function and replaced it in the parent functions with a flush, refresh search and only then commit. Committing in multiple places caused an invalidate of the object which resulted in a refetch without explicit joined loading and therefor raised exceptions

backend/app/datasets/service.py
```diff
@@ -90,7 +111,58 @@ def get_dataset(self, id: UUID, user: UserModel) -> DatasetGet:
...
+        result = self.db.execute(sql, {"dataset_id": dataset_id})
+        return result.rowcount
+
+    def recalculate_search_vector_datasets(self) -> int:
+        if settings.SEARCH_INDEXING_DISABLED:
+            logger.debug("Search indexing is disabled, skipping recalculation")
+            return 0
+        sql = recalculate_search_vector_datasets_statement()
+        result = self.db.execute(sql)
+        self.db.commit()
```
stijn-janssens, line 161: this function is only called explicitly in the seeding script so I allow the commit here

backend/app/data_outputs/service.py
stijndehaes: This has been changed because of issues

backend/app/datasets/router.py
stijndehaes: After some review, I agreeto keep it seperate since the functionallity is very different from just filtering in a list

backend/app/datasets/service.py
stijndehaes: Should we make it clear in the docs for the function that it should only be used in seeding?

PR #1965, opened 2025-11-03
Added justification to dataset access request

stijndehaes:
In support of: #1917

image
image
image
image


backend/app/database/alembic/README
```diff
@@ -1 +1,6 @@
-Generic single-database configuration.
+# Database migrations
+
+To create a new migration run:
+```bash
+poetry run alembic revision
```
stijn-janssens: I would suggest adding `-m "<message>"` which replaces the comment `empty message` on top of your migration script.

frontend/src/pages/dataset/components/dataset-tabs/data-product-tab/components/data-product-table/data-product-table-columns.tsx
```diff
@@ -62,21 +60,27 @@ export const getDatasetDataProductsColumns = ({
...
                                 status={getDecisionStatusBadgeStatus(status)}
                                 text={getDecisionStatusLabel(t, status)}
-                                className={styles.noSelect}
                             />
                         }
                     />
                 );
             },
-            width: '100%',
+            width: 300,
```
stijn-janssens: is this arbitrary? Does this scale on different screens?

stijndehaes: Arbitrary, not way to scale with different screens.
There is probably a better way, but since nothing is responsive at the moment I think it's fine.

backend/app/database/alembic/README
stijndehaes: Done

frontend/src/pages/dataset/components/dataset-tabs/data-product-tab/components/data-product-table/data-product-table-columns.tsx
stijndehaes: As @stijn-janssens pointed out, I can set this to a percentage of the widht. So I did that!

PR #1967, opened 2025-11-03
Design ADRs for tracking usage

pascal-knapen:
Contains 5 ADRs: 
- what charting library to use for the frontend
- db desing for quantitative usage data
- db design for qualitative usage data
- rest endpoint design
- ingestion design

### DIFF
--- a/docs/adr/0007-ingestion-usage-metrics.md
+++ b/docs/adr/0007-ingestion-usage-metrics.md
@@ -66,7 +66,7 @@ client.push_usage_stats(
 * **Good, because** Low SDK Maintenance: The SDK is just an HTTP wrapper.
 * **Medium, because** Producers still have to write/run the SQL.
 
-### Option 2: Advanced SDK
+### Option 3: Advanced SDK
 
 * **Good, because** Best DX (in theory): Producers just configure the SDK. We can enforce the correct query logic.
 * **Bad, because** SDK Complexity: We now own and maintain a complex client with snowflake-connector-python as a dependency.


docs/adr/0003-db-model-usage-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+* **Option 1: Store Raw Query Logs** Ingest every query event, potentially with an array of assets touched.
+* **Option 2: Aggregated Tables** Create two separate, specialized tables. One table aggregates query counts (ignoring assets) and the other aggregates asset quert counts per day.
+
+## Decision Outcome
+
+**Chosen option:** *Option 2: Aggregated Tables* Performance and efficiency as the main driver.
+
+### Confirmation
+
+#### Table 1: outpur_port_query_stats_daily
```
stijndehaes: This table is a duplicate of the data in table 2.

I would mention table 2 first, and then mention that this is an aggregate of that table. 
I maybe would even make this is a view, because what should we do when a technical assets is removed from an output port?

docs/adr/0004-db-model-usages-curated-queries.md
```diff
@@ -0,0 +1,45 @@
+# Database Model for Output Port Curated Queries
```
stijndehaes, line 1: Don't know if this needs to even be an ADR, the conclusion is to do things in line with how other things are done.
If the JSONB option was chosen an explanation was surely needed

docs/adr/0005-rest-endpoints-usage.md
```diff
@@ -0,0 +1,100 @@
...
+## Decision Drivers
+
+* Separation of Concerns: Clearly separate read-only, metadata management, and bulk ingestion endpoints.
+* Performance: Endpoints must be fast and lightweight.
+* Developer Experience: Endpoints must be logical and well-defined.
+
+## Considered Options
+
+* **Option 1: One Monolithic Endpoint** Create a single endpoint `/api/datasets/{id}/usage` that handles everything.
+* **Option 2: Separate Endpoints per Use Case** Create distinct endpoints for consumers (GET), producers (PUT/POST/DEL), and the ingestion pipeline.
```
stijndehaes: I prefer seperate endpoints per case/graph in the UI, when people do a filtering for example, you only need to fetch one graph set of data instead of everything.

docs/adr/0005-rest-endpoints-usage.md
```diff
@@ -0,0 +1,100 @@
...
+* Developer Experience: Endpoints must be logical and well-defined.
+
+## Considered Options
+
+* **Option 1: One Monolithic Endpoint** Create a single endpoint `/api/datasets/{id}/usage` that handles everything.
+* **Option 2: Separate Endpoints per Use Case** Create distinct endpoints for consumers (GET), producers (PUT/POST/DEL), and the ingestion pipeline.  
+
+## Decision Outcome
+
+**Chosen option:** *Option X: Separate Endpoints per Use Case*. Clean and more maintainable.
```
stijndehaes:
```suggestion
**Chosen option:** *Option 2: Separate Endpoints per Use Case*. Clean and more maintainable.
```

docs/adr/0005-rest-endpoints-usage.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Most Popular Assets
+    * `GET /api/datasets/{id}/usage/popular-assets`
+    * Query Params: `time_range=90d (default).`
+
+#### Producer-Facing Endpoints (Metadata Management)
+
+1. Create new curated query
+    * `POST /api/datasets/{id}/usage/curated-queries`
+    * Payload: `{"title": "string", "description": "string", "query_text": "string", sort_order: "integer"}`
```
stijndehaes: We should probably improve this, adding sort order to this call is annoying, certainly with the put when you want to update a query and you want to change the order.

Switching number 2 and 3 requires 3 calls, changing number 2 to 4, 3 to 2, and 4, to 3.
After call one you ordering is also weird.

I would propose the following:
```
Update all curated queries
* `PUT /api/datasets/{id}/usage/curated-queries`
* Payload: `{"curated_queries: [{"title": "string", "description": "string", "query_text": "string"}]}`
```

This way with one put you update all queries, and change the order.

Since I don't expect people to put 100 queries here this way of working is fine.
It also replaces everything with 1 call

docs/adr/0006-ingestion-usage-metrics.md
```diff
@@ -0,0 +1,72 @@
...
+* Reliability: The pipeline must be robust (retries, error handling).
+* Separation of Concerns: We must define a clear boundary of responsibility between DPP and the producer's systems.
+* Atomic Pushes: The push for both query stats and asset stats should be treated as a single unit of work for a given day.
+
+## Considered Options
+
+* **Option 1: Direct REST Endpoint Push** Producers use the REST endpoint directly in their pipeline. We provide the API endpoint and documentation. Producers are responsible for implementing all HTTP logic: authentication, request formatting, JSON serialization, and error/retry handling. They are also 100% responsible for generating the aggregation queries.
+* **Option 2: Fluent Python SDK (Simple API Wrapper)** We offer an SDK to give a more friendly interface to push this data. We provide a lightweight Python SDK (data-portal-sdk). This SDK handles all API complexity: authentication, request formatting, and retries. Producers are still responsible for running their own aggregation queries (we provide SQL snippets), but they simply pass the resulting list of dictionaries to a single SDK method.
```
stijndehaes, line 18: We can add to this, that we want to generate it from openapi spec in the end, but we can start with a simple single method

docs/adr/0003-db-model-usage-metrics.md
stijndehaes: As discussed repeat why this needs to be a seperate table, and deleting a technical asset will not change this data since this is historical and still relevant

docs/adr/0006-ingestion-usage-metrics.md
```diff
@@ -0,0 +1,72 @@
...
+* **Option 3: Fluent Python SDK (Full Snowflake Integration)** We offer an advanced SDK that provides a friendly interface to push data, an interface on how to gather data, and a Snowflake implementation of that interface. We provide an "advanced" SDK that includes snowflake-connector-python as a dependency. The SDK defines an interface (e.g., an abstract base class) for data gathering. We ship a concrete SnowflakeUsageCollector implementation. Producers would instantiate this collector, (potentially) pass in their Snowflake credentials, and the SDK would be responsible for both running the aggregation queries and pushing the data.
```
stijndehaes, line 19: SDK Project:
- basic SDK (generated + auth support)
- Usage interface
- usage implementation:
  - Snowflake

stijndehaes: For me this is preffered

docs/adr/0005-rest-endpoints-usage.md
```diff
@@ -0,0 +1,100 @@
...
+
+1. Delete an existing curated query
+    * `DELETE /api/datasets/{id}/usage/curated-queries/{curated_query_id}`
+    * Response (204): No Content.
+
+#### Ingestion Endpoint (Bulk Ingestion)
+
+Push daily stats
+
+* `POST /api/datasets/{id}/usage/ingest`
```
stijndehaes, line 58: We will have to put a limit on here, or at least document the limit on post bodies. (since a load balancer will always have a limit)
We can also fix this in the SDK, with retries and auto splitting.

docs/adr/0003-db-model-usage-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+
+#### Table 2: tech_asset_stats_daily
+
+Purpose: To power the "Most popular assets" chart.
+
+Schema:
+* date (DATE, PK)
+* output_port_id (UUID, PK, FK)
+* tech_asset_id (UUID, PK, FK)
+* consumer_id (UUID, PK, FK)
```
stijndehaes: Link to another data_product

docs/adr/0002-charting-library.md
stijn-janssens, line 1: minor comment but adr 0002 has already been created

docs/adr/0004-db-model-usages-curated-queries.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Decision Drivers
+
+* Relational Integrity: Curated queries must be directly and clearly associated with a single ouput port.
+
+* Write Simplicity: CREATE, UPDATE, and DELETE operations (from the producer-facing UI) should be simple, atomic, and efficient.
+
+## Considered Options
+
+* **Option 1: New output_port_curated_queries Table** Create a new, dedicated table with a foreign key to the data_products table. Each row represents one curated query.
```
stijn-janssens: Foreign key towards the output_ports table? (Currently datasets table)

docs/adr/0003-db-model-usage-metrics.md
pascal-knapen: Updated

docs/adr/0004-db-model-usages-curated-queries.md
pascal-knapen: Good catch

docs/adr/0007-ingestion-usage-metrics.md
```diff
@@ -0,0 +1,74 @@
...
+* **Bad, because** High implementation burden on producers. They must all individually implement auth, error handling, and retries. Very error-prone.
+
+### Option 2: Simple SDK
+
+* **Good, because** Good DX: Simple for producers
+* **Good, because** Clear Separation of Concerns: We own the API client (transmission); producers own their data (generation).
+* **Good, because** Low SDK Maintenance: The SDK is just an HTTP wrapper.
+* **Medium, because** Producers still have to write/run the SQL.
+
+### Option 2: Advanced SDK
```
jlteunissen: Option 3

docs/adr/0007-ingestion-usage-metrics.md
```diff
@@ -0,0 +1,74 @@
...
+
+* **Good, because** Best DX (in theory): Producers just configure the SDK. We can enforce the correct query logic.
+* **Bad, because** SDK Complexity: We now own and maintain a complex client with snowflake-connector-python as a dependency.
+* **Bad, because** Dependency/Security Risk: We must manage the Snowflake driver version and may have to handle producer credentials.
+* **Bad, because** Blurs Responsibility: We become responsible if their queries fail.
```
jlteunissen, line 74: I would indeed try to stay away as much as possible from making client-specific or tool-specific integrations. For an open source tool it always scares me off when only a few integrations exists.

docs/adr/0004-db-model-usage-metrics.md
```diff
@@ -0,0 +1,85 @@
...
+* Consumer (C1, OP1) query count: 1
+* Consumer (C2, OP1) query count: 1
+* Asset (orders) query count: 2
+* Asset (customers) query count: 1
+
+To summarise:
+* Asset query counts cannot be used to calculate consumer query counts as this would lead to incorrect aggregations
+* We want to be able to delete assets and their stats without impacting hystorical consumer query counts
+
+#### Table 1: outpur_port_query_stats_daily
```
stijndehaes:
```suggestion
#### Table 1: output_port_query_stats_daily
```

docs/adr/0007-ingestion-usage-metrics.md
stijndehaes:
```suggestion
### Option 3: Advanced SDK
```

PR #1983, opened 2025-11-04
Feature/rework pending request component

stijn-janssens:


frontend/src/pages/home/components/pending-requests-inbox/pending-requests-inbox.tsx
```diff
@@ -232,6 +271,12 @@ export function PendingRequestsInbox() {
                     action.pending_action_type === PendingActionTypes.DataProductDataset
                     ? datasetColor
                     : dataProductColor,
+                handleAcceptDataProductDatasetLink,
```
stijndehaes: Instead of passing these functions, can't we just use:
```
    const {
        handleAcceptDataProductDatasetLink,
        handleRejectDataProductDatasetLink,
        handleAcceptDataOutputDatasetLink,
        handleRejectDataOutputDatasetLink,
        handleGrantAccessToDataProduct,
        handleDenyAccessToDataProduct,
    } = usePendingActionHandlers();
```
inside of the `createPendingItem` action? I assume we can't because it's not a component?

frontend/src/pages/home/components/pending-requests-inbox/pending-requests-inbox.tsx
stijndehaes: We can migrate this and the function `slicedPendingActionItems` into one, they even apply the same sorting

frontend/src/pages/home/components/pending-requests-inbox/pending-requests-inbox.tsx
stijndehaes: Maybe we should remove the pending item function completely and move it inside of the `PendingRequestsList` the items aren't used anywhere else, so I don't really understand why they are currently passed

stijn-janssens: the items are used in the pagination, the badge, the tab changes reflect the filtering, ...
Moving them into the list does not seem very feasible at first sight. Which also means moving the usePendingActionHandlers is not easy either.

frontend/src/pages/home/components/pending-requests-inbox/pending-request-item.tsx
```diff
@@ -0,0 +1,254 @@
...
+            case PendingActionTypes.DataProductRoleAssignment:
+                handleDenyAccessToDataProduct({
+                    assignment_id: pendingAction.id,
+                    data_product_id: pendingAction.data_product.id,
+                });
+                break;
+        }
+    };
+
+    const getContent = () => {
```
stijndehaes: I would put this in a useMemo, to ensure this doesn't always get recalculated

frontend/src/pages/home/components/pending-requests-inbox/pending-request-item.tsx
```diff
@@ -0,0 +1,254 @@
...
+        const isDatasetType =
+            pendingAction.pending_action_type === PendingActionTypes.DataOutputDataset ||
+            pendingAction.pending_action_type === PendingActionTypes.DataProductDataset;
+        const color = isDatasetType ? datasetColor : dataProductColor;
```
stijndehaes: I think instead of doing this, you can just set the color in the switch case?

PR #1984, opened 2025-11-04
Fix biome warnings and make them error

### LINKED ISSUE #1982: Biome warnings should fail unit tests
## Summary
Biome warnings currently silently fail while running pre-commit hooks. They are visible but don't give errors.
However the project is doing fairly well with respect to those warnings so they should fail the hooks or a unit test instead. So we ensure we keep the warnings at 0.

### Motivation
Better code quality

stijn-janssens:
Closes #1982


frontend/src/pages/cart/cart.page.tsx
```diff
@@ -50,7 +50,7 @@ export function Cart() {
         if (savedData) {
             form.setFieldsValue(JSON.parse(savedData));
         }
-    }, []);
+    }, [form]);
```
stijndehaes: For this one I am unsure if it does what we want. Maybe the use of useEffect is also incorrect here.
Now when the form changes we set the fields value from localstorage, this might result in a race condition with storing the values in localstorage.

Perhaps the correct fix is to use the `initialValues` field on the form instead of this

frontend/src/pages/cart/cart.page.tsx
```diff
@@ -39,7 +39,7 @@ export function Cart() {
 
     const currentUser = useSelector(selectCurrentUser);
     const { data: userDataProducts, isFetching: isFetchingUserDataProducts } = useGetUserDataProductsQuery(
-        currentUser!.id,
+        currentUser?.id || '',
```
vlieven: This should be `??` (nullish coalesc) instead of `||` (boolean OR)
Blog: https://ginny.dev/blog/understanding-ts-symbols

I don't think there will be a runtime difference for this piece of code, but it is more correct.

```suggestion
        currentUser?.id ?? '',
```

frontend/src/pages/cart/cart.page.tsx
```diff
@@ -78,7 +78,7 @@ export function Cart() {
     const selectedDataProductId = Form.useWatch('dataProductId', form);
 
     const { data: selectedDataProduct, refetch: refetchSelectedDataProduct } = useGetDataProductByIdQuery(
-        selectedDataProductId!,
+        selectedDataProductId || '',
```
vlieven:
```suggestion
        selectedDataProductId ?? '',
```

stijndehaes (approved):

PR #1993, opened 2025-11-05
Feature/1917 create new data product

stijndehaes:


frontend/src/components/data-products/data-product-form/data-product-form.component.tsx
```diff
@@ -99,6 +104,7 @@ export function DataProductForm({ mode, dataProductId }: Props) {
     const userSelectOptions = dataProductOwners.map((owner) => ({
         label: `${owner.first_name} ${owner.last_name} (${owner.email})`,
         value: owner.id,
+        disabled: owner.id === currentUser?.id,
```
stijn-janssens, line 107: I like this approach, but it does mean you can't allow cleaning of the box anymore because cleaning also removes the currentUser as owner and you can't readd it since it is disabled.
So either disable cleaning or make it custom so it does not remove current user

frontend/src/pages/cart/cart.page.tsx
```diff
@@ -67,24 +77,47 @@ export function Cart() {
...
+        return [
+            {
+                value: CREATE_NEW_DATA_PRODUCT,
+                label: t('Create new data product'),
+            },
+            ...options,
+        ];
+    }, [userDataProducts, t]);
     const selectedDataProductId = Form.useWatch('dataProductId', form);
+    console.log(selectedDataProductId);
```
stijn-janssens: remove console.log

frontend/src/pages/cart/cart.page.tsx
```diff
@@ -174,9 +208,12 @@ export function Cart() {
                                 placeholder={t('Select a data product')}
                                 options={dataProductOptions}
                                 loading={isFetchingUserDataProducts}
-                                filterOption={(input, option) =>
-                                    (option?.label ?? '').toLowerCase().includes(input.toLowerCase())
-                                }
+                                filterOption={(input, option) => {
+                                    if (option?.value === CREATE_NEW_DATA_PRODUCT) {
```
stijn-janssens: Personal opinion but I'm not a fan of having the create new inside of the select. Maybe a + icon is clearer?

frontend/src/routes.tsx
```diff
@@ -34,7 +34,11 @@ import { Settings } from './pages/settings/settings.page.tsx';
 const router = createBrowserRouter([
     {
         path: ApplicationPaths.Home,
-        element: <AuthLayout />,
+        element: (
+            <NuqsAdapter>
```
stijn-janssens, line 38: what does this do?

frontend/src/pages/cart/cart.page.tsx
stijn-janssens: At the very least I would make it stand out more. If somebody now would create a data product with the name "Create new data product" (very bad choice)  there would be no difference. So maybe add the + icon inside the label as well then. Could also be enough, not sure

frontend/src/routes.tsx
stijndehaes: It ensure nuqs work, it's a handy tool. We used something like it in Conveyor, but that package doesn't get updates anymore so decided to do something else.
Basically it allows you to use state, but also ensure that that state is safed as query parameters in the URL.

Typically used to for example store your search query:

```
https://duckduckgo.com/?q=I+am+searching&t=chromentp&ia=web
```
This is a duck duck go url, where I was searching the string `I am searching`, this allows me to easily share the URL with other.

I would like to enable this pattern in the front end, and could make good use of for this PR

frontend/src/components/data-products/data-product-form/data-product-form.component.tsx
stijndehaes: What this basically does is remove the x button, see screenshot.
image

Ah the box clean button is the x at the right, this indeed removes you current user, it shouldn't remove your current user I guess. I will fix that.
Or at least it shouldn't in create mode.

frontend/src/routes.tsx
stijn-janssens: so without this adapter the search parameters would not be fetchable from state across components?

stijn-janssens: answered my own question as you import those methods from "nuqs" :D

frontend/src/components/data-products/data-product-form/data-product-form.component.tsx
stijndehaes: I see the select is disabled in edit mode, sot hat is not a problem.

We can also remove the allowClear, I don't think this will be heavily used anyway

frontend/src/pages/cart/cart.page.tsx
stijndehaes: Reworked to this:
image (3)

This also simplifies the code quite a lot

frontend/src/components/data-products/data-product-form/data-product-form.component.tsx
stijn-janssens: fine for me

PR #1996, opened 2025-11-05
Move marketplace searchbar underneath title

stijndehaes:
Got feedback from users they found it strange.

Also removed some unneeded css, and replaced
it with flex gap

image


stijn-janssens:
not convinced I like this better :P

PR #1997, opened 2025-11-05
Consistent search pages

stijndehaes:
3 search pages, pagination is now always bottom right

image
image
image


stijn-janssens:
I am ok with the title being above everything else and the role filter at the same height of the searchbar.
I don't like the fact that the pagination creates an enormous white space on the left now.
Putting it at the bottom can solve that but then we need to ensure that it is visible on small screens (original reason why we put it on top) and that we maybe change pagination to be at the bottom throughout the whole application

PR #2029, opened 2025-11-10
Update pipelines to auto-release helm chart

stijn-janssens:

### DIFF
--- a/.github/workflows/docker-build-main.yml
+++ b/.github/workflows/docker-build-main.yml
@@ -111,6 +111,7 @@ jobs:
           repo_name="demo-portal-infrastructure"
           event_type="apply-portal-release"
           version="$(git rev-parse --short=7 HEAD)"
+          chart="$(git describe --tags --abbrev=0 | cut -c2-)"
 
           curl -L \
             -X POST \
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build-main.yml
```diff
@@ -111,11 +111,12 @@ jobs:
           repo_name="demo-portal-infrastructure"
           event_type="apply-portal-release"
           version="$(git rev-parse --short=7 HEAD)"
+          chart="$(git describe --tags --abbrev=0 | cut -c2-)"
```
stijndehaes, line 114: What is the difference between chart and version?

.github/workflows/release-workflow.yml
```diff
@@ -201,3 +201,43 @@ jobs:
         run: |
           npm run build
           aws s3 cp build/ s3://data-product-portal-user-docs-tfhgye/ --recursive
+  deploy-on-portal-demo:
```
stijndehaes, line 204: Doesn't it make more sense to do this on our internal infra github? Since this is not a public demo environment?

.github/workflows/docker-build-main.yml
stijn-janssens: charts are only updated at releases, versions are the image tags and are updated with each push to main

.github/workflows/release-workflow.yml
stijn-janssens: yes, this dispatches to that git repo

stijndehaes: I am clearly not awake :D

.github/workflows/docker-build-main.yml
stijndehaes: But if we use the image of main, doesn't it make more sense to also use the chart from main?
We can push the chart to the container registry too. Maybe an improvement for the future?

stijn-janssens: Possible improvement, but I wanted to limit the amounts of charts we pushed. No real reason, has also already hurt me in the past :P Open for suggestions

stijndehaes: If we push it to a container registry, and clean it up just like we do with the normal images it's fine no? :) 
I would like to add it for a future ticket:)

stijndehaes (approved):

.github/workflows/docker-build-main.yml
stijn-janssens: Made issue #2077

PR #2055, opened 2025-11-12
Add Jinja templating to database seed script

stijn-janssens:


backend/sample_data.sql
```diff
@@ -105,13 +105,13 @@ begin
     RETURNING id INTO admin_role_id;
 
     -- ENVIRONMENTS
-    INSERT INTO public.environments ("name", context, acronym, is_default, created_on, updated_on, deleted_at) VALUES ('development', 'arn:aws:iam::123456789123:role/{{}}-dev-demo', 'dev', true, timezone('utc'::text, CURRENT_TIMESTAMP), NULL, NULL) returning id INTO returned_environment_id_dev;
-    INSERT INTO public.environments ("name", context, acronym, is_default, created_on, updated_on, deleted_at) VALUES ('production', 'arn:aws:iam::123456789123:role/{{}}-prd-demo', 'prd', false, timezone('utc'::text, CURRENT_TIMESTAMP), NULL, NULL) returning id INTO returned_environment_id_prd;
+    INSERT INTO public.environments ("name", context, acronym, is_default, created_on, updated_on, deleted_at) VALUES ('development', 'arn:aws:iam::{{ AWS_ACCOUNT_ID }}:role/{{ AWS_ROLE_WITH_CONTEXT_TEMPLATE_DEV }}', 'dev', true, timezone('utc'::text, CURRENT_TIMESTAMP), NULL, NULL) returning id INTO returned_environment_id_dev;
```
stijndehaes, line 108: Currently when this environment variable in unset what happens? Maybe we need to have a default?

stijn-janssens: empty and will crash upon using. Which is kinda what you want?
Only used in the "to AWS" action tile

stijndehaes: I was just thinking it's annoying for development to set this up without defaults. But maybe we change the readme file to include some defaults for development?

stijn-janssens: The issue I have with defaults is that this functionality is impossible to have working with defaults. You need to have a certain setup in place in AWS to allow for role assumption and credential federation etc etc. Without defaults it will just crash a little bit earlier. With (only) defaults it will still crash :P
To me either is fine, if you have a strong opinion on defaults I will add them

stijndehaes: After a quick call we figured out where we misunderstood each other.
I was woried the database seeding would crash when you did not supply these variables, making development annoying. But apparantly jinja just fills in empty strings so there is no issue.

Just to be sure @stijn-janssens will quickly double check it does not crash.

stijndehaes (approved):

PR #2068, opened 2025-11-13
Bad UX behaviour with long descriptions

stijn-janssens:

Scherm­afbeelding 2025-11-13 om 14 22 31

### DIFF
--- a/frontend/src/components/datasets/dataset-card/dataset-card.component.tsx
+++ b/frontend/src/components/datasets/dataset-card/dataset-card.component.tsx
@@ -217,7 +217,10 @@ export function DatasetCard({ datasetId, draggedDataOutputId }: Props) {
                             <Typography.Paragraph
                                 style={{ marginBottom: 0 }}
                                 type="secondary"
-                                ellipsis={{ expandable: 'collapsible' }}
+                                ellipsis={{
+                                    expandable: 'collapsible',
+                                    symbol: (expanded: boolean) => (expanded ? t('less') : t('more')),
+                                }}
                             >
                                 {dataset.description}
                             </Typography.Paragraph>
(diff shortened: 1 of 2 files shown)


frontend/src/components/datasets/dataset-card/dataset-card.component.tsx
```diff
@@ -214,9 +214,13 @@ export function DatasetCard({ datasetId, draggedDataOutputId }: Props) {
                                     </Popconfirm>
                                 </Flex>
                             </Flex>
-                            <Typography.Text type="secondary" ellipsis={{ tooltip: true }}>
+                            <Typography.Paragraph
+                                style={{ marginBottom: 0 }}
+                                type="secondary"
+                                ellipsis={{ expandable: 'collapsible' }}
```
stijndehaes: In the Justification component: `src/components/data-products/data-product-dataset-justification/justification.component.tsx`, I changed the symbol: `symbol: t('more'),` also to ensure translation, best if we do the same here?

stijndehaes (changes requested):

frontend/src/components/datasets/dataset-card/dataset-card.component.tsx
stijn-janssens: should we make a shareable elipsis component?

stijndehaes: Either that, or try to figure out how we change the default name in ant design. But I have no clue how to do that at the moment. Maybe spend 15min searching? If not found by then let's make a small component

stijndehaes: Decided to for now not make a new component, and keep it this way. If we have this pattern more often we will investigate a fix then

PR #2078, opened 2025-11-14
Pre fill output owners

stijn-janssens:

### DIFF
--- a/frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
+++ b/frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
@@ -53,6 +53,7 @@ import {
     createDatasetIdPath,
 } from '@/types/navigation.ts';
 import { getDatasetAccessTypeLabel } from '@/utils/access-type.helper.ts';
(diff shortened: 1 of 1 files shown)


frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -281,8 +282,9 @@ export function DatasetForm({ mode, modalCallbackOnSubmit, formRef, datasetId, d
         (namespace: string) => validateNamespace(namespace).unwrap(),
         [validateNamespace],
     );
-
-    const ownerIds = useGetDatasetOwnerIds(currentDataset?.id);
+    const datasetOwners = useGetDatasetOwnerIds(currentDataset?.id);
```
stijndehaes, line 285: In create mode there is no currentDataset ID right? Shouldn't this request be skipped in that case?

stijn-janssens: happens underneath indeed. As you can see in the previous state of the code this request would also be skipped at that point and not in this component

stijndehaes (approved):

PR #2107, opened 2025-11-19
API redesign ADR

stijndehaes:

### DIFF
--- a/docs/adr/0008-api-redesign.md
+++ b/docs/adr/0008-api-redesign.md
@@ -1,0 +1,185 @@
+# ADR Title
+
+## Context and Problem Statement
+
+The current API's don't match the naming of output ports and technical assets.
+Also because of [ADR 0001 - Datasets Belong to a Single Data Product](0001-datasets-single-parent.md) output ports
+are always part of a single data product.
+The API's are currently also split up quite a lot so there are opportunities to fix this
+
+
+## Decision Drivers
+
+* Outport, and technical asset rename
+* [ADR 0001 - Datasets Belong to a Single Data Product](0001-datasets-single-parent.md)
+* Api's very granular
+
+## Considered Options
+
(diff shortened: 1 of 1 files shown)


docs/adr/0008-api-redesign.md
```diff
@@ -0,0 +1,179 @@
...
+data_product_settings
+domains
+environments
+platforms
+tags
+theme_settings
+
+#### Namespace methods
+
+The namespace methods currently help to check if a namespace is valid for a cloud.
```
vlieven, line 97: I prefer resource_name instead of slug (I had never heard of the term)

docs/adr/0008-api-redesign.md
```diff
@@ -0,0 +1,179 @@
...
+move to users current user
+
+#### events: /api/events/latest
+
+Unused. Idea was to have a lambda call this to see if events have happened to update the infra.
+We can currently remove it
+
+#### auth: /api/auth
+
+Renamed to /api/v2/auth
```
vlieven: Should this maybe be `authn` to have the distinction with `authz`? Or is that not typically done?

docs/adr/0008-api-redesign.md
```diff
@@ -0,0 +1,179 @@
...
+platforms
+tags
+theme_settings
+
+#### Namespace methods
+
+The namespace methods currently help to check if a namespace is valid for a cloud.
+
+I would suggest to rename this to either:
+- resource_name, with the methods
```
stijn-janssens: out of these I prefer resource name but that is just personal taste

docs/adr/0008-api-redesign.md
stijndehaes: I wouldn't mind to be honest, it makes it more clear to me. @stijn-janssens was neither a for or against it.

So let's change it?

Issue #2120, opened 2025-11-20
Usage ADR implementation


stijndehaes:
Implement the usage ADRS:
https://github.com/conveyordata/data-product-portal/pull/1967

- [ ] Implement output_port_query_stats_daily, and implement the queries over time graphs
- [ ] Implement curated queries

PR #2123, opened 2025-11-21
Feature/2120 output port usage daily: first end-to-end flow

jlteunissen:
- This is a first minimal PR for issue: https://github.com/conveyordata/data-product-portal/issues/2120 
- This PR contains a first end-to-end flow of data flowing through database->backend->frontend->antv-graph according to:
  - ADR0003: using Ant Design Charts (AntV) as charting library
  - ADR0004: create table output_port_query_stats_daily. For now still dataset_query_stats_daily as rename will be consistently done later. 
  - ADR0006: create API endpoints. 
    - PATCH: /api/datasets/{id}/usage/time-series. This has been implemented as a PATCH instead of a POST as typically we want to upsert data. 
    - GET: /api/datasets/{id}/usage/time-series
    
 result ([local-dev-link](http://localhost:3000/datasets/bdad0dec-19e6-4c85-a655-0960ca3a484c#usage))

image

next TODOs:
- add granularity in API and graph
- Implement output_port_query_stats_daily
- Implement curated queries
- Improve the quest_states_daily inserted values in, samples data (use data today -X, and generate more values)
- Query daily stats graph, only show top x consumers, show the rest in others bucket
- When no usage is available put link to documentation
- Write documentation

### DIFF
--- a/backend/tests/app/datasets/query_stats_daily/test_service.py
+++ b/backend/tests/app/datasets/query_stats_daily/test_service.py
@@ -2,11 +2,6 @@ from datetime import date, timedelta
 
 import pytest
 from sqlalchemy.orm import Session
-from tests.factories import (
-    DataProductFactory,
-    DatasetFactory,
-    DatasetQueryStatsDailyFactory,
-)
 
 from app.datasets.query_stats_daily.model import DatasetQueryStatsDaily
 from app.datasets.query_stats_daily.schema_request import (
@@ -14,6 +9,11 @@ from app.datasets.query_stats_daily.schema_request import (
     DatasetQueryStatsDailyUpdate,
 )
 from app.datasets.query_stats_daily.service import DatasetQueryStatsDailyService
+from tests.factories import (
+    DataProductFactory,
+    DatasetFactory,
+    DatasetQueryStatsDailyFactory,
+)
 
 
 @pytest.fixture
--- a/backend/app/datasets/query_stats_daily/router.py
+++ b/backend/app/datasets/query_stats_daily/router.py
@@ -16,7 +16,7 @@ from app.datasets.query_stats_daily.service import DatasetQueryStatsDailyService
 router = APIRouter(prefix="/{id}/query_stats", tags=["datasets"])
 
 
-@router.get("", response_model=DatasetQueryStatsDailyResponses)
+@router.get("")
 def get_query_stats(
     id: UUID,
     db: Session = Depends(get_db_session),
--- a/backend/tests/app/datasets/query_stats_daily/test_router.py
+++ b/backend/tests/app/datasets/query_stats_daily/test_router.py
@@ -1,12 +1,11 @@
 from datetime import date, timedelta
 
+from app.datasets.query_stats_daily.model import DatasetQueryStatsDaily
 from tests.factories import (
     DataProductFactory,
     DatasetFactory,
     DatasetQueryStatsDailyFactory,
 )
-
-from app.datasets.query_stats_daily.model import DatasetQueryStatsDaily
 
 ENDPOINT = "/api/datasets"
 


backend/app/shared/router.py
```diff
@@ -51,3 +52,4 @@
 router.include_router(notification)
 router.include_router(pending_action)
 router.include_router(events)
+router.include_router(dataset_query_stats_daily)
```
stijndehaes: Either this is required or the `router.include_router(query_stats_daily_router)` in backend/app/datasets/router.py is required, but I don't think both are required?

backend/tests/app/datasets/query_stats_daily/test_service.py
```diff
@@ -0,0 +1,184 @@
...
+                date=today.isoformat(),
+                consumer_data_product_id=consumer.id,
+                query_count=100,
+            )
+        ]
+
+        service = DatasetQueryStatsDailyService(session)
+        service.update_query_stats_daily(dataset.id, updates)
+
+        stats = (
```
stijndehaes: You can make this a method, since it's basically the same function in the next test

frontend/src/pages/dataset/components/dataset-tabs/dataset-tabs.tsx
```diff
@@ -16,11 +16,12 @@ import { Explorer } from '@/components/explorer/explorer';
 import { HistoryTab } from '@/components/history/history-tab.tsx';
 import { DataOutputOutlined, DataProductOutlined } from '@/components/icons';
 import { LoadingSpinner } from '@/components/loading/loading-spinner/loading-spinner';
-import { UsageTab } from '@/components/tabs/usage-tab/usage-tab.tsx';
 import { PosthogEvents } from '@/constants/posthog.constants.ts';
 import { DataOutputTab } from '@/pages/dataset/components/dataset-tabs/data-output-tab/data-output-tab';
 import { DataProductTab } from '@/pages/dataset/components/dataset-tabs/data-product-tab/data-product-tab';
 import { TabKeys } from '@/pages/dataset/components/dataset-tabs/dataset-tabkeys';
+// import { UsageTab } from '@/components/tabs/usage-tab/usage-tab.tsx';
```
stijndehaes: Let's remove this line

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.tsx
```diff
@@ -0,0 +1,12 @@
+import { useGetDatasetQueryStatsDailyQuery } from '@/store/features/datasets/datasets-api-slice';
+import { UsageChart } from './usage-chart';
+
+type Props = {
+    datasetId: string;
+};
+
+export function UsageTab({ datasetId }: Props) {
+    const { data, isLoading } = useGetDatasetQueryStatsDailyQuery(datasetId);
```
stijndehaes, line 9: Instead of querying here and passing it, I would just query the data inside of the `UsageChart`

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
...
+    const now = new Date();
+
+    // transform date strings to Date objects
+    const processedData = responses
+        .map((stat) => {
+            const date = new Date(stat.date);
+            return {
+                timestamp: date.getTime(),
+                queryCount: stat.query_count,
+                consumer: stat.consumer_data_product_name || 'Unknown',
```
stijndehaes: Why the or with `Unknown`? 
If we keep it this should be trasnlated
```suggestion
                consumer: stat.consumer_data_product_name || t('Unknown'),
```

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
...
+            };
+        })
+        .filter((stat) => stat.timestamp >= oneMonthAgo.getTime())
```
stijndehaes, line 38: I guess we should filter this in the backend eventually?

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
...
+        .sort((a, b) => a.timestamp - b.timestamp);
+
+    // Get unique consumers
+    const consumers = [...new Set(processedData.map((d) => d.consumer))];
```
stijndehaes, line 42: Should we add the consumer list in the response?

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
...
+    while (currentDate <= now) {
+        const currentMonth = currentDate.getMonth();
+        const dateLabel = currentMonth !== lastMonth ? format(currentDate, 'MMM d') : format(currentDate, 'd');
+        lastMonth = currentMonth;
+
+        allDates.push({
+            date: dateLabel,
+            timestamp: currentDate.getTime(),
+        });
+        currentDate = new Date(currentDate.getTime() + 24 * 60 * 60 * 1000); // Add one day
```
stijndehaes: You can use `date-fns` for this:

```suggestion
import { subDays } from 'date-fns';
        currentDate = subDays(currentDate, 1)
```

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
...
+    // Get unique consumers
+    const consumers = [...new Set(processedData.map((d) => d.consumer))];
+
+    // Create a list of all dates in the range formatted as 'MMM d' or 'd'
+    const allDates: Array<{ date: string; timestamp: number }> = [];
+    let currentDate = new Date(oneMonthAgo);
+    let lastMonth = -1;
+    while (currentDate <= now) {
+        const currentMonth = currentDate.getMonth();
+        const dateLabel = currentMonth !== lastMonth ? format(currentDate, 'MMM d') : format(currentDate, 'd');
```
stijndehaes: You can probably use `date-fns` for this

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
+import { Area } from '@ant-design/charts';
```
stijndehaes, line 1: In geneneral all date magic can probably be simplified with: [date-fns](https://date-fns.org/docs/Getting-Started)

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,144 @@
...
+        legend: {
+            position: 'top-right' as const,
+        },
+    };
+
+    return (
+        <div style={{ padding: '24px' }}>
+            <h3>{t('Usage Statistics - Last Month')}</h3>
+            <Area {...config} />
+        </div>
```
stijndehaes:
```suggestion
        <div>
            <Typography.Title level={3}>{t('Usage Statistics - Last Month')}</Typography.Title>
            <Area {...config} />
        </div>
```

Typography.Title already has padding or margin at the bottom, so pading is not needed in div

backend/app/shared/router.py
jlteunissen: you are right, we already do this in datasets/router.py this one can be deleted.

backend/tests/app/datasets/query_stats_daily/test_service.py
jlteunissen: done. I also made the test more readable by using some fixture to prepare database entries

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.tsx
jlteunissen: I would prefer to do it here, as in the next PR I will add another (bar) chart that uses the same data, so I get easily give both charts the same data.

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: because we obtain the consumer_data_product_name from a join. I imagine their could be cases were the join doesn't find a matching id? Of course ideally this shouldn't occur, but when it happens, we don't need to fail hard on the usage graph?

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: No, I want the GET on the API to return just a array of elements very close to database table and don't do unrequested aggregations we can easily do here.

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: now moved to `currentDate = addDays(currentDate, 1);`

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: yes. I will do in next PR see comment below

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: Yes you are right. In the next PR I am adding a lot of extra date logic because of adding the time_range and granularity functionality so I will convert to use date-fns as much as possible there.

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: I changed it to 
```
        <Flex vertical>
            <Typography.Title level={3}>{t('Usage Statistics - Last Month')}</Typography.Title>
            <Area {...config} />
        </Flex>
 ```

File: frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx (hunk 6 above)
Comment (jlteunissen) [reply frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx:38]: indeed

File: frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx (hunk 5 above)
Comment (stijndehaes) [reply frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx]: That is reasonable thanks for the explanation :) 
We do need to translate the unknown still though

Comment (jlteunissen) [reply frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx]: That is done, although React hooks like useTranslation (which gives you t) are only allowed inside React components instead of regular utility functions so I am passing a translated string now

**Review by stijndehaes — APPROVED** [review]

File: backend/app/datasets/query_stats_daily/model.py [hunk 12]
```diff
@@ -0,0 +1,28 @@
...
+    )
+    consumer_data_product_id: Mapped[UUID] = mapped_column(
+        ForeignKey("data_products.id"), primary_key=True, index=True
+    )
+    query_count: Mapped[int] = mapped_column(Integer, default=0)
+
+    # Computed field: consumer_data_product_name from join
+    consumer_data_product_name = column_property(
+        select(Column("name", String))
+        .select_from(Base.metadata.tables["data_products"])
```
stijn-janssens: Is this the cleanest way to do this? Can't this be done through imports from `data_products.model`?
I am weary of the hardcoded strings here

backend/app/datasets/query_stats_daily/schema_response.py
```diff
@@ -0,0 +1,15 @@
...
+from uuid import UUID
+
+from app.shared.schema import ORMModel
+
+
+class DatasetQueryStatsDailyResponse(ORMModel):
+    date: date
+    consumer_data_product_id: UUID
+    query_count: int
+    consumer_data_product_name: str | None = None
```
stijn-janssens: please use `Optional[str]`

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    const chartData = useMemo(() => {
+        if (!data?.dataset_query_stats_daily_responses) {
+            return [];
+        }
+        return transformDataForChart(data.dataset_query_stats_daily_responses, t('Unknown'));
+    }, [data, t]);
+
+    if (isLoading) {
+        return <Spin size="large" style={{ display: 'flex', justifyContent: 'center', marginTop: '50px' }} />;
```
stijn-janssens: I prefer to have styles external in .scss files and include via Classnames, that way you can also reuse margin / padding variables across the application. But consider this a minor and not a breaking comment

frontend/src/store/features/datasets/datasets-api-slice.ts
```diff
@@ -152,6 +153,12 @@ export const datasetsApiSlice = baseApiSlice.enhanceEndpoints({ addTagTypes: dat
                 method: 'GET',
             }),
         }),
+        getDatasetQueryStatsDaily: builder.query<DatasetQueryStatsDailyResponses, string>({
```
stijn-janssens, line 171: should these queries provide a form of tag such that the request can be cached or do we not want to do that here?

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
stijndehaes: This could be replaced with:
```suggestion
        return (
           <Flex justify={"center"} align={"center"}>
               <Spin size="large"/>
           </Flex>
       );
```

This centers the spin in the midle of the screen, or you could use the `LoadingSpinner` which does this for you :) 

```suggestion
return <LoadingSpinner />
```

frontend/src/store/features/datasets/datasets-api-slice.ts
stijndehaes: Providing a tag is not necessary for caching as far as I know, it is required to invalidate the cache

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
stijn-janssens: Both are an improvement :)

backend/app/datasets/query_stats_daily/model.py
jlteunissen:
```[python]
    consumer_data_product_name = column_property(
        select(DataProduct.name)
        .where(DataProduct.id == consumer_data_product_id)
        .correlate_except(DataProduct)
        .scalar_subquery()
    )
```
this also seems to work

backend/app/datasets/query_stats_daily/schema_response.py
jlteunissen: done

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-chart.tsx
jlteunissen: `return <LoadingSpinner />` it is

stijn-janssens (approved):

Issue #2110, opened 2025-11-19
Implement API redesign ADR


stijndehaes:
https://github.com/conveyordata/data-product-portal/blob/main/docs/adr/0008-api-redesign.md

- [ ] Create the new API, while deprecating the old one

PR #2147, opened 2025-11-25
Migrate authorization

stijndehaes:
In support of: #2110


backend/app/core/authz/authorization.py
```diff
@@ -52,13 +51,13 @@ def enforce(
         resolver: type[SubjectResolver],
         *,
         object_id: str = "id",
-    ) -> Callable[[Request, User, Session], None]:
-        def inner(
+    ) -> Callable[[Request, User, Session], Awaitable[None]]:
+        async def inner(
```
stijn-janssens, line 55: why do you make it async? We used to have an async resolver but switched to sync

stijndehaes: Discussed in a call reason was that getting the request body is an async call.
I didn't switch casbin to async

PR #2165, opened 2025-11-26
Feature/improve explorer

stijn-janssens:
- Improves performance of explorer (x35 speedup)
- Removes some bugs of dangling edges
- Removes technical assets from explorer
- Adds datasets if they don't have technical assets yet to their data product owner


backend/app/data_products/service.py
```diff
@@ -638,6 +639,28 @@ def get_graph_data(self, id: UUID, level: int) -> Graph:
                                 )
                             )
 
+        # if no data outputs are present, still show the children datasets
+        if not data_outputs:
```
vlieven, line 643: We are deprecating the data_outputs from the graph right? Why are they still referenced here?

stijn-janssens: The explorers in the detail page still show technical assets. If you have output ports without technical assets they were not visible, this changes that.

vlieven (approved):

PR #2224, opened 2025-12-05
All users get systemaccount credentials

stijn-janssens:


backend/app/core/auth/auth.py
stijndehaes:
```suggestion
        if not api_key and not jwt_token::
```

I find this a bit more readable, I though the `if not api_key:` was covered by the statement here. 
Which obviously isn't true.

Maybe also not a bad idea to add a test to the function? Since this is pretty critical, and currently changing the behavior makes no test fail

vlieven: My first instinct if I see `not x and not y` would be to rewrite it as `not (x or y)` 😅

PR #2225, opened 2025-12-05
ADR for the output port lifecycle

vlieven:


docs/adr/0010-output-port-lifecycle.md
```diff
@@ -0,0 +1,86 @@
...
+Consumers that already have access can still view and access the output, but it cannot be requested by new Consumers, not even by invitation.
+
+#### Retired
+
+The data output has reached End-of-Life. By moving to this stage, all remaining consumers are removed (should be done with an explicit approval in case there are still consumers). At this point, only the team working on the parent Data Product can still view this Data Output. The team can now decide to remove the Data Output from portal altogether. Removal should not be possible in other lifecycle stages.
+
+### Constraints
+
+- Data Outputs can only be removed in the Retired stage, when they are guaranteed to no longer have consumers.
+- Data Outputs that are in draft (with consumers) can go directly to Retired if the producer does not want to move forward with publishing.
```
stijn-janssens: Can't we just allow draft projects to also be deleted?

docs/adr/0010-output-port-lifecycle.md
```diff
@@ -0,0 +1,86 @@
...
+- Once a Data Output has left the Draft phase to Active, it cannot move back.
+- Data Outputs can only reach Retired from Deprecated.
+- Data Outputs that are Deprecated can move back to Active should the product team decide so.
+- A Retired data output can move back to Draft should the product team decide so.
```
nclaeys: Do we need to add this option? I would guess that you can just as well create a new draft of your output?

stijn-janssens: Might be a lot of work to recreate a draft if you already added output ports and technical assets.
But then again, why would you have retired it?

docs/adr/0010-output-port-lifecycle.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Decision Drivers
+
+* Ease of use
+* Control for the Data Producer
+* Clarity for the Data Consumer
+
+## Considered Options
+
+* **Option 1: Simple toggle** A data output can be switched between 'draft' and 'published'.
```
stijndehaes:
```suggestion
* **Option 1: Simple toggle** An output port can be switched between 'draft' and 'published'.
```

Maybe best to change this everywhere

docs/adr/0010-output-port-lifecycle.md
```diff
@@ -0,0 +1,86 @@
...
+  Draft data outputs are not visible to people outside the data product, published outputs are visible.
+* **Option 2: State machine** A data output has four states: 'draft', 'active', 'deprecated' and 'retired'.
```
stijndehaes: I would add an option 3: A state machine for configurable lifecycles that uses the current lifecycles but adds a view strategy to them.
This allows our current users to just use the lifecycles they have, without us needing to figure out a migration from the configurable lifecycles to this.
They could for example have the following:
- `Draft`, with Draft view config
- `Request for Publishing`, with Draft view config, an architect needs to review and approve your output port
- `Published`, with Published view config
...

This allows customers to later add automation to the change from Draft to Request for Publishing, to automatically send and email. However we currently don't have that functionallity so the option could be overkill, and the migration from option 2 to this option 3 is relatively straight forward

stijndehaes: You added a fourth view option in your proposal, I like it

nclaeys: I agree it is a good 3rd option. I however think at the moment it feels overkill. I find it valuable that we are opinionated about this instead of leaving it open for customers. let's see what they say and indeed we might change it at some point.

docs/adr/0010-output-port-lifecycle.md
nclaeys: I agree, but feels to me it also felt strange that the output port was retired it in the first place. I do not see the usecase atm so I would not implement it yet, we can always change this later. @vlieven do you have a usecase for this?

docs/adr/0010-output-port-lifecycle.md
nclaeys: I changed that draft can be deleted. There are no consumers relying on this to work so we can safely delete it instead of the retirement step.

docs/adr/0010-output-port-lifecycle.md
nclaeys: changed it everywhere.

docs/adr/0010-output-port-lifecycle.md
nclaeys: Added the extra option

vlieven: The benefit of the state machine is that we can simply control the state transitions in our own code and which actions are permitted / connected to those. I.e. make the lifecycle into something that has functional meaning.

For example:
- Detaching all consumers upon retirement
- Allowing access by invite when in draft

If we want to make all of this configurable I fear things will become a mess.
I will clarify this in the options.

docs/adr/0010-output-port-lifecycle.md
vlieven: It is an escape hatch in case you change your mind to 'recover' a retired data output. 
From an implementation point it should be simple, and it might be a nice QoL thing for users that need it. Then again, it might indeed never be used.

The scenario I had envisioned was that you retire an output port, removing the consumers, and an important consumer that did not migrate is now broken (fault of the consumer, but still a problem). This would allow you to still recover from that.

docs/adr/0010-output-port-lifecycle.md
vlieven: The reason for the original design was to ensure the proper flow is followed. A draft output port can have consumers by invitation. Upon retirement, consumers are detached, and deletion afterwards is straightforward.
If you want to support this, you need extra logic to either:
- only allow deletion of output ports that don't have consumers, and redirect through retirement otherwise
- clear consumers on deletion as well, with the conditional message that consumers lose access by doing so

PR #2236, opened 2025-12-08
Feature/2120 usage api support granularity

jlteunissen:
- add granularity (day / week / month) to the daily stats usage API

### DIFF
--- a/backend/tests/app/datasets/query_stats_daily/test_service.py
+++ b/backend/tests/app/datasets/query_stats_daily/test_service.py
@@ -321,7 +321,7 @@ class TestDatasetQueryStatsDailyService:
 class TestDatasetQueryStatsDailyServicePrivateMethods:
     """Unit tests for private methods of DatasetQueryStatsDailyService."""
 
-    def test_consumer_totals_and_names(self, session: Session):
+    def test_consumer_totals(self, session: Session):
         """Test _consumer_totals_and_names aggregates totals."""
         service = DatasetQueryStatsDailyService(session)
         consumer1_id = uuid4()
@@ -348,7 +348,7 @@ class TestDatasetQueryStatsDailyServicePrivateMethods:
             ),
         ]
 
-        totals = service._consumer_totals_and_names(stats)
+        totals = service._consumer_totals(stats)
 
         assert totals[consumer1_id] == 150
         assert totals[consumer2_id] == 200


backend/app/datasets/query_stats_daily/router.py
```diff
@@ -11,18 +11,26 @@
...
+)
 
 router = APIRouter(prefix="/{id}/query_stats", tags=["datasets"])
 
 
 @router.get("")
 def get_query_stats(
     id: UUID,
+    granularity: QueryStatsGranularity = Query(default=QueryStatsGranularity.WEEK),
+    day_range: int = Query(default=DEFAULT_DAY_RANGE, ge=1),
```
stijndehaes, line 27: What does day range mean? Is it fetch for x days?

stijndehaes: I am fairly sure it does, for me it's ok if we also change this to an enum

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -16,22 +17,54 @@
...
         self.db = db
 
     def get_query_stats_daily(
-        self, dataset_id: UUID
+        self,
+        dataset_id: UUID,
+        granularity: QueryStatsGranularity = DEFAULT_GRANULARITY,
+        day_range: int = DEFAULT_DAY_RANGE,
     ) -> DatasetQueryStatsDailyResponses:
+        start_date = self._start_date_from_day_range(day_range)
```
stijndehaes: Maybe a bit overkill to make the method `_start_date_from_day_range` since it's only used here, and it's a single line?

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -16,22 +17,54 @@
...
+            )
+            .order_by(DatasetQueryStatsDaily.date.asc())
             .all()
         )
+        response_stats = [
+            DatasetQueryStatsDailyResponse.model_validate(stat) for stat in stats
+        ]
+
+        if granularity != QueryStatsGranularity.DAY:
+            response_stats = self._aggregate_by_granularity(response_stats, granularity)
```
stijndehaes: I am pretty sure we can do this aggregation in postgres no? 
If I aggregate by month I expect to see queries for the months october, november, december.
If I aggregate by week, we could just show the weeks of the year to.

Something like this should work:
```
SELECT 
    consumer_id,
    DATE_TRUNC('month', date) AS aggregation_month,
    SUM(query_count) AS total_query_count
FROM 
    data_sets_query_daily
GROUP BY 
    consumer_id, 
    DATE_TRUNC('month', date)
ORDER BY 
    consumer_id, 
    aggregation_month;
```

`DATE_TRUNC` can also work with the word `week` btw

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.tsx
```diff
@@ -20,12 +47,60 @@ export function UsageTab({ datasetId }: Props) {
...
+                    <Spin size="large" />
+                </Flex>
+            ) : !isLoadingState && !hasUsageData ? (
+                <Flex vertical className={styles.container} align="center" justify="center">
+                    <Empty description={t('No usage data available for this dataset')} />
+                </Flex>
+            ) : (
+                <Flex vertical className={styles.container}>
+                    <Flex className={styles.filters} gap={24} wrap>
+                        <Flex className={styles.filterGroup}>
```
stijndehaes: There is no other child in the outer flex, so fairly sure only one is needed

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.tsx
```diff
@@ -20,12 +47,60 @@ export function UsageTab({ datasetId }: Props) {
...
+                        </Flex>
+                    </Flex>
+                    <Flex className={styles.chartsGrid} gap={32}>
+                        <QueriesOverTimeChart data={chartData} isLoading={isLoadingState} hasData={hasUsageData} />
+                        <QueriesPerConsumerChart
+                            data={consumerTotals}
+                            isLoading={isLoadingState}
+                            hasData={hasUsageData}
+                        />
+                    </Flex>
```
stijndehaes: The following ensure we have the charts next to each to other with the exact same width. You can remove the flex and it's styling I believe
```suggestion
                    <Row gutter={[32, 32]}>
                        <Col span={12}>
                          <QueriesOverTimeChart data={chartData} isLoading={isLoadingState} hasData={hasUsageData} />
                        </Col>
                        <Col span={12}>
                          <QueriesPerConsumerChart
                              data={consumerTotals}
                              isLoading={isLoadingState}
                              hasData={hasUsageData}
                          />
                        </Col>
                    </Flex>
```

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.module.scss
```diff
@@ -0,0 +1,33 @@
+@use '@/styles/globals';
```
stijndehaes: Let's clean up this scss, fairly all of them can be removed.

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.tsx
```diff
@@ -1,16 +1,43 @@
...
+        return 'week';
+    }
+    return 'day';
+}
+
 export function UsageTab({ datasetId }: Props) {
-    const { data: usageData, isLoading: isUsageLoading } = useGetDatasetQueryStatsDailyQuery(datasetId, {
-        skip: !datasetId,
+    const { t } = useTranslation();
+    const longestDayRange = 365;
```
stijndehaes:
```suggestion
    const yearDayRange = 365;
```

frontend/src/pages/dataset/components/dataset-tabs/usage-tab/usage-tab.tsx
stijndehaes: We could even place the Row on the flex of line 75, and make the filters a `<Col span=24>` instead of flex

backend/app/datasets/query_stats_daily/router.py
jlteunissen: it is indeed a day range. I prefer having it as an integer. That way we can use the API to get whatever amount of time back we want.

backend/app/datasets/query_stats_daily/service.py
jlteunissen: good idea. I am now using
```[python]
            aggregated_query = (
                select(
                    func.cast(truncated_date, Date).label("date"),
                    DatasetQueryStatsDaily.consumer_data_product_id,
                    func.sum(DatasetQueryStatsDaily.query_count).label("query_count"),
                    DataProduct.name.label("consumer_data_product_name"),
                )
               ...
```
The bucketing is still done in the backend though

backend/app/datasets/query_stats_daily/service.py
jlteunissen: yes, that's indeed a bit overkill

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -16,22 +19,100 @@
...
+                .order_by(
+                    truncated_date.asc(),
+                    DataProduct.name.asc(),
+                    DatasetQueryStatsDaily.consumer_data_product_id.asc(),
+                )
+            )
+
+            results = self.db.execute(aggregated_query).all()
+            response_stats = [
+                DatasetQueryStatsDailyResponse(
```
stijndehaes: Can't you use `DatasetQueryStatsDailyResponse.model_validate(row)` here too?

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -74,3 +158,236 @@ def delete_query_stats_daily(
...
+        return sorted(
+            stats,
+            key=lambda stat: (
+                stat.date,
+                stat.consumer_data_product_name or EMPTY_STRING_FALLBACK,
+                str(stat.consumer_data_product_id),
+            ),
+        )
+
+    def _group_low_volume_consumers(
```
stijndehaes, line 190: This function is a bit of a duplicate of what the front end is doing to generate to one of the charts? Maybe we can just include it in the end result to only have to calculate this once?

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -74,3 +158,236 @@ def delete_query_stats_daily(
...
+        if len(consumer_totals) <= limit:
+            return stats
+
+        top_consumer_ids = self._top_consumer_ids(
+            consumer_totals, consumer_names, limit
+        )
+        grouped_stats = self._merge_other_consumers(stats, top_consumer_ids)
+        return self._sort_stats_by_date_and_consumer(grouped_stats)
+
+    def _consumer_totals_and_names(
```
stijndehaes: This could also be a query instead no? That might be faster, but I think it's ok to keep for now

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -74,3 +158,236 @@ def delete_query_stats_daily(
...
+    ) -> list[DatasetQueryStatsDailyResponse]:
+        if not stats:
+            return stats
+
+        consumer_totals, consumer_names = self._consumer_totals_and_names(stats)
+
+        if len(consumer_totals) <= limit:
+            return stats
+
+        top_consumer_ids = self._top_consumer_ids(
```
stijndehaes: Instread of sorting here, couldn't the function `_consumer_totals_and_names` not just return a sorted list?

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -74,3 +158,236 @@ def delete_query_stats_daily(
...
+        while current <= end:
+            buckets.append(current)
+            current = DatasetQueryStatsDailyService._increment_date(
+                current, granularity
+            )
+
+        return buckets
+
+    @staticmethod
+    def _align_to_granularity(value: date, granularity: QueryStatsGranularity) -> date:
```
stijndehaes: Does it make sense to implement this method on `QueryStatsGranularity` instead?

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -74,3 +158,236 @@ def delete_query_stats_daily(
...
+        if granularity == QueryStatsGranularity.DAY:
+            return value
+        if granularity == QueryStatsGranularity.WEEK:
+            return value - timedelta(days=value.weekday())
+        if granularity == QueryStatsGranularity.MONTH:
+            return value.replace(day=1)
+        return value
+
+    @staticmethod
+    def _increment_date(value: date, granularity: QueryStatsGranularity) -> date:
```
stijndehaes: Does it make sense to implement this method on `QueryStatsGranularity` instead?

backend/tests/app/datasets/query_stats_daily/test_service.py
```diff
@@ -184,6 +197,84 @@ def test_get_query_stats_daily(self, session: Session, dataset_with_daily_histor
...
+        service = DatasetQueryStatsDailyService(session)
+        response = service.get_query_stats_daily(
+            dataset.id,
+            granularity=QueryStatsGranularity.WEEK,
+            day_range=365,
+        )
+
+        stats = response.dataset_query_stats_daily_responses
+        # Verify buckets are filled (should have many entries due to bucket filling)
+        assert len(stats) > 1
```
stijndehaes: Instead of greather then can we verify the actual number?

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -74,3 +158,236 @@ def delete_query_stats_daily(
...
+        stats: list[DatasetQueryStatsDailyResponse],
+        granularity: QueryStatsGranularity,
+    ) -> dict[tuple[date, UUID], DatasetQueryStatsDailyResponse]:
+        """
+        Build a map of existing data with truncated dates.
+        Key: (bucket_date, consumer_id)
+        """
+        existing_data: dict[tuple[date, UUID], DatasetQueryStatsDailyResponse] = {}
+        for stat in stats:
+            bucket_date = self._align_to_granularity(stat.date, granularity)
```
stijndehaes: Shouldn't the result from the query already be aligned?

backend/tests/app/datasets/query_stats_daily/test_service.py
```diff
@@ -184,6 +197,84 @@ def test_get_query_stats_daily(self, session: Session, dataset_with_daily_histor
         assert stats_by_consumer[consumer1.id].date == today
         assert stats_by_consumer[consumer2.id].date == yesterday
```
stijndehaes, line 205: There is a lot of code added to the service, but not that many tests.

I was thinking that we can maybe unit test the following to make the testing a bit manageable:
- _consumer_totals_and_names
- _top_consumer_ids
- _merge_other_consumers 
- _fill_missing_buckets (probably one of the most important to test, because the logic isn't easy)

stijndehaes (changes requested):

backend/app/datasets/query_stats_daily/service.py
jlteunissen: Uhm, no its not a duplicate. This one is used when there are more than N=5 consumers, to group them into the 'Other' category. In the frontend we only have simple function to do a sum (aggregateQueriesPerConsumer) and to get the unique consumers.

backend/app/datasets/query_stats_daily/service.py
jlteunissen: Yes this process of getting the full series. i.e. if you have 
([(C1, day1, 10)], [(C1, day4, 20), (C2, day4, 30)]) 
transform it into something like:
([(C1, day1, 10), (C2, day1, 0)], [(C1, day2, 0), (C2, day2, 0)], [(C1, day3, 0), (C2, day3, 0)], [(C1, day4, 20), (C2, day4, 30)])
but this logic becomes also quite tedious in sql. You have to generate series, bucket them, etc. I asked the LLM and it looked awful, so I prefer to keep it in python because that is easier to understand.

backend/app/datasets/query_stats_daily/service.py
jlteunissen: no, because the function `_consumer_totals_and_names` is getting all the consumers. Next, if there are more than N=5 consumers, some of those are removed and grouped together in the "Other" category.

jlteunissen: I actually think we don't need the sorting at all in the backend. Because we do sort in the frontend.

backend/app/datasets/query_stats_daily/service.py
jlteunissen: Yes, that looks nice, and we dont need the @staticmethod. I also did it for increment_date because that function had the same call signature

backend/app/datasets/query_stats_daily/service.py
jlteunissen: I added the sorting back in, even more elaborately, such that the area chart puts the consumer with the biggest area on the bottom. Good thing: we don't need the sorting in the frontend anymore.

backend/app/datasets/query_stats_daily/service.py
jlteunissen: yes see next commetn

backend/tests/app/datasets/query_stats_daily/test_service.py
jlteunissen: yes we can, although for this one it is a bit tricky as the number of weeks depends here on the current day. So it has to be calculated. Which I do now.

backend/app/datasets/query_stats_daily/service.py
jlteunissen: yes, indeed, we now do it in the query already.

jlteunissen:
image
Notes:

- time range are now radio buttons
- granularity is automatically changed. day_range year -> automatically monthly, last 90 days -> automatically weekly, 30 days: granularity per day
- both charts now use the same consumer 2 color mapping
- other category is always put last
- timeseries are sorted with the data product with highest consumption put at the bottom (but other cat on top)
- aggregated bars are sorted with highest first (other category always at the bottom)

backend/app/datasets/query_stats_daily/service.py
jlteunissen: yes I do that now

backend/tests/app/datasets/query_stats_daily/test_service.py
jlteunissen: Yes test functions for all these private methods are now added

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -16,22 +20,123 @@
...
+        - WEEK: returns Monday of the week
+        - MONTH: returns the first day of the month
+        """
+        if self == QueryStatsGranularity.DAY:
+            return value
+        if self == QueryStatsGranularity.WEEK:
+            return value - timedelta(days=value.weekday())
+        if self == QueryStatsGranularity.MONTH:
+            return value.replace(day=1)
+        return value
```
stijndehaes:
```suggestion
        raise NotImplementedError(f"This is currently not implemented for {self}")
```

backend/app/datasets/query_stats_daily/service.py
```diff
@@ -16,22 +20,123 @@
...
+        Increment a date by one period according to the granularity.
+        """
+        if self == QueryStatsGranularity.WEEK:
+            return value + timedelta(weeks=1)
+        if self == QueryStatsGranularity.MONTH:
+            # Increment month, handling year rollover
+            if value.month == 12:
+                return value.replace(year=value.year + 1, month=1)
+            return value.replace(month=value.month + 1)
+        return value + timedelta(days=1)
```
stijndehaes: Make it similar to `align_date`

```suggestion
        if self == QueryStatsGranularity.DAY:
             return value + timedelta(days=1)
        raise NotImplementedError(f"This is currently not implemented for {self}")
```

backend/app/datasets/query_stats_daily/service.py
stijndehaes: Yeah sorry got confused because of the many methods :)

backend/app/datasets/query_stats_daily/service.py
stijndehaes: This method does not do anything with the day no? It just makes a dict of total per consumer ID, and the consumer names.

It probably could return a single dict to be honst:
```
dict[UUID, tuple[str | None, int]
```

With the UUID being the consumer id, str | None, the name, and int the total

Answer Jos: It doesn't do anything with the granularity, so that is removed now. The names was only used for a  unneeded second sort column, so it is removed

backend/app/datasets/query_stats_daily/service.py
jlteunissen: I will raise a valueerror now

backend/app/datasets/query_stats_daily/service.py
jlteunissen: same, will also raise a valueerror. I think the change of giving a value that is not implemented yet is not very probably, more likely there will be a spelling mistake so a valueerror seems more logical to me.

PR #2333, opened 2025-12-18
Refactor of datasets to output ports api

stijndehaes:
In support of: #2110


stijndehaes:
All datasets is currently deprecated without solution, we shall fix this when we use the new API in the front end

PR #2344, opened 2025-12-19
Migrate technical assets

stijndehaes:
In support of: #2110


backend/app/data_products/technical_assets/model.py
```diff
@@ -67,3 +68,18 @@ class DataOutput(Base, BaseORM):
         lazy="raise",
         viewonly=True,
     )
+
+
+def ensure_data_output_exists(
```
stijn-janssens, line 73: we don't want to rename this already to ensure_technical_asset_exists?

stijndehaes: I would rename this when we do the actual code, I know it's confusing for now though

PR #2350, opened 2025-12-22
Migrate namespace to resource names

stijn-janssens:

### DIFF
--- a/backend/app/configuration/data_product_settings/router.py
+++ b/backend/app/configuration/data_product_settings/router.py
@@ -24,6 +24,10 @@ from app.core.namespace.validation import (
     NamespaceValidation,
 )
 from app.database.database import get_db_session
+from app.resource_names.service import (
+    DataProductSettingResourceNameValidator,
+    ResourceNameService,
+)
 
 router = APIRouter()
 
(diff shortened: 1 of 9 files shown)


backend/app/configuration/data_product_settings/router.py
```diff
@@ -43,12 +43,12 @@ def create_data_product_setting(
     return DataProductSettingService(db).create_data_product_setting(setting)
 
 
-@router.get("/namespace_suggestion")
+@router.get("/namespace_suggestion", deprecated=True)
 def get_data_product_settings_namespace_suggestion(name: str) -> NamespaceSuggestion:
     return DataProductSettingService.data_product_settings_namespace_suggestion(name)
```
stijndehaes: Should we call the new `validate_resource_name` function here so we can get rid of the functions on the other services?

backend/tests/app/data_product_settings/test_router.py
```diff
@@ -135,12 +185,24 @@ def test_validate_namespace_duplicate(self, client):
         assert response.status_code == 200
         assert response.json()["validity"] == NamespaceValidityType.DUPLICATE_NAMESPACE
 
+    def test_validate_namespace_duplicate_dataset(self, client):
+        namespace = "test"
+        DataProductSettingFactory(
+            namespace=namespace, scope=DataProductSettingScope.DATASET
+        )
+        response = self.validate_namespace_output_port(
```
stijndehaes, line 193: Shpuld we place this test in a resources_names/test_router?

stijn-janssens: I left them in their original places, to keep the logic grouped per entity. If you want I can move it.

stijndehaes: I understand I tend to find it hard to navigate most of the tests since they are spread over the place instead of where the routes are defined.

I think it makes a lot more sense to place them next to the routes. But in this case it does spread it out a bit. So currently unsure to be honest, what would you do?

stijndehaes (approved):

PR #2376, opened 2025-12-30
Refactor data_outputs_datasets

stijndehaes:
Moved it under technical assets route
since there we already method there
that removed them.

But it feels a bit strange, certainly since
RBAC is on output port. So looking for feedback.
The route can be changed rather quickly

In support pf: #2110


stijn-janssens:
IMO a route like this feels more logical
`/data_products/{dpId}/output_ports/{opId}/technical_assets` because we add technical assets to output ports, not the other way around. But happy to discuss.

stijndehaes:
> IMO a route like this feels more logical `/data_products/{dpId}/output_ports/{opId}/technical_assets` because we add technical assets to output ports, not the other way around. But happy to discuss.

I agree, I would make the following routes:
- `/data_products/{dpId}/output_ports/{opId}/add_technical_assets`
- `/data_products/{dpId}/output_ports/{opId}/remove_technical_assets`
- `/data_products/{dpId}/output_ports/{opId}/approve_link_technical_assets_request`
- `/data_products/{dpId}/output_ports/{opId}/deny_link_technical_assets_request`

So we don't add the technical asset it to the route

stijn-janssens (approved):

PR #2409, opened 2026-01-07
Add ADR v1

stijn-janssens:


docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,92 @@
...
+
+---
+
+## Decision Outcome
+
+**Chosen option:** *Option: Python Plugin Architecture and Option 2: JSON Blobs*
+
+We introduce a plugin-based architecture where each data output integration is defined as a self-contained Python class. These classes:
+
+* Inherit from a common base (`DataOutputConfigurationPlugin`)
```
vlieven: We need a different name for this. We no longer have the notion of a `Data Output`, and it's a bit broader than configuration IMO. Maybe `AssetProviderPlugin`? Although that also is extremely generic...

vlieven (commented):
What I am still missing in this ADR (and don't have an answer for myself) is how we can allow users to load these plugin files into their portal deployment without needing to do a rebuild. Or is this out of scope of the ADR?

stijn-janssens:
Fair, I shall think about a proposal

vlieven:
Maybe an idea:
- We mount a volume to our backend container, users can put their files in a `plugins` directory.
- On startup of the API, we iterate over the files in that directory, and use `importlib` to load the files, and afterwards `inspect` to extract the actual subclasses.
- Once we have our list of classes, we can instantiate them.

We will have to think about this a bit to guard against duplicate classnames, bad inputs etc. But this could work I think?

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,111 @@
...
+* Allow for extendable architecture to introduce infrastructure provisioning later on.
+
+---
+
+## Considered Options
+
+* **Shared option: Python Plugin Architecture (Chosen)**
+  Define each integration as a Python class using a shared base class.
+  Frontend will be generated dynamically out of the provided fields (over REST API).
+  Optional: Add auto-registering of new plugin technologies.
```
stijndehaes: Airflow has some kind of auto regstritation flow for plugins, or used to have it. We can draw inspiration from there

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,111 @@
...
+
+* **Option 1:** Strict columns and fields for configurations in database
+  Current approach. Allows for easier migration of configurations because we know the content of the columns. Requires more dev-work and creation of migration script if extra columns are needed. Can introduce the risk of artificial (Wrong) mapping of certain technology columns onto our provided column names.
+
+* **Option 2:** Free column choice, JSON blobs in database. Makes migrations almost impossible because JSON content is arbitrary. Makes mapping easier and does not require migration script creation at plugin creation.
+
```
stijndehaes: But you might need a migration path at some point here too, we can support some kind of json migration

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,111 @@
...
+### Option 1:
+
+* **Good, because** Migrations are relatively easy
+* **Bad, because** If a developer wants other columns for their integrations a migration script is needed
+* **Bad, because** Artificial (wrong) mappings might be introduced to make technologies work without a new migration
+
+### Option 2:
+
+* **Good, because** No migrations are needed for new integrations
```
stijndehaes: Migrations might still be needed for upgrading plugins to new versions, also indexing becomes a lot more complex

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,111 @@
...
+### Option: Python Plugin Architecture
+
+* **Good, because** it is type-safe, extensible, and keeps backend fields and UI configuration aligned
+* **Good, because** in theory only 1 file needs to be changed to add a new integration
+* **Bad, because** it requires clear conventions and discipline to avoid overly complex plugins
+
+### Option 1:
+
+* **Good, because** Migrations are relatively easy
+* **Bad, because** If a developer wants other columns for their integrations a migration script is needed
```
stijndehaes: I don't think a migration script is bad, you are thinking of a shared table here right? While every integration could have their own table

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,111 @@
...
+* **Bad, because** Artificial (wrong) mappings might be introduced to make technologies work without a new migration
```
stijndehaes: Why? Every integration might introdude their own table, and own migration?

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,111 @@
...
+## Second discussion: Plugin provisioning
+
+* **Option 1:** Plugins are only an internal system to extend
+* **Option 2:** Users can deliver plugins from a different source, e.g. through volume mounts
+
+---
+
+## Decision Outcome
+
+**Chosen option:** *Option: Python Plugin Architecture and Option 2: JSON Blobs*
```
stijndehaes: I would rethink the JSON blobs decision, I think we can make it work with just migrations and tables.
Each plugin just has their own migration folder. 

Otherwise we are just kicking the can down the road if you ever want to suport migrating these json blobs.
If we do decide on JSON blobs I would already add a required version field to them (or a verison column in the table), so we can support upgrades later.

stijn-janssens: Your comments make sense, and I think it is fair to say that a plugin has their own table and own migration script.
That does mean I think that having the plugins be addable at runtime becomes impossible.

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,352 @@
+# ADR: Introduce Plugin-Based Architecture for Data Output Configuration
+
+## Context and Problem Statement
+
+The portal supports multiple data output integrations (e.g. S3, Snowflake, Databricks, Glue, Redshift). Each integration serves a common goal:
+
+* Generate UI in the frontend where users can input technical fields
+* Allow for infrastructure provisioning
+* Persist configuration safely in the database
+* Allow for access tiles in the frontend if needed
```
stijndehaes, line 10: The access tiles are the link to the cloud on the right?

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,352 @@
...
+* Improve developer experience and maintainability
+* Avoid YAML-based configuration with limited type safety
+* Prepare the codebase for future extensibility without committing to a full plugin runtime yet
+* Align with a longer-term vision where integrations can be developed and deployed independently
+
+---
+
+## Considered Options
+
+* **Shared option: Python Plugin Architecture (Chosen)**
```
stijndehaes: It's unclear to me why this is called shared option?

stijndehaes: There are basically 2 discussions right. How to store it and how to provision plugins. So maybe just remove this, and in your decision outcome you just say for question X, option Y, and question Z, Option A

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,352 @@
...
+            "platform_tile": null
+        },
+        {
+            // ...
+        }
+    ]
+  }
+```
+
+#### Example response GET /api/v2/plugins/<GLUE_ID>
```
stijndehaes:
```suggestion
#### Example response GET /api/v2/plugins/<GLUE_ID>/form
```

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,352 @@
...
+    "result_tooltip": "The table you can access through this technical asset",
+    "platform": "glue",
+    "display_name": "Glue",
+    "icon_name": "glue-logo.svg",
+    "parent_platform": "aws",
+    "platform_tile": null
+  }
+```
+
+#### Database representation
```
stijndehaes, line 210: A json as database representation is a bit weird? The idea I assume is to have each of these fields in a column in the DB?

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,352 @@
...
+    required: bool
+    name: str
+    tooltip: Optional[str] = None
+    initial_value: Optional[str | int | float | bool] = None
+    depends_on: Optional[FieldDependency] = None
+    max_count: Optional[int] = None
+    disabled: Optional[bool] = None
+    use_namespace_when_not_source_aligned: Optional[bool] = None
+
+class AssetProviderPlugin(BaseModel):
```
stijndehaes, line 254: Maybe add comments to the method, as they are a bit vague to me :D

stijndehaes: These are vague for me:
- validate
- render_template
- get_configuration

docs/adr/0009-plugin-data-output-configs.md
stijn-janssens: yes

docs/adr/0009-plugin-data-output-configs.md
vlieven: I also think a SQLAlchemy representation would make more sense here maybe

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,356 @@
...
+
+## Confirmation
+
+This decision is reflected in the application by:
+
+* Migrating all existing integrations to class-based plugins
+* Discover new plugins at startup (environment variable is sufficient for discovery)
+* Exposing integration metadata via REST APIs
+* Replacing hardcoded frontend forms with dynamically generated UI
+* Persisting configuration as JSON blobs scoped to the plugin type
```
pascal-knapen: Above it is stated we go for a table per plugin type? This still valid

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,356 @@
...
+
+class AssetProviderPlugin(BaseModel):
+    name: ClassVar[str]
+    version: ClassVar[str]
+
+    @classmethod
+    def get_ui_metadata(cls) -> List[UIElementMetadata]:
+        raise NotImplementedError
+
+    def validate(self):
```
stijndehaes: I know this is pseudocode, but I guess in actuall implementation this needs to take arguments right?

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -0,0 +1,356 @@
...
+    required: bool
+    name: str
+    tooltip: Optional[str] = None
+    initial_value: Optional[str | int | float | bool] = None
+    depends_on: Optional[FieldDependency] = None
+    max_count: Optional[int] = None
+    disabled: Optional[bool] = None
+    use_namespace_when_not_source_aligned: Optional[bool] = None
+
+class AssetProviderPlugin(BaseModel):
```
stijndehaes, line 254: We are missing migration file location here in the plugin, I would assume that the plugin tells you were migration files are located. And portal exectutes them?

docs/adr/0009-plugin-data-output-configs.md
stijn-janssens: yes, to be seen in the future what we need I would say. I'll add **context :D

docs/adr/0009-plugin-data-output-configs.md
stijndehaes: Can we quickly add a pseudo function so this is clear for other reviewers :)

stijndehaes (approved):

PR #2414, opened 2026-01-07
Move to generated client for authorization-roles

vlieven:


frontend/src/pages/people/people-table.component.tsx
```diff
@@ -33,11 +33,12 @@ function filterUsers(users: UsersGetContract, searchTerm?: string) {
 export function PeoplePage() {
     const { t } = useTranslation();
     const { data: users = [], isFetching } = useGetAllUsersQuery();
-    const { data: roles = [] } = useGetRolesQuery(Scope.GLOBAL);
+    const { data: rolesResponse } = useGetRolesQuery('global');
```
stijn-janssens: why can't this be done with the `Scope.GLOBAL`?

frontend/src/pages/roles/components/roles-table.component.tsx
```diff
@@ -75,8 +78,10 @@ type RolesTableProps = {
     scope: Scope;
 };
 export function RolesTable({ scope }: RolesTableProps) {
-    const { data: globalRoles = [], isFetching: isFetchingGlobalRoles } = useGetRolesQuery(Scope.GLOBAL);
-    const { data: rawRoles = [], isFetching: isFetchingRoles } = useGetRolesQuery(scope);
+    const { data: globalRolesResponse, isFetching: isFetchingGlobalRoles } = useGetRolesQuery('global');
```
stijn-janssens: same question here

frontend/src/types/roles/enums.ts
```diff
@@ -0,0 +1,30 @@
...
+export const Prototype = {
+    CUSTOM: 0 as ApiPrototype,
+    EVERYONE: 1 as ApiPrototype,
+    OWNER: 2 as ApiPrototype,
+    ADMIN: 3 as ApiPrototype,
+} as const;
+
+export type Prototype = (typeof Prototype)[keyof typeof Prototype];
+
+export const Scope = {
```
stijn-janssens, line 16: why is this still needed? Why can't you just work with the Scope from the generated spec?

vlieven: It's not strictly needed. What is generated from the spec is a string union, meaning you can only specify the value as one of the valid string literals. This provides an enum with the same values, but the function is the same, so the difference is mainly stylistic.

PR #2436, opened 2026-01-09
Feature/frontend generation of output configs

stijn-janssens:
Start of implementation for removing output configuration specific / hardcoded stuff from the frontend.
Let the frontend be generated completely from metadata provided over the API.

This PR can be created and merged without looking at the ADR yet. It is prework for the ADR implementation that can come afterwards.


backend/app/data_output_configuration/router.py
```diff
@@ -0,0 +1,60 @@
...
+@router.get("/{plugin_name}/form")
+def get_plugin_form(
+    plugin_name: str,
+    db: Session = Depends(get_db_session),
+) -> UIElementMetadataResponse:
+    """Get form metadata for a specific plugin (ADR-compliant endpoint)"""
+    all_plugins = PluginService(db).get_technical_asset_ui_metadata()
+
+    # Find the plugin by name
+    plugin = next((p for p in all_plugins if p.plugin == plugin_name), None)
```
nclaeys: I think it make sense to push this down to the plugin service no? I guess that getting a plugin by name will be common no?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -7,11 +9,100 @@
...
+    type: UIElementType
+    required: bool
+    name: str
+    tooltip: Optional[str] = None
+    options: Optional[dict] = None
+    hidden: Optional[bool] = None
+    pattern: Optional[str] = None
+    initial_value: Optional[str | int | float | bool] = None
+    value_prop_name: Optional[str] = None
+    depends_on: Optional[FieldDependency] = None
```
stijndehaes: Should this be a list?

Also can we add some documentation on the `FieldDependency` on what it means and how it's supposed to be rendered?

stijndehaes:
Can we add screenshots of a technical asset?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -7,11 +9,100 @@
...
+    value: Any
+
+
+class UIElementMetadata(ORMModel):
+    label: str
+    type: UIElementType
+    required: bool
+    name: str
+    tooltip: Optional[str] = None
+    options: Optional[dict] = None
```
stijndehaes: options is currently not used right?

I also find it pretty weird that it is a dict. What is the purpose behind it? Are these select options?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -7,11 +9,100 @@
...
+    hidden: Optional[bool] = None
+    pattern: Optional[str] = None
+    initial_value: Optional[str | int | float | bool] = None
+    value_prop_name: Optional[str] = None
+    depends_on: Optional[FieldDependency] = None
+    max_count: Optional[int] = None
```
stijndehaes: I noticed `max_count` is only used for a select option.

Maybe we can group this a bit, something like this:
``` 
selectConfig: {
    "maxOption": 1,
    "options": ["bla", "bla"]
}
```

backend/app/data_output_configuration/base_schema.py
```diff
@@ -7,11 +9,100 @@
...
+class UIElementMetadata(ORMModel):
+    label: str
+    type: UIElementType
+    required: bool
+    name: str
+    tooltip: Optional[str] = None
+    options: Optional[dict] = None
+    hidden: Optional[bool] = None
+    pattern: Optional[str] = None
+    initial_value: Optional[str | int | float | bool] = None
```
stijndehaes: Maybe add this to the select options, or checkbox options to have better typing?

I also think select initialvalue could be a list?

stijndehaes (changes requested):

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+
+class SelectOption(ORMModel):
+    """Option for select UI elements"""
+
+    label: str
+    value: str | bool
+
+
+class UIElementMetadata(ORMModel):
+    label: str  # The label shown in the UI, passed through a translating layer
```
stijndehaes: Where is this translation done? The front end can't know the translation right, since the plugin implmentation decides the name

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+    type: UIElementType  # The type of UI element
+    required: bool  # Whether this field is required to be filled in in the form
+    name: str  # The internal name of the field, used as key in the configuration
+    tooltip: Optional[str] = (
+        None  # Tooltip text shown in the UI, passed through a translating layer
+    )
+    hidden: Optional[bool] = (
+        None  # Whether this field is hidden in the UI, used to pass generated values without user input.
+    )
+    initial_value: Optional[str | bool] = (
```
stijndehaes: I would move `initial_value` to `UIElementCheckbox` and `UIElementSelect` so we can have simpler typing for it.

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+        None  # Initial value for the field in the UI, select fields can't have an initial value yet.
+    )
+    value_prop_name: Optional[str] = (
+        None  # Used to override the default value_prop_name if needed, e.g. checked for checkboxes
```
stijndehaes: The explanation here is unclear to me, since I am unsure what `value_prop_name` actually means

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+    )
+    depends_on: Optional[FieldDependency] = (
```
stijndehaes: Should we make this a list?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+
+
+class SelectOption(ORMModel):
+    """Option for select UI elements"""
+
+    label: str
+    value: str | bool
+
+
+class UIElementMetadata(ORMModel):
```
stijndehaes, line 58: I notice this one still combines all fields possible for our users, and we override specific fields when needed.
Isn't it posslbe to move fields like `max_count` to only exist in: `UIElementSelect` for example?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+
+    name: ClassVar[str]
+    version: ClassVar[str] = "1.0"
     configuration_type: DataOutputTypes
 
+    # Platform metadata - should be overridden in subclasses
+    _platform_metadata: ClassVar[Optional[PlatformMetadata]] = None
+
     def render_template(self, template: str, **context) -> str:
+        """Render a template with configuration values"""
```
stijndehaes: What is this method used for?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
         return template.format(**self.model_dump(), **context)
 
     def get_configuration(self, configs: list[ConfigType]) -> Optional[ConfigType]:
+        """Get platform and environment specific configuration"""
         raise NotImplementedError
+
+    def validate(self) -> None:
```
stijndehaes: After a quick check I believe this method is unused?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+            )
+        return cls._platform_metadata
+
+    @classmethod
+    def get_logo(cls) -> str:
+        """Get the logo filename for this plugin"""
+        platform_meta = cls.get_platform_metadata()
+        return platform_meta.icon_name
+
+    def get_url(self, environment: str) -> Optional[str]:
```
stijndehaes: This method is unused?

backend/app/data_output_configuration/base_schema.py
```diff
@@ -1,17 +1,178 @@
...
+        """Get the URL for accessing this resource in the given environment"""
+        return None
+
+    @classmethod
+    def has_environnments(cls) -> bool:
```
stijndehaes: method unused?

backend/app/data_output_configuration/redshift/schema.py
```diff
@@ -1,26 +1,48 @@
...
     configuration_type: Literal[DataOutputTypes.RedshiftDataOutput]
     table: str = "*"
     bucket_identifier: str = ""
     database_path: str = ""
     table_path: str = ""
+    entire_schema: bool = False
+
+    _platform_metadata = PlatformMetadata(
+        display_name="Redshift",
+        icon_name="aws-redshift-logo.svg",
```
stijndehaes, line 42: Maybe one for later, but shouldn't the plugin provide the image? If we want to support out of tree plugins we can't relly on these images being available in the front end

backend/app/data_output_configuration/router.py
```diff
@@ -0,0 +1,50 @@
...
+) -> PlatformTileResponse:
+    configs = PlatformServiceConfigurationService(
+        db
+    ).get_all_platform_service_configurations()
+    return PlatformTileResponse(
+        platform_tiles=PluginService(db).get_platform_tiles(configs)
+    )
+
+
+# ADR-compliant endpoints
```
stijndehaes:
```suggestion
```

backend/app/data_output_configuration/service.py
```diff
@@ -0,0 +1,127 @@
...
+        data_output_configurations = AssetProviderPlugin.__subclasses__()
+
+        return [
+            y
+            for y in [
+                self._build_metadata_response(plugin)
+                for plugin in data_output_configurations
+            ]
+            if y is not None
+        ]
```
stijndehaes, line 30: I believe you can write it like this:

```suggestion
        return [
            metadata_response for plugin in data_output_configurations
            if (metadata_response := self._build_metadata_response(plugin)) is not None
        ]
```

Also renamed y to `metadata_response` for readability

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
+from fastapi.testclient import TestClient
+
+from tests.factories.platform_service import PlatformServiceFactory
+from tests.factories.platform_service_config import PlatformServiceConfigFactory
+
+ENDPOINT = "/api/v2/plugins"
+
+
+class TestPluginEndpoints:
+    """Test ADR-compliant plugin endpoints"""
```
stijndehaes:
```suggestion
```

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+        data = response.json()
+
+        # Verify response structure
+        assert "plugins" in data
+        assert isinstance(data["plugins"], list)
+        assert len(data["plugins"]) == 0
+
+    def test_list_plugins_returns_all_available_plugins(self, client: TestClient):
+        """Test GET /v2/plugins returns list of all available plugins"""
+        s3 = PlatformServiceFactory(name="S3")
```
stijndehaes, line 29: Do we need this to be created? Doesn't the endpoint just return all plugins even when they have not config? Or am I missing something?

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+            "RedshiftDataOutput",
+            "SnowflakeDataOutput",
+            "DatabricksDataOutput",
+        }
+        assert expected_plugins.issubset(plugin_names)
+
+    def test_get_plugin_form_by_name_returns_correct_plugin(self, client: TestClient):
+        """Test GET /v2/plugins/{plugin_name}/form returns specific plugin"""
+        s3 = PlatformServiceFactory(name="S3")
+        glue = PlatformServiceFactory(name="Glue")
```
stijndehaes:
```suggestion
```

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+        PlatformServiceConfigFactory(service=s3)
+        PlatformServiceConfigFactory(service=glue)
```
stijndehaes:
```suggestion
```

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+        assert data["plugin"] == "S3DataOutput"
+        assert data["platform"] == "s3"
+        assert data["display_name"] == "S3"
+        assert data["icon_name"] == "s3-logo.svg"
+        assert "ui_metadata" in data
+        assert isinstance(data["ui_metadata"], list)
+
+    def test_get_plugin_form_includes_all_fields(self, client: TestClient):
+        """Test that plugin form includes all expected fields"""
+        s3 = PlatformServiceFactory(name="S3")
```
stijndehaes:
```suggestion
```

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+        glue = PlatformServiceFactory(name="Glue")
+        PlatformServiceConfigFactory(service=s3)
```
stijndehaes:
```suggestion
```

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+        # Verify expected fields are present
+        expected_fields = {"database", "database_suffix", "entire_schema", "table"}
+        assert expected_fields.issubset(field_names)
+
+    def test_get_plugin_form_with_invalid_name_returns_404(self, client: TestClient):
+        """Test GET /v2/plugins/{plugin_name}/form with invalid name returns 404"""
+        s3 = PlatformServiceFactory(name="S3")
+        glue = PlatformServiceFactory(name="Glue")
+        PlatformServiceConfigFactory(service=s3)
+        PlatformServiceConfigFactory(service=glue)
```
stijndehaes:
```suggestion
```

backend/tests/app/plugins/test_router.py
```diff
@@ -0,0 +1,298 @@
...
+        assert "NonExistentPlugin" in data["detail"]
+        assert "not found" in data["detail"].lower()
+
+    def test_get_plugin_form_for_each_available_plugin(self, client: TestClient):
+        """Test that each plugin from list can be retrieved individually"""
+        # Get all plugins
+        s3 = PlatformServiceFactory(name="S3")
+        glue = PlatformServiceFactory(name="Glue")
+        PlatformServiceConfigFactory(service=s3)
+        PlatformServiceConfigFactory(service=glue)
```
stijndehaes:
```suggestion
```

Don't think this setup is required?

frontend/src/components/data-products/data-output-form/generic-data-output-form.component.tsx
```diff
@@ -0,0 +1,169 @@
...
+    form: FormInstance<DataOutputCreateFormSchema>;
+    namespace: string;
+    sourceAligned: boolean;
+    configurationType: string;
+    uiMetadataGroups: UiElementMetadata[];
+    resultLabel?: string;
+    resultTooltip?: string;
+};
+
+export function GenericDataOutputForm({
```
stijndehaes: I think we can just name it DataOutputForm but this is just a small remark

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,169 @@
...
+            use_namespace_when_not_source_aligned,
+        } = fieldMetadata;
+
+        // Check if field should be hidden based on dependencies
+        let isHidden = hidden;
+        if (depends_on) {
+            const dependentValue = watchedFields[depends_on.field_name];
+            isHidden = dependentValue !== depends_on.value;
+            if (isHidden) {
+                // Clear the field value if it's hidden
```
stijndehaes: Why do we need the value to be removed? I expected the field to be hidden and not shown, maybe I am misunderstanding the purpose of hidden :)

frontend/src/components/data-products/data-output-form/generic-data-output-form.component.tsx
```diff
@@ -0,0 +1,169 @@
...
+            isHidden = dependentValue !== depends_on.value;
+            if (isHidden) {
+                // Clear the field value if it's hidden
+                // biome-ignore lint: dynamic field names can't be statically typed at compile time.
+                form.setFieldValue(configurationFieldName(name) as any, initial_value ?? undefined);
+            }
+        }
+
+        // Build validation rules
+        const rules: { required?: boolean; pattern?: RegExp; message?: string }[] = [];
```
stijndehaes: Instead of making our own typing can't we import the correct typing class from ant?

frontend/src/components/data-products/data-output-form/generic-data-output-form.component.tsx
```diff
@@ -0,0 +1,169 @@
...
+        if (required && !isHidden) {
+            rules.push({
+                required: true,
+                message: t(`Please input ${label.toLowerCase()}`),
+            });
+        }
+
+        // Build select options based on sourceAligned for fields that use namespace
+        let selectOptions: { label: string; value: string }[] = [];
```
stijndehaes: Same here, can't we import the correct typing class from ant?

frontend/src/utils/icon-loader.ts
```diff
@@ -0,0 +1,19 @@
+// Dynamically import all SVG icons from the assets/icons directory
+const icons = import.meta.glob('@/assets/icons/*.svg', {
```
stijndehaes, line 2: I am unsure if this is a good idea, this could results in all svg's being part of the main bundle

stijndehaes: After a quick verification I noticed it doesn matter, bundle size is not increasing. They probably are already part of the bundle/or very small

stijndehaes (changes requested):

backend/app/data_output_configuration/base_schema.py
stijn-janssens: Well, you can provide translations for it if you want. By overriding the file. The string is passed through the translation layer in the frontend but indeed is not statically known upfront. We can remove this if we want but that means there is no way of overriding our default integrations?

backend/app/data_output_configuration/base_schema.py
stijn-janssens: value_prop_name is the element that decides the value of a form field. For some reason if you want the checked value of a checkbox to be passed in a form you have to explicitly set it to look for the checked attribute prop in the checkbox. https://ant.design/components/checkbox#faq-form-item-limitations

backend/app/data_output_configuration/base_schema.py
stijn-janssens: If you do then (at first sight) these attributes are not passed through the API. Because we return the base model only, so they got filtered out.

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
stijn-janssens: This functionality is currently being used in the `table_name` and `use_entire_schema` combination. If the checkbox of entire_schema is true, then table_name is not shown and defaults to `*`. However if you check the checkbox, change the table name and then uncheck the checkbox again, I want the behavior to be that table name once again becomes `*`.
Up for discussion on this one, but this is how the functionality currently works.

frontend/src/utils/icon-loader.ts
stijn-janssens: I don't believe we have looked into that yet, but I propose we leave this maybe as a future improvement? Since I believe functionally this does not change much because of this change? Correct me if I'm wrong on this

backend/app/data_output_configuration/base_schema.py
stijn-janssens: this shows the "end result" of all the filled in form values, rendered in a certain template.
e.g. in the case of S3, it shows your `bucket/suffix/path`, with glue it shows `db__schema.table` etc etc

backend/app/data_output_configuration/base_schema.py
stijn-janssens: for now yes, I included it to follow the ADR but can remove it if needed.

backend/app/data_output_configuration/base_schema.py
stijn-janssens: same as above

backend/app/data_output_configuration/base_schema.py
stijn-janssens: same as above

backend/app/data_output_configuration/redshift/schema.py
stijn-janssens: well to be discussed indeed. I am not really a fan of having the images somewhere outside of the frontend, but if we support plugins to be registered dynamically we will indeed need to figure something out like what you propose. If fine for you I would leave it indeed for future work

backend/tests/app/plugins/test_router.py
stijn-janssens: if they don't have a config `get_platform_options` (which might get called, but does not have to) might throw a NotImplementedError and then the plugin will return None.
I am not yet to fuzzed about this approach either :/

backend/app/data_output_configuration/base_schema.py
stijn-janssens: For now I wouldn't do that so we can follow strictly what we have in place in the frontend. We can extend / improve the structure later on maybe?

backend/tests/app/plugins/test_router.py
stijn-janssens: as explained above: to be seen if this is what we want but currently each plugin requires a service config to be available (to populate select options). If not available -> plugin is not shown.

backend/tests/app/plugins/test_router.py
stijn-janssens: as explained above: to be seen if this is what we want but currently each plugin requires a service config to be available (to populate select options). If not available -> plugin is not shown.

backend/tests/app/plugins/test_router.py
stijn-janssens: as explained above: to be seen if this is what we want but currently each plugin requires a service config to be available (to populate select options). If not available -> plugin is not shown.

backend/tests/app/plugins/test_router.py
stijn-janssens: as explained above: to be seen if this is what we want but currently each plugin requires a service config to be available (to populate select options). If not available -> plugin is not shown.

backend/tests/app/plugins/test_router.py
stijn-janssens: as explained above: to be seen if this is what we want but currently each plugin requires a service config to be available (to populate select options). If not available -> plugin is not shown.

backend/tests/app/plugins/test_router.py
stijn-janssens: as explained above: to be seen if this is what we want but currently each plugin requires a service config to be available (to populate select options). If not available -> plugin is not shown.

backend/app/data_output_configuration/base_schema.py
stijn-janssens: because of the typing of pydantic it needs to stay in this place as well. But I added it (single typed) to the children elements as well

backend/app/data_output_configuration/base_schema.py
stijndehaes: Proposal to make this similar to k8s volumes field in pods, by passing a single volume type config the volume type is decided. This is easy to generate client for every programming language, both typed and untyped.

Image

backend/app/data_output_configuration/base_schema.py
stijndehaes: Agreed to remove translations since it's very unclear for our users that these labels even exist. If we ever want translations we should implement them backend for plugins

backend/app/data_output_configuration/base_schema.py
stijndehaes: Decided to remove this in backend, hardcore in front end

backend/app/data_output_configuration/base_schema.py
stijndehaes: TODO make list

backend/app/data_output_configuration/base_schema.py
stijndehaes: Let's replace **context with optional[Environment], also add comment that template is stored in db

backend/tests/app/plugins/test_router.py
stijndehaes: Return the plugins as not_configured=True, this allows us to later show this to admins

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
stijndehaes: Would change it to fix this in backend when we save, the front end can just hide the field when needed but does not reset the value. This makes it easier for when we implement updates for technical assets

backend/app/data_output_configuration/redshift/schema.py
stijndehaes: Let's not implement for now, todo for future work.

frontend/src/utils/icon-loader.ts
stijndehaes: We don't make it worse so ok

backend/app/data_output_configuration/databricks/model.py
```diff
@@ -12,6 +12,9 @@ class DatabricksDataOutput(BaseDataOutputConfiguration):
     catalog_path: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
     table: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
     table_path: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
+    entire_schema: Mapped[bool] = mapped_column(
```
vlieven: `entire_schema` is a bit unclear to me. Based on the other changes, this boolean indicates access, so maybe `access_entire_schema` is better?

```suggestion
    access_entire_schema: Mapped[bool] = mapped_column(
```

Alternatively, this can be a "write_access" enum, like such:

```suggestion
    write_access: Mapped[Literal["schema" | "table" | "column"]] = mapped_column(
```

backend/app/data_output_configuration/glue/schema.py
```diff
@@ -1,26 +1,48 @@
-from typing import Literal, Optional, Self
+from typing import ClassVar, List, Literal, Optional, Self
```
vlieven: `typing.List` does not need to imported, on Python 3.9+, you can simply use `list[SomeType]` as annotation.

backend/app/data_output_configuration/service.py
```diff
@@ -0,0 +1,124 @@
...
+                display_name=platform_meta.display_name,
+                icon_name=platform_meta.icon_name,
+                parent_platform=platform_meta.parent_platform,
+                result_label=platform_meta.result_label,
+                result_tooltip=platform_meta.result_tooltip,
+            )
+        except NotImplementedError:
+            return None
+
+    def get_platform_tiles(self, configs: Sequence) -> Sequence[PlatformTile]:
```
vlieven: I am missing a type annotation for `configs` here.

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+    form,
+    namespace,
+    sourceAligned,
+    uiMetadataGroups,
+    configurationType,
+    resultLabel,
+    resultTooltip,
+}: Props) {
+    const { t } = useTranslation();
+    const uiMetadata = uiMetadataGroups as UiElementMetadata[] | undefined;
```
vlieven: What is the point of broadening this type to also allow `undefined` if the component does not accept it in the `Props`?

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+        entire_schema: entireSchema,
+    };
+
+    // Auto-populate fields based on sourceAligned and namespace
+    useEffect(() => {
+        uiMetadata?.forEach((field) => {
+            // Handle suffix field
+            if (field.name === 'suffix') {
+                // biome-ignore lint: dynamic field names can't be statically typed at compile time.
+                form.setFieldValue(configurationFieldName('suffix') as any, sourceAligned ? '' : namespace);
```
vlieven:
```suggestion
                form.setFieldValue(configurationFieldName('suffix'), sourceAligned ? '' : namespace);
```

backend/app/data_output_configuration/databricks/model.py
```diff
@@ -12,6 +14,11 @@ class DatabricksDataOutput(BaseDataOutputConfiguration):
     catalog_path: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
     table: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
     table_path: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
+    access_granularity: Mapped[AccessGranularity] = mapped_column(
+        Enum(AccessGranularity),
```
vlieven: We should not use postgres enums - we want to have the flexibility to easily add to this enum later, and postgres enums don't allow for it, defeating the purpose. This can be just a string representation in the DB I think.

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+    form,
+    namespace,
+    sourceAligned,
+    uiMetadataGroups,
+    configurationType,
+    resultLabel,
+    resultTooltip,
+}: Props) {
+    const { t } = useTranslation();
+    const uiMetadata = uiMetadataGroups as UiElementMetadata[];
```
vlieven:
```suggestion
```

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+    // Watch all fields - we must call hooks unconditionally and in the same order every render
+    // Since uiMetadata is stable for a given form, we can iterate through all fields
+    const watchedFields: Record<string, unknown> = {};
+    if (uiMetadata) {
+        for (const field of uiMetadata) {
+            // biome-ignore lint: hooks in loops are safe when array length is stable across renders
+            watchedFields[field.name] = Form.useWatch(configurationFieldName(field.name), form);
+        }
+    }
```
vlieven:
```suggestion
    for (const field of uiMetadata) {
        // biome-ignore lint: hooks in loops are safe when array length is stable across renders
        watchedFields[field.name] = Form.useWatch(configurationFieldName(field.name), form);
    }
```

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+    // Auto-populate fields based on sourceAligned and namespace
+    useEffect(() => {
+        uiMetadata?.forEach((field) => {
```
vlieven:
```suggestion
        uiMetadataGroups.forEach((field) => {
```

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+                form.setFieldValue(configurationFieldName('suffix'), sourceAligned ? '' : namespace);
+            }
+
+            // Handle fields that should auto-populate from namespace when not source-aligned
+            if (field.use_namespace_when_not_source_aligned) {
+                // biome-ignore lint: dynamic field names can't be statically typed at compile time.
+                form.setFieldValue(configurationFieldName(field.name) as any, !sourceAligned ? namespace : undefined);
+            }
+        });
+    }, [form, sourceAligned, namespace, uiMetadata]);
```
vlieven:
```suggestion
    }, [form, sourceAligned, namespace]);
```

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+    if (!uiMetadata) {
```
vlieven:
```suggestion
    if (!uiMetadataGroups) {
```

frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+    };
+
+    return (
+        <ConfigurationSubForm
+            form={form}
+            configurationType={configurationType}
+            resultLabel={resultLabel}
+            resultTooltip={resultTooltip}
+        >
+            {uiMetadata.map(renderFormField)}
```
vlieven:
```suggestion
            {uiMetadataGroups.map(renderFormField)}
```

PR #2442, opened 2026-01-12
Added improved search ADR

stijndehaes:

### DIFF
--- a/docs/adr/0010-improve-search.md
+++ b/docs/adr/0010-improve-search.md
@@ -1,0 +1,68 @@
+# Improve search
+
+## Context and Problem Statement
+
+We want to improve the search in the application. We want data consumers to be able to find relevant output ports,
+using natural language queries.
+During a hackathon we have investigated several options. And we want to implement one to improve the search.
+
+Currently we will only focus on providing support for English queries. If by popular demand we want to support other
+languages, we can add support for them later.
+
+## Decision Drivers
+
+* Speed (How fast does it offer the results to the users)
+* Precision (How accurate are the results)
+* Recall (How many relevant results are returned)
+
+## Considered Options
+
+* **Option 1: LLM Search** Pass on results to an LLM and let it rank and return the relevant results.
+* **Option 2: Embedding search** Option 2 Use an embedding model to rank and return the relevant results.
+* **Option 3: Combined search** Option 3 Use either option 1 or 2, and use a reranker to combine the results with our traditional search and rerank them.
+
+## Decision Outcome
+
+**Chosen option:** *Option 2: Embedding search*. Through our investigation, Option 2 had the best results, and it was fast enough to be used in the application.
+The best results were with an opensource small model, ensuring our dependency isn't too big.
+
+### Confirmation
+
+We will switch from the current search to the embedding search.
+
+## Pros and Cons of the Options
+
+### Option 1: LLM Search
+
+* **Good, because** It performs very well.
+* **Bad, because** Unclear how to scale this to a large number of output ports and data products.
+* **Bad, because** Even with <100 output ports, it's already slow
+* **Bad, because** It's not trivial to ensure the LLM returns results in the correct way
+* **Bad, because** Need to use a cloud service, which can be expensive, and we probably have to implement multiple integrations
+
+### Option 2: Embedding search
+
+* **Good, because** Small embedding models are fast and good enough for our use case.
+* **Good, because** We are able to use open source models, ensure we don't need to implement different models for different clouds/vendors
+* **Neutral, because** We have a dependency on pgvector being available on postgres, this is available on Azure and AWS, but might not be on other clouds
+* **Bad, because** It increases our image size since we need to include the embedding model
+* **Bad, because** It isn't the best performing model
+* **Bad, because** It returns all results in a ranked order, and setting a cut-off point isn't trivial
+
+### Option 3: Combined search
+
+* **Good, because** It can combine the best of both worlds
+* **Good, because** There are open source models we can use
+* **Neutral, because** We haven't investigated how well it performs
+* **Bad, because** It increases our image size since we need to include the reranker model
+* **Bad, because** Complicated because there are even more knobs to tune
+
+# Investigation results
+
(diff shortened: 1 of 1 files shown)


docs/adr/0010-improve-search.md
```diff
@@ -0,0 +1,64 @@
...
+to implement one to improve the search.
+
+## Decision Drivers
+
+* Speed (How fast does it offer the results to the users)
+* Precision (How accurate are the results)
+* Recall (How many relevant results are returned)
+
+## Considered Options
+
```
nclaeys, line 19: Why is one of the options not extend the current search to include more fields? As I remember you had some queries on fields that were not indexed on the current search which would lead to a higher recall.

Is the goal to also support querying through human text or only using keywords?

stijndehaes: Good question, one of the goals is indeed to support querying through human text and improve the UX that way.

nclaeys: k that implies that the current search is indeed not feasible. Maybe good to add this to the context/problem statement.

stijndehaes: Added

docs/adr/0010-improve-search.md
```diff
@@ -0,0 +1,68 @@
...
+* **Bad, because** Even with <100 output ports, it's already slow
+* **Bad, because** It's not trivial to ensure the LLM returns results in the correct way
+* **Bad, because** Need to use a cloud service, which can be expensive, and we probably have to implement multiple integrations
+
+### Option 2: Embedding search
+
+* **Good, because** Small embedding models are fast and good enough for our use case.
+* **Good, because** We are able to use open source models, ensure we don't need to implement different models for different clouds/vendors
+* **Neutral, because** We have a dependency on pgvector being available on postgres, this is available on Azure and AWS, but might not be on other clouds
+* **Bad, because** It increases our image size since we need to include the embedding model
```
vlieven, line 48: Should we maybe run it as a separate service / container? That also might make it more easy to swap out at a later time? Although that's probably premature optimization at this point.

docs/adr/0010-improve-search.md
```diff
@@ -0,0 +1,68 @@
...
+* **Bad, because** It isn't the best performing model
+* **Bad, because** It returns all results in a ranked order, and setting a cut-off point isn't trivial
```
vlieven, line 50: Do you get a score or only a rank? If you get a score, there are ways of choosing a sensible cutoff point (e.g. elbow method)

docs/adr/0010-improve-search.md
```diff
@@ -0,0 +1,68 @@
...
+| Current search                   | ~20ms  | 86%       | 48%    | Extremely fast, good precision, very poor coverage (misses many relevant items, especially not adapted for broader business questions) |
+| LLM Search with: Claude-3-Haiku  | ~7s    | 91%       | 81%    | Very high precision and coverage, but very slow and inconsistent depending on input/output. Dependency on cloud (AWS Bedrock)          |
+| Embedding searcH: BAAI/bge-large | ~2s    | 81%       | 91%    | Excellent coverage, good precision, speed not that good.Open source Large model, needs a while to download                             |
+| Embedding Bedrock/Cohere v4.0    | ~1.5s  | 71%       | 77%    | Medium speed, lowest precision among embeddings, okay coverage. Dependency on cloud (AWS Bedrock)                                      |
+| Embedding BAAI/bge-small         | ~150ms | 76%       | 86%    | Very fast, slightly lower precision and coverage than large, good trade-off for speed-sensitive use.Open source small model            |
```
vlieven:
```suggestion
| Embedding BAAI/bge-small         | ~150ms | 76%       | 86%    | Very fast, slightly lower precision and coverage than large, good trade-off for speed-sensitive use. Open source small model           |
```

docs/adr/0010-improve-search.md
```diff
@@ -0,0 +1,68 @@
...
+* **Bad, because** Complicated because there are even more knobs to tune
+
+# Investigation results
+
+| Method                           | Speed  | Precision | Recall | Notes                                                                                                                                  |
+|----------------------------------|--------|-----------|--------|----------------------------------------------------------------------------------------------------------------------------------------|
+| Current search                   | ~20ms  | 86%       | 48%    | Extremely fast, good precision, very poor coverage (misses many relevant items, especially not adapted for broader business questions) |
+| LLM Search with: Claude-3-Haiku  | ~7s    | 91%       | 81%    | Very high precision and coverage, but very slow and inconsistent depending on input/output. Dependency on cloud (AWS Bedrock)          |
+| Embedding searcH: BAAI/bge-large | ~2s    | 81%       | 91%    | Excellent coverage, good precision, speed not that good.Open source Large model, needs a while to download                             |
```
vlieven:
```suggestion
| Embedding search: BAAI/bge-large | ~2s    | 81%       | 91%    | Excellent coverage, good precision, speed not that good. Open source Large model, needs a while to download                            |
```

docs/adr/0010-improve-search.md
stijndehaes: You get a score, I have been investigating cutoff points during innovations days. But basically changing the dataset or embedding model used wildly changes the valid cutoff point that makes sense.
Both a static cutoff point and elbow method were tried. Neither were easy

docs/adr/0010-improve-search.md
stijndehaes: I don't think it makes a lot of sense because of the following:
- You have to run another container, making deployment more complex. I would rather have that we let the backend host the frontend files at some point. So we only have a single container to deploy. Makes things easier
- We have to send quite a lot of data over the wire. Probably best if we send a text representation over the wire, and get embeddings back. But still quite a lot of data

In the end currently don't think it's worth the effort as you said. The model `BAAI/bge-small` is about 130mb I believe, should I verify the impact on the image? I can test it out. Because the model will be compressed in the image.

vlieven: No, I was just thinking of ways to mitigate the downside, we have run containers that are larger than 130 mb :)

vlieven (approved):

stijn-janssens (approved):

PR #2454, opened 2026-01-13
Add ADR for ingesting data quality information

nclaeys:

### DIFF
--- a/docs/adr/0011-ingestion-data-quality.md
+++ b/docs/adr/0011-ingestion-data-quality.md
@@ -7,6 +7,7 @@ We would like to show a high-level view or summary in Portal, without the detail
 
 This summary should easily be calculated based on the output of the DQ tooling using a simple script.
 The portal API will support ingesting the summary of data quality results.
+We will follow the same approach for ingesting as described in [ADR 0007 - Ingestion usage metrics](./0007-ingestion-usage-metrics.md) to be consistent.
 
 In the future we can simplify the ingestion process by providing an SDK or creating a plugin for different DQ tooling.
 We currently store the summary results on output ports, but the summary contains high-level information about the technical assets.


docs/adr/0010-ingestion-data-quality.md
```diff
@@ -0,0 +1,99 @@
...
+* **Bad, because** it requires a simple script to calculate the summary as an extra step after running the DQ checks
+
+Summary submitted by a producer could look as follows:
+
+```
+{
+  "generated_at": "2026-01-12T08:15:00Z",
+  "overall_status": "WARN",
+  "summary": "Optional extra user info",
+  "asset_details": [
```
stijndehaes: `asset_details`. are these technical assets or output ports?

docs/adr/0011-ingestion-data-quality.md
```diff
@@ -0,0 +1,99 @@
...
+    {
+      "asset_name": "orders",
+      "status": "PASS"
+    },
+    {
+      "asset_name": "order_items",
+      "status": "WARN"
+    }
+  ],
+  "dimensions": {
```
stijndehaes, line 80: Are the dimension options (`completeness`, `validity` ...) something we decide up front or dynamic?

docs/adr/0010-ingestion-data-quality.md
```diff
@@ -0,0 +1,99 @@
...
+      "asset_name": "order_items",
+      "status": "WARN"
+    }
+  ],
+  "dimensions": {
+    "completeness": "WARN", #Are records missing or partial
+    "validity": "PASS" # Do values have the expected format, ranges,...
+    "coverage": # Can be calculated by portal as assets with checks/total assets
+  },
+  "dimension_coverage": {
```
stijndehaes: unsure what this adds compared to dimensions list?

docs/adr/0011-ingestion-data-quality.md
```diff
@@ -0,0 +1,99 @@
...
+
+* **Good, because** it can provide all the necessary details for showing actionable DQ in portal
+* **Good, because** it support many extension possibilities in the future, when we come up with extra dimensions.
+* **Neutral, because** it is relatively simple to calculate a summary based on the different DQ results of Soda, dbt, Great expectations
+* **Bad, because** it requires a simple script to calculate the summary as an extra step after running the DQ checks
+
+Summary submitted by a producer could look as follows:
+
+```
+{
```
stijndehaes, line 62: How would we show this information in portal? I have some questions below, so it's a bit vague to me. Maybe when the questions are answered it will be a bit clearer :)

docs/adr/0011-ingestion-data-quality.md
```diff
@@ -0,0 +1,99 @@
+# Ingest data quality results in Portal
```
stijndehaes, line 1: One big question I have is how will people get the data in? Will we just give them and endpoint? And SDK? A plugin for their quality framework?

docs/adr/0010-ingestion-data-quality.md
nclaeys: good remark, these are technical_assets. Well it could be that the checks contain technical_assets that do not exist in Portal, these we should filter out. The goal is to have quality info for every technical asset in the output port.

docs/adr/0011-ingestion-data-quality.md
nclaeys: My idea is that they are optional for now and we can add more later. This were 3 I think could be interesting.
I do not see them as dynamic, that every company can choose their own.

docs/adr/0010-ingestion-data-quality.md
nclaeys: True, this should indeed not be provided by the producer. It only specifies whether this dimension was calculated. I will remove it.

docs/adr/0011-ingestion-data-quality.md
nclaeys: We start with the API to feed the summary info with an example script to convert run_results.json from dbt or soda and feed it to portal as a step after running the quality checks. I guessed this is similar to what you do for usage?

In the future we can create plugin for dq tool, portal sdk to simplify the ingestion

docs/adr/0011-ingestion-data-quality.md
nclaeys: Well the idea is to give some info on DQ for every output port. This to show users whether they can trust the data. It was not in scope of the ticket to come up with a full design, but potentially it could like as follows:

Screenshot from 2026-01-13 16-17-39

docs/adr/0011-ingestion-data-quality.md
stijndehaes: For usage we made multiple ADRS at once, one of them being [0007 Ingestion usage metrics](https://github.com/conveyordata/data-product-portal/blob/main/docs/adr/0007-ingestion-usage-metrics.md), this one describes how people will ingest the data. But probably we can build on that decision. Nothing has been build yet though

docs/adr/0011-ingestion-data-quality.md
stijndehaes: The reason I asked was to ensure we capture all the information we want to show. 
So we basically capture it on the level of technical assets, and show aggregated information on output ports right?

nclaeys: Yes that is the current idea.

docs/adr/0011-ingestion-data-quality.md
nclaeys: Ah yes, k I would then use the same approach to be consistent. I will document it explicitly.

docs/adr/0011-ingestion-data-quality.md
```diff
@@ -0,0 +1,104 @@
...
+### Option 3: Extract a summary of DQ results
+
+* **Good, because** it can provide all the necessary details for showing actionable DQ in portal
+* **Good, because** it support many extension possibilities in the future, when we come up with extra dimensions.
+* **Neutral, because** it is relatively simple to calculate a summary based on the different DQ results of Soda, dbt, Great expectations
+* **Bad, because** it requires a simple script to calculate the summary as an extra step after running the DQ checks
+
+Summary submitted by a producer could look as follows:
+
+```
```
stijndehaes, line 61: I suppost we store this in a seperate table linked to the data product right?

nclaeys: yes separate table linked to the output port

Issue #2476, opened 2026-01-15
Improve search ADR implementation


stijndehaes:
- [x] Implement embedding
- [x] Improve keyword search to allow for `keywo` to find `keyword`, so basically add * behind query. Also add splits between words? For example `what are we cooking` should try to match for those 4 words instead of seeing this as one word => This is a lot harder then it seems the functions `websearch_to_tsquery` changes alles words to a english dictionary version, for example `apple` will be `appl` so the distinction between singular and plural dissapears, the actual fix is to use trigrams `pg_trgm` but that removes us farther from what we want, so won't do

For later:
- [ ] Investigating filtering out of search results, currently this happens after loading results from the DB, however this will have issues since to get good performance we need to put a limit on the embedding search
- [x] We now return all items through embedding search to we want to investigate a limit => this is acceptable
- [x] Do we want the debounce on search? Or do we want people to press enter to search. Investigation learns that everything works with trying to provide suggestions while typing, and for searching you need to press enter (coolblue, amazon, google work this way) => debounce removed

PR #2477, opened 2026-01-15
First implementation of embedding search

stijndehaes:
In support of: #2476

### DIFF
--- a/backend/app/database/alembic/versions/2026_01_07_1455-dbe6fbe45eef_add_entire_schema_checkbox.py
+++ b/backend/app/database/alembic/versions/2026_01_07_1455-dbe6fbe45eef_add_entire_schema_checkbox.py
@@ -1,7 +1,7 @@
 """Add entire_schema checkbox
 
 Revision ID: dbe6fbe45eef
-Revises: 49990691cd30
+Revises: ca5f5782790a
 Create Date: 2026-01-07 14:55:10.067397
 
 """
@@ -13,7 +13,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "dbe6fbe45eef"
-down_revision: Union[str, None] = "49990691cd30"
+down_revision: Union[str, None] = "ca5f5782790a"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/backend/app/database/alembic/versions/2026_01_14_1250-ca5f5782790a_add_cascade_delete_to_dataset_query_stats_daily.py
+++ b/backend/app/database/alembic/versions/2026_01_14_1250-ca5f5782790a_add_cascade_delete_to_dataset_query_stats_daily.py
@@ -1,7 +1,7 @@
 """add cascade delete to dataset_query_stats_daily
 
 Revision ID: ca5f5782790a
-Revises: dbe6fbe45eef
+Revises: 49990691cd30
 Create Date: 2026-01-14 12:50:00.000000
 
 """
@@ -12,7 +12,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "ca5f5782790a"
-down_revision: Union[str, None] = "dbe6fbe45eef"
+down_revision: Union[str, None] = "49990691cd30"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/backend/app/database/alembic/versions/2026_01_27_1019-c3fecd278d0c_add_output_port_embeddings.py
+++ b/backend/app/database/alembic/versions/2026_01_27_1019-c3fecd278d0c_add_output_port_embeddings.py
@@ -1,8 +1,8 @@
 """Add output-port embeddings
 
 Revision ID: c3fecd278d0c
-Revises: ca5f5782790a
-Create Date: 2026-01-14 14:14:28.166837
+Revises: dbe6fbe45eef
+Create Date: 2026-01-27 10:19:31.581396
 
 """
 
@@ -14,7 +14,7 @@ from pgvector.sqlalchemy import Vector
 
 # revision identifiers, used by Alembic.
 revision: str = "c3fecd278d0c"
-down_revision: Union[str, None] = "ca5f5782790a"
+down_revision: Union[str, None] = "dbe6fbe45eef"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 


backend/app/data_products/output_ports/schema.py
```diff
@@ -22,3 +24,28 @@ class OutputPort(ORMModel):
 class Dataset(OutputPort):
     def convert(self) -> OutputPort:
         return OutputPort(**self.model_dump())
+
+
+class DataProductEmbed(ORMModel):
```
stijn-janssens, line 29: is this the most logical file to place the schema in?
Shouldn't we put the Embed version of each entity in it's own corresponding schema file?

backend/app/embed.py
```diff
@@ -0,0 +1,15 @@
+import typer
+from fastembed import TextEmbedding
+
+from app.core.embed.model import EMBEDDING_MODEL
+
+app = typer.Typer(help="Embeddings toolkit for the Data product portal.")
+
+
+@app.command(short_help="Preload embedding model for caching")
+def load_embeddings_model():
```
stijn-janssens, line 10: what is this used for?

backend/app/core/embed/model.py
```diff
@@ -0,0 +1 @@
+EMBEDDING_MODEL = "BAAI/bge-small-en-v1.5"
```
stijn-janssens, line 1: is there a reason why you make this a single python file and not put it in settings?

docs/docs/developer-guide/installation.md
```diff
@@ -33,6 +33,13 @@ Update your `values.yaml` to match your environment. By default, images will use
 
 #### Database
 
+:::warning
+
+We use PGVector to empower search functionality in Portal.
+Most cloud vendors do support PGVector so it's most likely not an issue.
```
stijn-janssens: Do we need a minimum version of Postgres or need to do any migration / installation up front?
Are we at risk of losing data at any point by doing this auto migration?

backend/app/data_products/output_ports/schema.py
stijndehaes: It's very specific to the outport service, this isn't used by other services. So I don't think that makes sense.

backend/app/embed.py
stijndehaes: So it preloads the embedding model, it is used when building the docker container to ensure the model is download so it doesn't have to be downloaded at runtime.

Downloading at runtime is bad since the service where you download it from might be down introducing an outage.

backend/app/core/embed/model.py
stijndehaes: It's not a setting that users can override since changes a model changes the vector size we need to store and would need a database change.

I put it in this file since we need to use this model name in multiple places. I first put it in the DAtasetService, however the preload needs it and when trying to load the variable from DatasetService all environment variables needed for the application where needed. While we don't really need the whole application for that function

docs/docs/developer-guide/installation.md
stijndehaes: Good question, I can put a minimum version of pgvector and postgres here.

I would choose postgres 17 (it's what we do our testing with, so an older one might not work with current code), for pgvector I'll check what version we can use.

There is no risk of losing data, there is also no need for an auto migration, but I might not understand properly what you mean by that

backend/tests/app/data_products/output_ports/test_service.py
```diff
@@ -60,148 +58,20 @@ def test_get_private_dataset_by_member_of_consuming_data_product(self):
         ds = self.get_dataset(ds)
         assert DatasetService(test_session).is_visible_to_user(ds, user) is True
 
-    def test_create_search_vector_dataset(self):
```
nclaeys, line 63: We hardly have any tests for our search using embeddings. Shouldn't we run a baseline precission,recall,... test here or in CI to see when we make changes that it improves or not?

backend/app/data_products/output_ports/schema.py
vlieven: Would it not make more sense to have a separate service for the embeddings and then compose that into the output port service?

backend/app/data_products/router.py
```diff
@@ -169,6 +170,10 @@ def create_data_product(
         extra_receiver_ids=owners,
     )
     RefreshInfrastructureLambda().trigger()
+    db.flush()
+    DatasetService(db).recalculate_embeddings_for_output_ports_of_product(
+        created_data_product.id
+    )
```
vlieven: Should the flush not be part of the `recalculate_embeddings_for_output_ports_of_product`?

backend/app/data_products/output_ports/service.py
```diff
@@ -122,53 +120,115 @@ def get_dataset(
...
+        query_embedding = self.embedding_model.embed(query)
+        semantic_score = (
+            1 - DatasetModel.embeddings.cosine_distance(*query_embedding)
+        ).label("sem_score")
+        ts_query = func.websearch_to_tsquery("english", query)
+        keyword_score = func.ts_rank_cd(DatasetModel.search_vector, ts_query, 32).label(
+            "key_score"
+        )
+
+        alpha = 2 / 3
```
stijndehaes:
```suggestion
        embedding_search_weight = 2 / 3
        keyword_search_weight = 1 / 3
```

backend/app/data_products/output_ports/service.py
```diff
@@ -122,53 +120,115 @@ def get_dataset(
...
-        sql = recalculate_search_vector_datasets_statement()
-        result = self.db.execute(sql)
+                visible_candidates.append((dataset, score))
+
+                if len(visible_candidates) >= limit:
+                    split_index = limit
+                    break
+                elif (
+                    len(visible_candidates) > 1
+                    and (gap := score - visible_candidates[-1][1]) > max_gap
```
stijndehaes: So basically if the (current score - the previous score) > (previous score - previous previous score) we set split_index.

however we don't stop the looping, so if that same statement is true in the next step we update split_index again.

We we stop updating split_index when (current score - the previous score) < (previous score - previous previous score) for all the next items right?

Maybe we should have a link to a good explanation here as to why this is a good idea. I assume this is elbow method

backend/app/data_products/output_ports/search_dataset.py
stijndehaes, line 1: We need to put this back most likely

backend/app/data_products/router.py
stijndehaes: We could let it work like that, I just followed the template of how the previous search worked. But I agree flushing in the `recalculate_embeddings_for_output_ports_of_product` makes more sense

backend/app/database/alembic/versions/2026_01_07_1455-dbe6fbe45eef_add_entire_schema_checkbox.py
```diff
@@ -1,7 +1,7 @@
 """Add entire_schema checkbox
 
 Revision ID: dbe6fbe45eef
-Revises: ca5f5782790a
+Revises: 49990691cd30
```
stijndehaes: We probably shouldn't change this, but change our own migration to be the last no?

backend/tests/app/search_output_ports/test_router.py
```diff
@@ -1,35 +1,192 @@
...
+
+        output = GetDataProductOutputPortsResponse.model_validate(response.json())
+        assert len(output.output_ports) <= limit, "Query limit exceeded"
+
+        positives = {port.name for port in output.output_ports}
+        true_positives = len(expected.intersection(positives))
+        false_negatives = len(expected.difference(positives))
+
+        precision = true_positives / len(positives)
+        recall = true_positives / (true_positives + false_negatives)
```
stijndehaes: Isn't this just `true_positives / len(expected)`?

backend/tests/app/search_output_ports/test_router.py
```diff
@@ -1,35 +1,192 @@
...
+                "expected": {
+                    "Production Planning Forecast",
+                    "Production Variance Report",
+                },
+            },
+            {
+                "query": "inventory levels by warehouse",
+                "expected": {"Inventory Status"},
+            },
+        ]
```
stijndehaes, line 106: We need to update this when we update the test set I assume? Otherwise the scores might introduce failing tests all of a sudden no?

backend/app/data_products/output_ports/service.py
vlieven: Indeed elbow method, but I'm removing this again as discussed

backend/tests/app/search_output_ports/test_router.py
vlieven: Yes, but CI should ideally catch that when making updates :)

backend/app/database/alembic/versions/2026_01_07_1455-dbe6fbe45eef_add_entire_schema_checkbox.py
vlieven: We could also do that, but then the dates are out of order

backend/app/data_products/output_ports/search_dataset.py
vlieven: I am not sure - I think I was able to replace it by just a single function here
https://github.com/conveyordata/data-product-portal/pull/2477/changes#diff-3c76482d731f0f902b4932d4c2a321c7aad913ad7a5e354f8992057b1de7a4abR204

backend/app/data_products/output_ports/schema.py
stijndehaes: I would then make a search service, for sure now that we combine searches again

backend/app/database/alembic/versions/2026_01_07_1455-dbe6fbe45eef_add_entire_schema_checkbox.py
stijndehaes: I believe that this breaks the migrations however, because I assume that in the database the current revision is kept. By placing our revision in the middle, it will never get executed

PR #2500, opened 2026-01-21
Fix conditional hooks changed on rerender bug, bug in Databricks expected schema

stijn-janssens:


frontend/src/components/data-products/data-output-form/data-output-configuration-form.component.tsx
vlieven, line 54: This is not related to the PR, but `configurationFieldName(field.name) as any` should not be needed I think? If you have `configurationFieldName(...)` return a string, then this casting to any should not be required? The field name is always a string right?

stijn-janssens: will look into it

PR #2503, opened 2026-01-21
Feature/add data quality info

nclaeys:


backend/app/data_products/output_ports/data_quality/enums.py
```diff
@@ -0,0 +1,10 @@
+from enum import UNIQUE, Enum, verify
+
+
+@verify(UNIQUE)
+class DataQualityStatus(str, Enum):
```
vlieven: You can replace this by inheriting from StrEnum directly

```suggestion
class DataQualityStatus(StrEnum):
```

backend/app/data_products/output_ports/data_quality/enums.py
```diff
@@ -0,0 +1,10 @@
...
+    PASS = "pass"  # noqa: S105
```
vlieven: I think the QA warning is valid in that you're overwriting a python keyword here. Can we rename it to "SUCCESS" instead?

backend/app/data_products/output_ports/data_quality/model.py
```diff
@@ -0,0 +1,68 @@
...
+from app.database.database import Base
+from app.shared.model import utcnow
+
+
+class DataQualityTechnicalAssetModel(Base):
+    __tablename__ = "data_quality_technical_assets"
+
+    name: Mapped[str] = mapped_column(String(length=255), primary_key=True)
+    status: Mapped[DataQualityStatus] = mapped_column(
+        Enum(DataQualityStatus), default=DataQualityStatus.UNKNOWN
```
vlieven: Postgres Enums cannot be migrated (i.e. when we want to add a new state). We should just represent these as a string in the database, and map them to the enums in the application code.

backend/app/data_products/output_ports/data_quality/router.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+
+@router.post(
+    route,
+    responses={
+        404: {
+            "description": "Dataset not found",
+            "content": {
+                "application/json": {"example": {"detail": "Dataset id not found"}}
```
vlieven:
```suggestion
        404: {
            "description": "Output Port not found",
            "content": {
                "application/json": {"example": {"detail": "Output Port ID not found"}}
```

backend/app/data_products/output_ports/data_quality/enums.py
nclaeys: I think it is wrong as nor PASS nor "pass" is a keyword in python.I think it just wrongly matches the pass within a string.

backend/app/data_products/output_ports/data_quality/model.py
nclaeys: hmm that is annoying. I was just following the convention of what was done in the other tables in the codebase. So we need to change it everywhere?

backend/app/data_products/output_ports/data_quality/model.py
```diff
@@ -0,0 +1,68 @@
...
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from app.data_products.output_ports.data_quality.enums import DataQualityStatus
+from app.database.database import Base
+from app.shared.model import utcnow
+
+
+class DataQualityTechnicalAssetModel(Base):
+    __tablename__ = "data_quality_technical_assets"
```
stijndehaes, line 28: Wouldn't the reverse name make more sense `technical_assets_data_quality`? Just like we named `dataset_query_stats_daily`

stijndehaes: I just noticed where the confusion stems from, this actually has not link with the technical assets inside of Portal right? Hence we don't have a link but a new name. Maybe just name it `data_quality_assets`?

backend/app/data_products/output_ports/data_quality/model.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    # Relationships
+    data_quality_summary_id: Mapped[uuid.UUID] = mapped_column(
+        UUID(as_uuid=True),
+        ForeignKey("dataset_data_quality_summaries.id", ondelete="CASCADE"),
+        nullable=False,
+        primary_key=True,
+    )
+    data_quality_summary: Mapped["DataQualitySummary"] = relationship(
```
stijndehaes, line 44: Either we should make this `relationship` lazy, or if we never need it we can also just remove it for now no?

backend/app/data_products/output_ports/data_quality/router.py
```diff
@@ -0,0 +1,89 @@
...
+        404: {
+            "description": "Output Port not found",
+            "content": {
+                "application/json": {"example": {"detail": "Output Port ID not found"}}
+            },
+        }
+    },
+    dependencies=[
+        Depends(
+            Authorization.enforce(Action.DATASET__UPDATE_PROPERTIES, DatasetResolver)
```
stijndehaes: Should this be a different action?

backend/app/data_products/output_ports/data_quality/router.py
```diff
@@ -0,0 +1,89 @@
...
+        ),
+    ],
+)
+def add_output_port_data_quality_run(
+    data_product_id: UUID,
+    id: UUID,
+    data_quality_summary: OutputPortDataQualitySummary,
```
stijndehaes, line 96: If we want to correct one, can we supply a summary with that same date?

frontend/src/pages/dataset/components/dataset-quality/dataset-quality.component.tsx
```diff
@@ -0,0 +1,112 @@
...
+                return 'Error';
+            case 'warn':
+                return 'Warning';
+            case 'unknown':
+                return 'Unknown';
+        }
+    }
+
+    return (
+        <Flex vertical className={styles.sectionWrapper}>
```
stijndehaes:
```suggestion
        <Flex vertical gap="small">
```

styles.sectionWrapper only sets a gap, we can do that directly in ant design. Unsure if small is a good size though

frontend/src/store/api/services/generated/completeServiceApi.ts
```diff
@@ -1,7 +1,6 @@
-// @ts-nocheck
```
stijndehaes, line 1: Why is this removed? The autogeneration normally places this here, and I see no changes to `generate-api` command

integrations/data_quality/extract_dbt_summary.py
```diff
@@ -0,0 +1,89 @@
+import json
```
stijndehaes, line 1: Are these examples we offer, or a real integration?

Maybe we should add a readme to the `integration/data_quality` folder

backend/app/data_products/output_ports/data_quality/enums.py
nclaeys: I will change it.

integrations/data_quality/extract_dbt_summary.py
nclaeys: It is an example, documented in the docs.

backend/app/data_products/output_ports/data_quality/router.py
nclaeys: No, if you would want to correct one, I would need to implement a put request on the id. I can add that if you want it.

backend/app/data_products/output_ports/data_quality/router.py
nclaeys: I have no idea how to decide on actions for a dataset. In my idea it does not make sense to give permissions on the data quality of a dataset so I think it fits within the dataset permisisons.

backend/app/data_products/output_ports/data_quality/model.py
nclaeys: SqlAlchemy needs it, which is why it is there. I can make it raise to be inline with the rest

backend/app/data_products/output_ports/data_quality/model.py
nclaeys: It does not have a link to keep it more flexible atm. but they do refer the technical assets so I would not use a different name. I just fear that the exact mapping will not work which is why I do not link them.

frontend/src/store/api/services/generated/completeServiceApi.ts
nclaeys: because that is what the current command generates. It is the same on main.

backend/app/data_products/output_ports/data_quality/model.py
stijndehaes: Yeah I also think an exact mapping will not work, maybe we can add this explanation to the docs later?

backend/app/data_products/output_ports/data_quality/model.py
stijndehaes: Let's do lazy=raise then

backend/app/data_products/output_ports/data_quality/router.py
stijndehaes: Yeah maybe a question for @vlieven or @stijn-janssens , how do we decide which actions to assign to new calls? 
Maybe we can write this down as a dev guide article?

backend/app/data_products/output_ports/data_quality/router.py
stijndehaes: Don't think we need it for now, can't come up with a decent reason. So let's keep it like this

integrations/data_quality/extract_dbt_summary.py
stijndehaes: Missed that thanks!

backend/app/data_products/output_ports/data_quality/router.py
stijn-janssens: Actions have been created in a single snapshot up until now. We just looked at the state of portal and made a suggestion.
Output ports need their actions (probably) completely refined anyway after the rework. It could be a useful dev guide indeed. Currently there is not a guideline in place at all

backend/app/data_products/output_ports/data_quality/model.py
```diff
@@ -0,0 +1,77 @@
...
+    Text,
+)
+from sqlalchemy.dialects.postgresql import JSONB, UUID
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from app.database.database import Base
+from app.shared.model import utcnow
+
+
+class DataQualityTechnicalAssetModel(Base):
```
stijn-janssens: minor remark: to keep in sync with the rest of the codebase I would not call this class `...Model`. The suffix `Model` is only added when importing it next to a schema.

backend/app/data_products/output_ports/data_quality/model.py
```diff
@@ -0,0 +1,77 @@
...
+        nullable=False,
+        primary_key=True,
+    )
+    data_quality_summary: Mapped["DataQualitySummary"] = relationship(
+        lazy="raise", back_populates="technical_assets"
+    )
+
+
+class DataQualitySummary(Base):
+    __tablename__ = "dataset_data_quality_summaries"
```
stijn-janssens: shouldn't we use new terminology here?

backend/app/data_products/output_ports/data_quality/router.py
```diff
@@ -0,0 +1,98 @@
...
+
+def convert_dimensions_to_api(
+    dimensions: dict[str, str],
+) -> dict[str, DataQualityStatus]:
+    return {key: DataQualityStatus(value) for key, value in dimensions.items()}
+
+
+def convert_technical_assets_to_api(
+    technical_assets: list[DataQualityTechnicalAssetModel],
+):
```
stijn-janssens: miss a return type hint here

backend/app/data_products/output_ports/data_quality/model.py
nclaeys: I added this info in the code and also extend the docs explanation

backend/app/data_products/output_ports/data_quality/service.py
```diff
@@ -0,0 +1,83 @@
...
+    OutputPortDataQualitySummary,
+)
+from app.data_products.output_ports.model import ensure_dataset_exists
+
+
+def convert_dimensions_db(dimensions: dict[str, DataQualityStatus]) -> dict[str, str]:
+    return {key: value.value for key, value in dimensions.items()}
+
+
+class DatasetDataQualityService:
```
stijn-janssens: I think we should try and rename all the datasets here to output ports, unless you have a reason why we should keep the old terminology here for now?

backend/app/data_products/output_ports/data_quality/model.py
nclaeys: This is why suffixing models with Model makes things explicit. I find it confusing to always have 2 the same objects with the same name. For consistency I can remove it but I don't understand why we didn't use this pattern to start from.

backend/app/data_products/output_ports/data_quality/model.py
nclaeys: mainly consistency. I find it quite confusing to mix terminology everywhere. To be inline with the existing tables (datasets, usage of dataset,...) I kept that naming convention.

backend/app/data_products/output_ports/data_quality/router.py
vlieven: I think we want this to be an action that is machine-executable. So in the ideal case, we can generate a token, which only has this specific permission. I think in that light, it would be good to make this an explicit action. We have to revisit this when we redesign the output port access, but we can already add it I think?

backend/app/data_products/output_ports/data_quality/router.py
vlieven:
```suggestion
) -> Sequence[DataQualityTechnicalAsset]:
```

backend/app/data_products/output_ports/data_quality/service.py
```diff
@@ -0,0 +1,83 @@
...
+            ]
+        )
+        assets_with_issues = len(
+            [
+                a
+                for a in technical_assets
+                if a.status
+                in [DataQualityStatus.FAILURE.value, DataQualityStatus.ERROR.value]
+            ]
+        )
```
vlieven: Do we need the `.value` here? I would expect the comparison to just work against the enum directly?

backend/app/data_products/output_ports/data_quality/model.py
```diff
@@ -0,0 +1,77 @@
...
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from app.database.database import Base
+from app.shared.model import utcnow
+
+
+class DataQualityTechnicalAssetModel(Base):
+    """
+    We store the data quality for technical assets in a separate table instead of linking to technical assets table.
+    The reason for this is that creating an exact link will be difficult to achieve always.
```
stijndehaes:
```suggestion
    The reason for this is that creating an exact link will be difficult to always achieve.
```

backend/app/data_products/output_ports/data_quality/model.py
stijndehaes: Well then should we rename `data_quality_technical_assets` to use old terminolgy as well?

stijndehaes: To be honest it's very confusing, so I would just try to use new terminology for new things, makes the work when we do the rename a bit less.
In the mean time things will be confusing sadly

backend/app/data_products/output_ports/data_quality/router.py
nclaeys: Yes, I will add it.

backend/app/data_products/output_ports/data_quality/service.py
nclaeys: it is not necessary no. I will change it.

stijn-janssens (approved):

frontend/src/store/api/services/generated/completeServiceApi.ts
stijndehaes: Apparently is currently broken on main here is the fix: https://github.com/conveyordata/data-product-portal/pull/2579

The command currently fails when you run it

PR #2554, opened 2026-01-28
Migrate technical asset configurations to be stored in a table per configuration type instead of a big table

stijn-janssens:
In support of ADR 0009 #2409 

### DIFF
--- a/backend/app/database/alembic/versions/2026_01_28_1241_snowflake_separate_table.py
+++ b/backend/app/database/alembic/versions/2026_01_28_1241_snowflake_separate_table.py
@@ -1,7 +1,7 @@
 """Migrate Snowflake configurations to separate table
 
 Revision ID: snowflake_separate_table
-Revises: b1de5e9ba1fa
+Revises: 7601ac14662a
 Create Date: 2026-01-28 12:41:00.000000
 
 """
@@ -16,7 +16,7 @@ from app.shared.model import utcnow
 
 # revision identifiers, used by Alembic.
 revision: str = "snowflake_separate_table"
-down_revision: Union[str, None] = "b1de5e9ba1fa"
+down_revision: Union[str, None] = "7601ac14662a"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 


backend/app/data_output_configuration/base_model.py
```diff
@@ -9,12 +9,46 @@
...
+
+    The configuration_type field is used for:
+    1. Pydantic serialization (discriminator for Union types)
+    2. SQLAlchemy polymorphic loading (to load the correct subclass)
+    """
+
     __tablename__ = "data_output_configurations"
+
     id = Column(UUID(as_uuid=True), primary_key=True, default=uuid.uuid4)
     configuration_type: Mapped[DataOutputTypes] = Column(Enum(DataOutputTypes))
```
vlieven: I see an enum here - can we still swap that out?

backend/app/database/alembic/versions/2026_01_28_1241_snowflake_separate_table.py
```diff
@@ -0,0 +1,81 @@
...
+    # Migrate existing Snowflake data from polymorphic table to new table
+    # Note: The base table data_output_configurations already has the id and timestamps,
+    # we just need to copy the Snowflake-specific columns
+    op.execute(
+        """
+        INSERT INTO snowflake_data_output_configurations
+            (id, database, schema, "table", bucket_identifier, database_path, table_path, access_granularity, created_on, updated_on, deleted_at)
+        SELECT
+            id, database, schema, "table", bucket_identifier, database_path, table_path, access_granularity, created_on, updated_on, deleted_at
+        FROM data_output_configurations
```
vlieven, line 56: You are inserting into `snowflake_data_output_configurations` in the up migration, but not dropping rows from `data_output_configurations`. In the down migration, you are updating rows again in the `data_output_configurations`, but dropping new rows that were inserted after.

I have the feeling that this migration is not entirely correct. I would expect that the up migration removes the converted rows from `data_output_configurations` and that the down migration moves everything back from `snowflake_data_output_configurations` to `data_output_configurations` (including newly created rows). That way, you only have a single copy of a row living in the DB, avoiding confusion, and don't lose data in a down migration.

backend/app/database/alembic/versions/2026_01_28_1242_databricks_separate_table.py
```diff
@@ -0,0 +1,76 @@
...
+from app.shared.model import utcnow
+
+# revision identifiers, used by Alembic.
+revision: str = "databricks_separate_table"
+down_revision: Union[str, None] = "snowflake_separate_table"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
```
vlieven, line 24: Same remark as for the snowflake migration

backend/app/database/alembic/versions/2026_01_28_1243_s3_separate_table.py
```diff
@@ -0,0 +1,68 @@
...
+        sa.Column("bucket", sa.String(), nullable=True),
+        sa.Column("suffix", sa.String(), nullable=True),
+        sa.Column("path", sa.String(), nullable=True),
+        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
+        sa.Column("deleted_at", sa.DateTime(timezone=False), nullable=True),
+    )
+
+    # Migrate existing S3 data
+    op.execute(
```
vlieven, line 43: Same remark as for the snowflake table

backend/app/database/alembic/versions/2026_01_28_1244_glue_separate_table.py
```diff
@@ -0,0 +1,76 @@
...
+        sa.Column("database_path", sa.String(), nullable=True),
+        sa.Column("table_path", sa.String(), nullable=True),
+        sa.Column("access_granularity", sa.String(), nullable=True),
+        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
+        sa.Column("deleted_at", sa.DateTime(timezone=False), nullable=True),
+    )
+
+    # Migrate existing Glue data
+    op.execute(
```
vlieven, line 47: Same remark as for the snowflake migration

backend/app/database/alembic/versions/2026_01_28_1245_redshift_separate_table.py
```diff
@@ -0,0 +1,76 @@
...
+        sa.Column("database_path", sa.String(), nullable=True),
+        sa.Column("table_path", sa.String(), nullable=True),
+        sa.Column("access_granularity", sa.String(), nullable=True),
+        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
+        sa.Column("deleted_at", sa.DateTime(timezone=False), nullable=True),
+    )
+
+    # Migrate existing Redshift data
+    op.execute(
```
vlieven, line 47: Same remark as for the snowflake migration

backend/app/data_output_configuration/databricks/model.py
```diff
@@ -1,23 +1,20 @@
+from sqlalchemy import String
 from sqlalchemy.orm import Mapped, mapped_column
 
 from app.data_output_configuration.base_model import BaseDataOutputConfiguration
 
 
 class DatabricksDataOutput(BaseDataOutputConfiguration):
```
nclaeys: You are changing the tables anyway, is this not a good time to use the new naming convention?

backend/app/database/alembic/versions/2026_01_28_1241_snowflake_separate_table.py
stijn-janssens: up migration: fair point
down migration: this happens already in this migration, newly created rows still get an ID in the base table

backend/app/data_output_configuration/databricks/model.py
```diff
@@ -1,23 +1,20 @@
...
-    )
-    catalog_path: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
-    table: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
-    table_path: Mapped[str] = mapped_column(nullable=True, use_existing_column=True)
-    access_granularity: Mapped[str] = mapped_column(
-        nullable=True,
-        use_existing_column=True,
-    )
+class DatabricksTechnicalAssetConfiguration(BaseTechnicalAssetConfiguration):
+    __tablename__ = "databricks_technical_asset_configurations"
```
stijndehaes, line 8: Just a thought, if we name it `technical_asset_configuration_databricks` they are all nicely sorted when you open the database view, however it's a big rename so not sure if it's worth it

backend/app/data_output_configuration/base_model.py
```diff
@@ -1,20 +1,53 @@
...
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import Mapped
 
-from app.data_output_configuration.data_output_types import DataOutputTypes
 from app.database.database import Base
 from app.shared.model import BaseORM
 
 
-class BaseDataOutputConfiguration(Base, BaseORM):
+class TechnicalAssetConfiguration(Base, BaseORM):
```
stijndehaes, line 11: Is this table needed? I am unsure what it adds? Since you can't search on this table

backend/app/data_output_configuration/base_model.py
```diff
@@ -1,20 +1,53 @@
...
+    """
+    Abstract base class for specific configuration type tables.
+    Each configuration type (Snowflake, Databricks, etc.) inherits from this
+    and defines its own table with configuration-specific columns.
+
+    The 'id' column in each child table is both a primary key AND a foreign key
+    back to data_output_configurations.id, creating a Class Table Inheritance pattern.
+    """
+
+    __abstract__ = True  # This is an abstract base class, not a table
```
vlieven, line 49: Why does this need to be defined through a dunder variable instead of the standard python [abc](https://docs.python.org/3/library/abc.html) module? Or is this a sqlalchemy thing?

backend/app/database/alembic/versions/2026_01_28_1241_snowflake_separate_table.py
```diff
@@ -0,0 +1,112 @@
...
+
+    op.execute(
+        """
+        UPDATE data_output_configurations
+        SET configuration_type = 'SnowflakeTechnicalAssetConfiguration'
+        WHERE configuration_type = 'SnowflakeDataOutput'
+        """
+    )
+
+    # Remove Snowflake-specific columns from base table to avoid duplicate data
```
stijndehaes, line 69: It says remove byt entries are not removed, they are set to null?
@vlieven's comment is still valid, was there a discussion you had I am missing by any chance? :)

backend/app/data_output_configuration/base_model.py
stijn-janssens: https://docs.sqlalchemy.org/en/21/orm/declarative_config.html#abstract
Derived it from this, so this definitely looks like the standard way of doing it in sqlalchemy. Will test it out with abc

stijn-janssens: Update: It is a sqlalchemy thing and ABC does not work. __abstract__ informs SQLAlchemy to not map the class to a table. abc does no such thing and only enforces the class to be uninstantiatable.

vlieven: Yes, indeed. Confusing way of implementing this IMO, but not something we can fix :)

backend/app/data_output_configuration/base_model.py
stijn-janssens: This table is still used. It handles the polymorphism and correct mapping of technical asset configurations to the table of their specific type.

backend/app/database/alembic/versions/2026_01_28_1241_snowflake_separate_table.py
stijn-janssens: Because of the above comment, this table still holds all the IDs (and new IDs are also added here) together with the correct mapping of the type to the specific configuration table

PR #2579, opened 2026-02-02
Fix generate api command

stijndehaes:

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,7 +15,7 @@
         "validate-translations": "i18next 'src/**/*.{ts,tsx}' --config i18next-parser.config.js --fail-on-update",
         "typecheck": "tsc --strict --noEmit",
         "dep-viz": "npm run build && npx open-cli dependencies.html",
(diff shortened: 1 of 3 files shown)


frontend/package.json
```diff
@@ -15,7 +15,7 @@
         "validate-translations": "i18next 'src/**/*.{ts,tsx}' --config i18next-parser.config.js --fail-on-update",
         "typecheck": "tsc --strict --noEmit",
         "dep-viz": "npm run build && npx open-cli dependencies.html",
-        "generate-api": "rm -rf src/store/api/services/generated/* && npx @rtk-query/codegen-openapi src/store/api/openapi-config.ts && sed -i '' '1s/^/\\/\\/ @ts-nocheck\\n/' src/store/api/services/generated/completeServiceApi.ts"
+        "generate-api": "rm -rf src/store/api/services/generated/* && npx @rtk-query/codegen-openapi src/store/api/openapi-config.ts && sed -i '1s/^/\\/\\/ @ts-nocheck\\n/' src/store/api/services/generated/completeServiceApi.ts"
```
stijn-janssens: I introduced the previous "fix" because the new command gives the following error for me.
`sed: 1: "src/store/api/services/ ...": bad flag in substitute command: 'v'`

stijndehaes: Issue is apparently with me having installed `gnu-sed` via brew, and setting this as the default. So let's look at another solution

stijndehaes: Changed it to a javascript script

PR #2602, opened 2026-02-04
Feature/data product studio

stijn-janssens:
Please do play around with the page to see how it feels

Scherm­afbeelding 2026-02-04 om 13 42 56
Scherm­afbeelding 2026-02-04 om 13 43 05
Scherm­afbeelding 2026-02-04 om 13 43 12
Scherm­afbeelding 2026-02-04 om 13 43 20

### DIFF
--- a/frontend/src/components/layout/navbar/breadcrumbs/breadcrumbs.component.tsx
+++ b/frontend/src/components/layout/navbar/breadcrumbs/breadcrumbs.component.tsx
@@ -30,7 +30,10 @@ export const Breadcrumbs = () => {
     const { pathname } = useLocation();
     const params = useParams<DynamicPathParams>();
     const pathnames = useMemo(
-        () => (pathname === ApplicationPaths.Home ? [ApplicationPaths.Home] : pathname.split('/').filter((x) => x)),
+        () =>
+            pathname === ApplicationPaths.Home
+                ? [ApplicationPaths.Home]
+                : pathname.split('/').filter((x) => x && x !== 'output-port'),
         [pathname],
     );
     const { dataProductId = '', datasetId = '', dataOutputId = '' } = params;
@@ -62,7 +65,6 @@ export const Breadcrumbs = () => {
                 path,
                 title: pathname,
             };
-
             switch (path) {
                 case ApplicationPaths.Studio:
                     Object.assign(breadcrumbItem, {
@@ -139,7 +141,6 @@ export const Breadcrumbs = () => {
                     Object.assign(breadcrumbItem, {
                         title: '',
                     });
-
                     // Case for Data Product and Output Port
                     if (
                         dataProduct &&
@@ -148,7 +149,8 @@ export const Breadcrumbs = () => {
                     ) {
                         if (
                             isDataProductEditPage(path, dataProduct.id) ||
-                            (dataOutput && isDataOutputEditPage(path, dataOutput.id, dataProduct.id))
+                            (dataOutput && isDataOutputEditPage(path, dataOutput.id, dataProduct.id)) ||
+                            (dataset && isDatasetEditPage(path, dataset.id))
                         ) {
                             Object.assign(breadcrumbItem, {
                                 title: t('Edit'),
@@ -167,17 +169,31 @@ export const Breadcrumbs = () => {
                                     ),
                                 });
                             } else {
-                                Object.assign(breadcrumbItem, {
-                                    path: `${path}#${DataProductTabKeys.About}`,
-                                    title: (
-                                        <Typography.Text
-                                            ellipsis={{ tooltip: dataProduct.name }}
-                                            rootClassName={styles.title}
-                                        >
-                                            {dataProduct.name}
-                                        </Typography.Text>
-                                    ),
-                                });
+                                if (dataset && !isFetchingDataset && path.split('/').length === 4) {
+                                    Object.assign(breadcrumbItem, {
+                                        path: `${path}#${DatasetTabKeys.About}`,
+                                        title: (
+                                            <Typography.Text
+                                                ellipsis={{ tooltip: dataset.name }}
+                                                rootClassName={styles.title}
+                                            >
+                                                {dataset.name}
+                                            </Typography.Text>
+                                        ),
+                                    });
+                                } else {
+                                    Object.assign(breadcrumbItem, {
+                                        path: `${path}#${DataProductTabKeys.About}`,
+                                        title: (
+                                            <Typography.Text
+                                                ellipsis={{ tooltip: dataProduct.name }}
+                                                rootClassName={styles.title}
+                                            >
+                                                {dataProduct.name}
+                                            </Typography.Text>
+                                        ),
+                                    });
+                                }
                             }
                         }
                     }
--- a/frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
+++ b/frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
@@ -15,7 +15,7 @@ import { useCheckAccessQuery } from '@/store/features/authorization/authorizatio
 import { useGetAllDatasetsQuery, useGetUserDatasetsQuery } from '@/store/features/datasets/datasets-api-slice.ts';
 import { AuthorizationAction } from '@/types/authorization/rbac-actions';
 import type { DatasetsGetContract } from '@/types/dataset/datasets-get.contract';
-import { ApplicationPaths, createDatasetIdPath } from '@/types/navigation.ts';
+import { ApplicationPaths, createOutputPortPath } from '@/types/navigation.ts';
 import styles from './output-ports-tab.module.scss';
 import { getOutputPortTableColumns } from './output-ports-table-columns';
 
@@ -110,8 +110,8 @@ export function OutputPortsTab() {
     );
     const { data: access } = useCheckAccessQuery({ action: AuthorizationAction.GLOBAL__CREATE_DATAPRODUCT });
     const canCreateDataProduct = access?.allowed ?? false;
-    const navigateToOutputPort = (datasetId: string) => {
-        navigate(createDatasetIdPath(datasetId));
+    const navigateToOutputPort = (dataProductId: string, outputPortId: string) => {
+        navigate(createOutputPortPath(dataProductId, outputPortId));
     };
 
     const handleRoleChange = (selected: { productIds: string[]; roles: string[] }) => {
@@ -164,7 +164,7 @@ export function OutputPortsTab() {
             {/* Table */}
             <Table
                 onRow={(record) => ({
-                    onClick: () => navigateToOutputPort(record.id),
+                    onClick: () => navigateToOutputPort(record.data_product_id, record.id),
                 })}
                 columns={columns}
                 dataSource={filteredOutputPorts}
--- a/frontend/src/routes.tsx
+++ b/frontend/src/routes.tsx
@@ -75,6 +75,14 @@ const router = createBrowserRouter([
                                 path: ApplicationPaths.DataProductEdit,
                                 element: <DataProductEdit />,
                             },
+                            {
+                                path: ApplicationPaths.OutputPort,
+                                element: <Dataset />,
+                            },
+                            {
+                                path: ApplicationPaths.OutputPortEdit,
+                                element: <DatasetEdit />,
+                            },
                         ],
                     },
                     {
--- a/frontend/src/types/navigation.ts
+++ b/frontend/src/types/navigation.ts
@@ -15,6 +15,8 @@ export enum ApplicationPaths {
     DatasetEdit = '/datasets/:datasetId/edit',
     Marketplace = '/marketplace',
     MarketplaceCart = '/marketplace/cart',
+    OutputPort = '/studio/:dataProductId/output-port/:datasetId',
+    OutputPortEdit = '/studio/:dataProductId/output-port/:datasetId/edit',
     DataOutput = '/studio/:dataProductId/:dataOutputId',
     DataOutputEdit = '/studio/:dataProductId/:dataOutputId/edit',
     AuditLogs = '/audit-logs',
@@ -44,6 +46,17 @@ export function createDataOutputIdPath(
 
 export function createDatasetIdPath(datasetId: string, tabKey: DatasetTabKeys = DatasetTabKeys.About) {
     return `${ApplicationPaths.Dataset.replace(':datasetId', encodeURIComponent(datasetId))}#${tabKey}`;
+}
+
+export function createOutputPortPath(
+    dataProductId: string,
+    datasetId: string,
+    tabKey: DatasetTabKeys = DatasetTabKeys.About,
+) {
+    return `${ApplicationPaths.OutputPort.replace(':dataProductId', encodeURIComponent(dataProductId)).replace(
+        ':datasetId',
+        encodeURIComponent(datasetId),
+    )}#${tabKey}`;
 }
 
 export enum DynamicPathParams {


frontend/src/components/filters/role-filter.component.tsx
```diff
@@ -68,89 +64,40 @@ export function RoleFilter({ selectedRole, onRoleChange, mode }: RoleFilterProps
...
-                                {' > '}
-                                <span>{selectedRoleLabel}</span>
-                            </>
-                        )}
-                        <DownOutlined />
-                    </Button>
-                </Dropdown>
-            </Radio.Button>
-        </Radio.Group>
+        <Select
```
stijndehaes, line 93: What happens when we select like 5 things?

vlieven:
Functional / visual feedback:
- I would like to use a new icon to accompany the "Product Studio" label. Maybe one of `<ReconciliationOutlined />`, `<DashboardOutlined />`, `<DesktopOutlined />`, `<ProductOutlined />` can be a good candidate. Maybe @katarinamil can weigh in?
- The Output Ports tab has the column "Users", with the value "Not used". I would change this to "Consumers" and "No consumers yet".
- On the Data Products tab, it looks like the "Filter by role" selector is only functional when combined with the "My Data Products" button, correct? If find the current UI a bit unintuitive in the sense that if I'm on "All Data Products", by selecting a role, the button also switches to "My Data Products". Maybe something we could try is swapping the order, so you get ["All Data Products", "My Data Products", "Filter by role"] in the same component group, and then only showing "Filter by role" when "My Data Products" is selected?
- It's a bit strange that there is no "Create Data Output" button on the list page, although that can also be added in a follow-up PR.
- I am not sure about the added value of the "Consumes" and "Produces" columns for the "Data Product" tab.
- I would rename the "Access" column on the "Data Product" tab to "Team", and say "x members" instead of "x users". Ideally the word "user" never shows up in our application :)

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,157 @@
...
+    const navigate = useNavigate();
+    const currentUser = useSelector(selectCurrentUser);
+
+    const [searchTerm, setSearchTerm] = useQueryState('dp-search', parseAsString.withDefault(''));
+    const [showAllProducts, setShowAllProducts] = useQueryState('dp-showAll', parseAsBoolean.withDefault(false));
+    const [selectedRoles, setSelectedRoles] = useState<string[]>([]);
+    const [selectedProductIds, setSelectedProductIds] = useState<string[]>([]);
+
+    const { data: userDataProducts = [], isFetching: isFetchingUserProducts } = useGetUserDataProductsQuery(
+        currentUser?.id || '',
```
vlieven:
```suggestion
        currentUser?.id ?? '',
```

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,157 @@
...
+    };
+
+    return (
+        <Flex vertical gap="small">
+            {/* Search and Actions */}
+            <Flex justify="space-between" align="center">
+                <Flex gap="middle" flex={1} align="center">
+                    <Input.Search
+                        placeholder={t('Search data products by name')}
+                        value={searchTerm || ''}
```
vlieven:
```suggestion
                        value={searchTerm ?? ''}
```

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,157 @@
...
+                rowClassName={styles.row}
+                columns={columns}
+                dataSource={filteredDataProducts}
+                pagination={{
+                    size: 'small',
+                    showTotal: (total, range) =>
+                        t('Showing {{range0}}-{{range1}} of {{total}} data products', {
+                            range0: range[0],
+                            range1: range[1],
+                            total: total,
```
vlieven:
```suggestion
                        t('Showing {{range0}}-{{range1}} of {{count}} data products', {
                            range0: range[0],
                            range1: range[1],
                            count: total,
```

frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    const [searchTerm, setSearchTerm] = useQueryState('op-search', parseAsString.withDefault(''));
+    const [showAllPorts, setShowAllPorts] = useQueryState('op-showAll', parseAsBoolean.withDefault(false));
+    const [selectedRoles, setSelectedRoles] = useState<string[]>([]);
+    const [selectedPortIds, setSelectedPortIds] = useState<string[]>([]);
+    const [isInitialized, setIsInitialized] = useState(false);
+
+    // Fetch user's role assignments to find the Owner role
+    const { data: userDatasetRoles } = useListOutputPortRoleAssignmentsQuery(
+        { userId: currentUser?.id || '' },
```
vlieven:
```suggestion
        { userId: currentUser?.id ?? '' },
```

frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+            setSelectedRoles([ownerRoleId]);
+            setSelectedPortIds(ownerPortIds);
+        }
+
+        setIsInitialized(true);
+    }, [userDatasetRoles, isInitialized]);
+
+    const { data: userOutputPorts = [], isFetching: isFetchingUserPorts } = useGetUserDatasetsQuery(
+        currentUser?.id || '',
```
vlieven:
```suggestion
        currentUser?.id ?? '',
```

frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
```diff
@@ -0,0 +1,173 @@
...
+    };
+
+    return (
+        <Flex vertical gap="small">
+            {/* Search and Filters */}
+            <Flex justify="space-between" align="center">
+                <Flex gap="middle" flex={1} align="center">
+                    <Input.Search
+                        placeholder={t('Search output ports by name or data product')}
+                        value={searchTerm || ''}
```
vlieven:
```suggestion
                        value={searchTerm ?? ''}
```

frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
```diff
@@ -0,0 +1,173 @@
...
+                })}
+                columns={columns}
+                dataSource={filteredOutputPorts}
+                pagination={{
+                    size: 'small',
+                    showTotal: (total, range) =>
+                        t('Showing {{range0}}-{{range1}} of {{total}} output ports', {
+                            range0: range[0],
+                            range1: range[1],
+                            total: total,
```
vlieven:
```suggestion
                        t('Showing {{range0}}-{{range1}} of {{count}} output ports', {
                            range0: range[0],
                            range1: range[1],
                            count: total,
```

frontend/src/utils/quality-status.helper.ts
```diff
@@ -0,0 +1,53 @@
...
+            return 'Failed';
+        case 'error':
+            return 'Error';
+        case 'warning':
+            return 'Warning';
+        case 'unknown':
+        default:
+            return 'Unknown';
+    }
+};
```
vlieven, line 54: I think these need to be translated?

frontend/src/routes.tsx
```diff
@@ -71,6 +71,10 @@ const router = createBrowserRouter([
                             },
                         ],
                     },
+                    {
+                        path: ApplicationPaths.Studio,
+                        element: <Navigate to={ApplicationPaths.DataProducts} />,
```
vlieven: I think this needs to be the other way around. The old location (DataProducts) should navigate to the new one (Studio).

stijn-janssens: No, because then the child pages break. Unless we make big changes to our whole routing setup the studio URL is not used for the moment

katarinamil:
> Functional / visual feedback:
> 
> * I would like to use a new icon to accompany the "Product Studio" label. Maybe one of `<ReconciliationOutlined />`, `<DashboardOutlined />`, `<DesktopOutlined />`, `<ProductOutlined />` can be a good candidate. Maybe @katarinamil can weigh in?

icon for Product studio: I suggest (in prioritised order) ProductOutlined, ControlOutlined, ToolOutlined, BuildOutlined, AddstoreAddOutlined. 

> * The Output Ports tab has the column "Users", with the value "Not used". I would change this to "Consumers" and "No consumers yet".

Agree

> * On the Data Products tab, it looks like the "Filter by role" selector is only functional when combined with the "My Data Products" button, correct? If find the current UI a bit unintuitive in the sense that if I'm on "All Data Products", by selecting a role, the button also switches to "My Data Products". Maybe something we could try is swapping the order, so you get ["All Data Products", "My Data Products", "Filter by role"] in the same component group, and then only showing "Filter by role" when "My Data Products" is selected?

Looking at the latest screenshot the order is already changed? @stijn-janssens but I agree to show Filter only when my data products is selected, otherwise it makes no sense.

> * It's a bit strange that there is no "Create Data Output" button on the list page, although that can also be added in a follow-up PR.

I vote we redesign OP page for Producer / Consumer in a separate story, I planned brainstorm for this.

> * I am not sure about the added value of the "Consumes" and "Produces" columns for the "Data Product" tab.

Agree, what is shown seems clutter to me, but I do think in the future we should think about how to show in that overview already what is the usage of DP (OP has it, but I'd like at least some indicator for DP owner of the usage - now they need to navigate to the OP of that DP to see it).

> * I would rename the "Access" column on the "Data Product" tab to "Team", and say "x members" instead of "x users". Ideally the word "user" never shows up in our application :)

I very much agree with this.

Adding also: 
- I'd call quality Health
- All products are in Draft so I am not sure if they would not be in Draft do we show the status as Green? Would recommend that if it's not the case.
- Empty states: can you screenshot what's the empty state right now? If they chose My data products and they don't have any or All Data Products and there are no DP, could we show a "Create Data Product" button with a friendly message? or that's already the case?

frontend/src/routes.tsx
vlieven: Can't you just use the refactoring tool in your IDE for that? Rename `ApplicationPaths.DataProducts` to `ApplicationPaths.Studio`, change the string value, and then add `ApplicationPaths.DataProducts` back as fallback? IMO we should always have old stuff being backed by new stuff instead of vice-versa. Otherwise you can never get rid of legacy content

frontend/src/components/filters/role-filter.component.tsx
stijn-janssens: Scherm­afbeelding 2026-02-09 om 10 51 59

frontend/src/components/layout/navbar/breadcrumbs/breadcrumbs.component.tsx
```diff
@@ -144,7 +145,8 @@ export const Breadcrumbs = () => {
                     if (
                         dataProduct &&
                         !isFetchingDataProduct &&
-                        pathnames.includes(ApplicationPaths.DataProducts.replace('/', ''))
+                        (pathnames.includes(ApplicationPaths.DataProducts.replace('/', '')) ||
+                            pathnames.includes(ApplicationPaths.Studio.replace('/', '')))
```
vlieven: This only still needs to be Studio I think? Since Everything in DataProducts gets redirected?

frontend/src/pages/dataset/components/dataset-quality/dataset-quality.component.tsx
```diff
@@ -28,62 +20,17 @@ export function DatasetQuality({ dataProductId, datasetId }: Props) {
...
-
     return (
         <Flex vertical gap="small">
             <Typography.Title level={5}>{t('Quality Status')}</Typography.Title>
             {summary ? (
                 <Flex vertical gap="small" align="flex-start">
                     <Tag
-                        icon={getStatusIcon(summary.overall_status)}
-                        color={getStatusColor(summary.overall_status)}
+                        icon={<CustomSvgIconLoader iconComponent={getQualityStatusIcon(summary.overall_status)} />}
```
vlieven: Should the `getQualityStatusIcon()` not just return a correct `<Icon>` directly?
(I think the <CustomSvgIconLoader> should be inside the `getQualityStatusIcon`)

vlieven: Or you could even extract everything inside the <Tag> block into it's own component? That might even make more sense

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+function filterDataProductsByRoles(dataProducts: DataProductsGetContract, selectedProductIds: string[]) {
+    if (!selectedProductIds.length) {
+        return dataProducts;
+    }
+
+    return dataProducts.filter((dataProduct) => {
+        return selectedProductIds.includes(dataProduct.id);
+    });
+}
```
vlieven, line 39: It's a bit confusing for me that we have `DataProductsGetContract` as an alias for something that (based on the implementation) is a list of something. Can't we have something like `DataProduct[]` as the type here? And also use that to annotate the result type.

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,188 @@
...
+        navigate(createDataProductIdPath(dataProductId));
+    };
+
+    const handleRoleChange = (selected: { productIds: string[]; roles: string[] }) => {
+        setSelectedProductIds(selected.productIds);
+        setSelectedRoles(selected.roles);
+        // Switch to "My Data Products" when roles are selected
+        if (selected.roles.length > 0) {
+            setShowAllProducts(false);
+        }
```
vlieven: This should no longer be necessary I think?

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,188 @@
...
+                            mode={'data_products'}
+                            selectedRoles={selectedRoles}
+                            onRoleChange={handleRoleChange}
+                        />
+                    )}
+                </Flex>
+                {filteredDataProducts.length > 0 && createButton}
+            </Flex>
+
+            {/* Pending Requests Section */}
```
vlieven:
```suggestion
```

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+            {/* Table */}
+            <Table<DataProductsGetContract[0]>
```
vlieven, line 138: Would something like this not be more clear?

```suggestion
            <Table<DataProduct>>
```

frontend/src/pages/product-studio/components/output-ports-tab/output-ports-tab.component.tsx
```diff
@@ -0,0 +1,213 @@
...
+        navigate(createDatasetIdPath(datasetId));
+    };
+
+    const handleRoleChange = (selected: { productIds: string[]; roles: string[] }) => {
+        setSelectedPortIds(selected.productIds);
+        setSelectedRoles(selected.roles);
+        // Switch to "My Output Ports" when roles are selected
+        if (selected.roles.length > 0) {
+            setShowAllPorts(false);
+        }
```
vlieven: Also here I think this is not needed anymore?

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
stijn-janssens: I propose to leave this like this for this PR. This should and shall be tackled when we migrate the frontend to v2 and generate all of the types. These are still the old types where sequences of products can be comprised of random shenanigans. Fine for you? If not I change it here but I feel it is not part of this change.

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
stijn-janssens: It would be more clear but it is not the same logic :P The product**s** get call returns some calculated fields which are needed in the columns. And the table has as row type a single entry of the get**s** call. Makes sense?

frontend/src/types/navigation.ts
```diff
@@ -4,18 +4,21 @@ import { TabKeys as DatasetTabKeys } from '@/pages/dataset/components/dataset-ta
...
+    DataProductNew = '/studio/new',
+    DataProductEdit = '/studio/:dataProductId/edit',
     Datasets = '/datasets',
     Dataset = '/datasets/:datasetId',
     DatasetEdit = '/datasets/:datasetId/edit',
     Marketplace = '/marketplace',
     MarketplaceCart = '/marketplace/cart',
-    DataOutput = '/data-products/:dataProductId/:dataOutputId',
-    DataOutputEdit = '/data-products/:dataProductId/:dataOutputId/edit',
+    OutputPort = '/studio/:dataProductId/output-port/:datasetId',
```
stijn-janssens, line 18: I added these new "product studio output port" routes.
For now I only navigate there directly from within the product studio. Other output ports still navigate to the marketplace.
This can be followed up in a next PR or when we design the new Producer view for output ports

nclaeys (approved):

frontend/src/pages/product-studio/components/data-products-tab/data-products-tab.component.tsx
vlieven: Ok, fine for me to leave it like it is for now

PR #2603, opened 2026-02-04
Use consistent capitalization for key concepts across the frontend

vlieven:
- [/] Update the release notes document if needed in the [release notes](../docs/docs/release-notes.md)
- [/] When updating the UI please provide screenshots or videos to the reviewers

### DIFF
--- a/frontend/src/components/search-page/search-page.component.tsx
+++ b/frontend/src/components/search-page/search-page.component.tsx
@@ -44,7 +44,9 @@ export default function SearchPage({
                 />
             </Col>
             <Col span={14}>{actions}</Col>
-            <Col span={10}>{searchSuggestions && !searched && <SearchSuggestions suggestions={searchSuggestions} />}</Col>
+            <Col span={10}>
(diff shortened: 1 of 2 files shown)


frontend/src/components/search-page/search-page.component.tsx
```diff
@@ -44,9 +44,7 @@ export default function SearchPage({
                 />
             </Col>
             <Col span={14}>{actions}</Col>
-            <Col span={10}>
-                {searchSuggestions && !searched && <SearchSuggestions suggestions={searchSuggestions} />}
-            </Col>
+            <Col span={10}>{!searched && <SearchSuggestions suggestions={searchSuggestions} />}</Col>
```
stijndehaes: I would revert this change, to not add extra stuff to the dom when searchsuggestions is undefined

stijndehaes (changes requested):
Maybe we can add a test to ensure we keep using the proper capitalisation everywhere?

PR #2614, opened 2026-02-05
Stabilize OpenAPI JSON

vlieven:
- [/] Update the release notes document if needed in the [release notes](../docs/docs/release-notes.md)
- [/] When updating the UI please provide screenshots or videos to the reviewers

Rename class definitions so that they are globally unique instead of locally


stijn-janssens (commented):
As discussed separately, I find the change in sqlalchemy models a bit strange / unfitting for this PR. Because in theory these can never be included in the openapi spec and should therefor have no impact.
Ofcourse I might be wrong! Will be investigated

PR #2626, opened 2026-02-06
Added more search suggestions

stijndehaes:
Also ensure we start at a random one
to ensure people won't get bored


frontend/src/components/search-page/search-suggestions.tsx
```diff
@@ -11,6 +12,11 @@ export function SearchSuggestions({ suggestions }: Props) {
     const { t } = useTranslation();
     const { token } = theme.useToken();
 
+    //Start on a random suggestion so people won't get bored by seeing the same suggestion over and over
+    const initialSuggestions = useMemo(() => {
+        return Math.floor(Math.random() * suggestions.length);
+    }, [suggestions]);
```
vlieven: I would add a type hint to indicate this is a number, and make it singular (it's an index, not a collection)

```suggestion
    const initialSuggestion: number = useMemo(() => {
        return Math.floor(Math.random() * suggestions.length);
    }, [suggestions]);
```

PR #2637, opened 2026-02-09
Generate our front end api with enum support

stijndehaes:
Makes the enums nicer to use


stijndehaes:
> We have to see how to best handle this for the AuthorizationActions, as we manually define the mapping of the integer codes to the name of the action. Maybe no changes are required, but I would double check that we can still use the named actions instead of the number in the application code.

Removed them from enum generation for now

frontend/src/pages/settings/components/settings-tabs/components/data-product-settings-table/new-data-product-setting-modal.component.tsx
```diff
@@ -209,8 +210,13 @@ export function CreateSettingModal({ isOpen, onClose, scope, mode, initial }: Pr
         }
     }, [form, mode, canEditNamespace, namespaceSuggestion]);
 
-    const validateNamespaceCallback = useCallback(
-        (namespace: string) => validateNamespace({ namespace, scope }).unwrap(),
+    const validateNamespaceCallback: (namespace: string) => Promise<NamespaceValidationResponse> = useCallback(
```
stijndehaes, line 213: For the reviewer, this is a workaround, in a next PR the namespace component will change to use the new method

PR #2639, opened 2026-02-09
Migrate resource names

stijndehaes:
I migrated the backend, but kept the front end names still namespace. 
I propose to switch this later


backend/app/resource_names/router.py
```diff
@@ -24,12 +24,12 @@
 router = APIRouter(tags=["Resource names"], prefix="/v2/resource_names")
 
 
-@router.post("/sanitize")
+@router.get("/sanitize")
```
stijndehaes, line 24: Switched from post to get, since these don't really change any resource

backend/app/resource_names/service.py
```diff
@@ -19,7 +19,7 @@ class ResourceNameValidityType(str, Enum):
     VALID = "VALID"
     INVALID_LENGTH = "INVALID_LENGTH"
     INVALID_CHARACTERS = "INVALID_CHARACTERS"
-    DUPLICATE_NAMESPACE = "DUPLICATE_NAMESPACE"
+    DUPLICATE = "DUPLICATE"
```
stijndehaes, line 22: Changed to not contain namespace again, this is a breaking change but I think it's ok. SInce it's only used in front end I believe.
Should I add this to the release notes?

vlieven: I don't think anyone is depending on this technical detail

backend/app/resource_names/router.py
vlieven: They indeed do not change any resources, but I believe that post was taken because of the default caching behavior (GET caches, POST does not). For now the`/sanitize` is idempotent, so caching should be possible, but for `validate`, the validity depends on the state of the DB to check if something is unique or not. I don't know if GET should be preferred over POST here?

stijndehaes: Well I don't think it's a good idea to make this a post just because rtk query caches get requests.
Maybe it's better to tell rtk query to not cache this request? Then again, the result should also be the same when using the same input no?

stijndehaes: I can disable the cache in the front end to be sure, but don't think it's necessary

PR #2627, opened 2026-02-06
Feature/missing frontend plugin dependencies

stijn-janssens:
There was still a hidden hardcoded Platforms file in the frontend.
This PR removes the dependency and introduces a "platform plugin only conveyor plugin" Not too big of a fan of how it is implemented / works but functionally nothing changed and it helped me forward at least.

Still todo but will do on a different branch: Remove the hardcoded URL logic of what happens when clicking on a tile.
This might mean I make the platform tiles their own separate concept. To be seen

- [x] Update the release notes document if needed in the [release notes](../docs/docs/release-notes.md)
- [x] When updating the UI please provide screenshots or videos to the reviewers

PR #2652, opened 2026-02-10
Feature/backend url in plugins

stijn-janssens:
Builds further on #2627
Removes the dependency of URL triggers on action tiles out of the frontend and into the responsibility of the backend plugins.
Removes the need for the individual API calls on data product level.


backend/app/configuration/environments/platform_configurations/service.py
```diff
@@ -23,3 +26,26 @@ def get_environment_platform_config(
             EnvironmentPlatformConfigurationModel.platform_id == platform_id,
         )
         return self.db.scalar(stmt)
+
+    def get_env_platform_config(self, environment: str, platform_name: str) -> str:
+        environment_model = self.db.scalar(
```
nclaeys: Can't you rewrite this in 1 query instead of 3? A bit as follows:

```
stmt = (
    select(EnvironmentPlatformConfigurationModel)
    .join(
        EnvironmentModel,
        EnvironmentPlatformConfigurationModel.environment_id == EnvironmentModel.id,
    )
    .join(
        PlatformModel,
        EnvironmentPlatformConfigurationModel.platform_id == PlatformModel.id,
    )
    .where(
        EnvironmentModel.name == environment,
        PlatformModel.name == platform_name,
    )
)
```
Not sure if this is valid sqlalchemy code though 😅

backend/app/core/aws/get_url.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _get_data_product_role_arn(id: UUID, environment: str, db: Session) -> str:
+    environment_context = (
+        db.execute(select(EnvironmentModel).where(EnvironmentModel.name == environment))
+        .scalar_one()
+        .context
+    )
+    namespace = db.get(DataProductModel, id).namespace
+    role_arn = environment_context.replace("{{}}", namespace)
```
vlieven, line 41: What does `"{{}}"` signify exactly? Shouldn't we use a templating engine for this?

backend/app/core/aws/get_url.py
```diff
@@ -0,0 +1,74 @@
...
+    }
+    json_dump = json.dumps(url_credentials)
+
+    request_parameters = "?Action=getSigninToken"
+    SESSION_DURATION = 900
+    request_parameters += f"&SessionDuration={SESSION_DURATION}"
+    request_parameters += f"&Session={parse.quote_plus(json_dump)}"
+    request_url = "https://signin.aws.amazon.com/federation" + request_parameters
+
+    r = httpx.get(request_url)
```
vlieven: I think you can simply construct the params instead of building the entire request url through string manipulation

```suggestion
aws_signin_url = "https://signin.aws.amazon.com/federation"
r = httpx.get(aws_signin_url, params={
    "Action": "getSigninToken",
    "SessionDuration": 900,
    "Session": json_dump
})
```

backend/app/core/aws/get_url.py
```diff
@@ -0,0 +1,74 @@
...
+    signin_token = json.loads(r.text)
+
+    request_parameters = "?Action=login"
+    request_parameters += f"&Issuer={settings.HOST}"
+    athena_link = "https://console.aws.amazon.com/athena/home#/query-editor"
+    request_parameters += f"&Destination={parse.quote_plus(athena_link)}"
+    request_parameters += f"&SigninToken={signin_token['SigninToken']}"
+    request_url = "https://signin.aws.amazon.com/federation" + request_parameters
+
+    return request_url
```
vlieven: Similar change here, but use httpx only to construct the URL without sending a request

```suggestion
    request = httpx.Request('GET', aws_signin_url, params={
        "Action": "login",
        "Issuer": settings.HOST,
        "Destination": "https://console.aws.amazon.com/athena/home#/query-editor",
        "SigninToken": signin_token["SigninToken"],
    })
    
    return str(request.url)
```

backend/app/data_output_configuration/databricks/schema.py
```diff
@@ -75,6 +81,22 @@ def get_configuration(
...
+        platform_config = EnvironmentPlatformConfigurationService(
+            db
+        ).get_env_platform_config(environment, "Databricks")
+        data_product = db.get(DataProductModel, id)
+        config = json.loads(platform_config)["workspace_urls"]
+        if str(data_product.domain_id) not in config:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail=(
+                    f"Workspace not configured fordomain {data_product.domain.name}"
```
vlieven:
```suggestion
                    f"Workspace not configured for domain {data_product.domain.name}"
```

backend/app/data_output_configuration/service.py
```diff
@@ -88,6 +92,46 @@ def get_platform_tiles(
...
 
+    def get_url(
+        self,
+        plugin_name: str,
+        id: UUID,
+        db: Session,
+        actor: "User",
+        environment: Optional[str] = None,
+    ) -> str:
+        data_output_configurations = AssetProviderPlugin.__subclasses__()
```
vlieven, line 108: Just some thought on doing things this way:
- This requires that the subclasses are loaded in the python interpreter already - how is this set up?
- I believe __subclasses__ only looks to direct subclasses. Probably not an issue now, but if we ever introduce a shared baseclass, this will no longer work I think. You don't need to resolve anything for this comment, just making sure you are aware :)

frontend/src/pages/data-product/components/data-product-actions/data-product-actions.component.tsx
```diff
@@ -94,75 +89,44 @@ export function DataProductActions({ dataProductId }: Props) {
         return null;
     }
 
-    async function handleAccessToData(environment: string, dataPlatform: string) {
+    // TODO: This function should make use of the environment string. If the environment string is available (i.e. Not conveyor) then the direct tile click should be disabled.
```
vlieven: What is keeping you from resolving this TODO?

stijn-janssens: woops, already resolved

backend/app/data_output_configuration/service.py
stijn-janssens: Correct, unfortunately this means that all the plugins need to be loaded in the __init__.py file. If not then they might not be picked up. The only solution I can think of that does not need this is a full fledged plugin system

backend/app/core/aws/get_url.py
stijn-janssens: this is legacy code, it is used to replace parts of an aws role with the namespace. Templating would indeed be better here, but this has quite some impact (needs a db migration etc.) so if fine for you, I log it as tech debt and keep it like this here. To keep this PR focussed

PR #2664, opened 2026-02-11
Feature/full v2 migration

stijndehaes:
- [x] When updating the UI please provide screenshots or videos to the reviewers -> This PR is pretty big but nothing really changes for users

TODOs:
- [x] Finish tags migrations 
- [x] Test out edits on every page
- [x] Fix breadcrumbs
- Migrate left over calls -> For a follow up PR
- Add list output ports call seperate from search -> For a follow up PR


backend/app/data_products/output_ports/service.py
```diff
@@ -511,3 +518,40 @@ def get_output_ports(self, data_product_id: Optional[UUID]) -> Sequence[OutputPo
             ensure_data_product_exists(data_product_id, self.db)
             query = query.filter(DatasetModel.data_product_id == data_product_id)
         return self.db.scalars(query).unique().all()
+
+    def get_data_outputs_for_data_product(self, id: UUID) -> Sequence[DataOutputGet]:
```
stijn-janssens: why is this placed in the output ports service?

stijndehaes: Thanks for spotting I moved this to the incorrect location will move it to to technical assets service where it belongs

stijndehaes: It was already there so removed this one

PR #2715, opened 2026-02-19
Remove old get all datasets calls

stijndehaes:
Replaced by the revamped search call, revamped
because it returns less resulting in a dramatic speedup. The call now takes about 150ms, compared to 5s before.

In the cart we probably need another solution, because the current solution only scales to 1000 datasets, which is currently plenty. The previous solution would also run into issues when running tons of datasets because the response would get too big and/or returning
10000 datasets does not make sense.


backend/app/data_products/output_ports/model.py
```diff
@@ -102,19 +108,23 @@ def quality_status(self) -> Optional[str]:
...
-            for link in self.data_product_links
-            if link.status == DecisionStatus.APPROVED
-        ]
-        return len(accepted_product_links)
-
     @property
     def data_product_name(self) -> str:
         return self.data_product.name
 
+    data_product_count = column_property(
```
stijn-janssens, line 116: is this more efficient? Because it is less readable imo

stijn-janssens: Also this changes the functionality I think, because you no longer filter on status = `Approved` . Is this intended?

backend/app/database/alembic/versions/2026_02_19_1545-bb31223f935e_dataset_index_on_name.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "bb31223f935e"
+down_revision: Union[str, None] = "2bac081ab9d0"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_index("idx_dataset_name", "datasets", ["name"], unique=False)
```
stijn-janssens, line 21: why this new index?

backend/app/data_products/output_ports/model.py
stijndehaes: It fetches a lot less data, this shaved off about 1.5seconds I believe. And you are right the logic is changed, I will add the filter back.
I got confused because of the naming, would have expected something like `approved_data_product_count`

backend/app/database/alembic/versions/2026_02_19_1545-bb31223f935e_dataset_index_on_name.py
stijndehaes: When you use the search without any query, it lists all items alphabetically, without this index that is slow.

PR #2772, opened 2026-02-25
Improve the loading speed of explorer

stijndehaes:
The main improved is to use raw queries,
for explorer this is the right approach since
we want very minimal info and using the sql
alchemy models is way to bloated. We are fetching
lots of data products etc.

A single call now takes about 50ms instead of 10s
of seconds.

Some functionality got removed, we now do not fetch the domains as nodes. We did not use that in the front end anyway. We also do not show the assignment anymore, if we want to add that I propose to fetch the relevant data product or dataset assignments when the user clicks. It would make the graph info bloated again.


stijndehaes:
It loads very quickly now, the explorer view on product or output port is way slower than this at the moment.
image

PR #2808, opened 2026-03-01
Fix reversed arrows in Explorer Data Products view

peeterskris:

## Summary
- Swap `source` and `target` in edge creation for the "Data Products" only view so arrows point from Producer → Consumer, consistent with the "All" view

## Screenshots

**Before** (Data Products view — arrows point Consumer → Producer, reversed):
Arrows incorrectly point from right to left in the "Data Products" tab.

**After** (Data Products view — arrows point Producer → Consumer, matching "All" view):
Arrows now consistently point from left to right (Producer → Consumer) in both views.

## Test plan
- [x] Existing graph service tests pass (4/4)
- [x] Visually verify in the portal: arrows should point the same direction in both "All" and "Data Products" views

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/tests/app/graph/test_router.py
+++ b/backend/tests/app/graph/test_router.py
@@ -57,9 +57,7 @@ class TestGraphRouter:
         producer = DataProductFactory()
         dataset = DatasetFactory(data_product=producer)
         consumer = DataProductFactory()
-        DataProductDatasetAssociationFactory(
-            data_product=consumer, dataset=dataset
-        )
+        DataProductDatasetAssociationFactory(data_product=consumer, dataset=dataset)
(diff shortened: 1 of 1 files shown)


backend/app/graph/service.py
```diff
@@ -162,9 +162,9 @@ def get_graph_data(
             edges.extend(
                 [
                     Edge(
-                        id=f"{link['consumer_id']}-{link['producer_id']}",
-                        source=link["consumer_id"],
-                        target=link["producer_id"],
+                        id=f"{link['producer_id']}-{link['consumer_id']}",
```
stijndehaes, line 165: Can we add a test to verify the behavior so we ensure this does not happen again?

PR #2810, opened 2026-03-01
Enable domain visualization on graph explorer

peeterskris:
## Summary
- Activate the existing but disabled domain container node infrastructure in the Explorer view
- Synthesize domain parent nodes from the `domain_id` and `domain` fields already returned by the API, assigning each domain a distinct transparent color from an 8-color palette
- Pass `domainsEnabled` through to `parseRegularNode` so child nodes are nested inside their domain boxes via `parentId`
- Keep domain container nodes visible during node selection highlighting when any child is connected

## Test plan
- [ ] Open http://localhost:8080/explorer and confirm colored domain boxes appear around grouped data products/datasets
- [ ] Confirm the sidebar Domains filter toggle hides/shows domain boxes
- [ ] Confirm node selection highlighting still works — domain boxes stay visible when a child node is selected
- [ ] Confirm ELK layout handles sizing/positioning cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/src/components/global-explorer/internal-explorer.tsx
+++ b/frontend/src/components/global-explorer/internal-explorer.tsx
@@ -1,10 +1,10 @@
 import '@xyflow/react/dist/base.css';
 
+import { G2 } from '@ant-design/plots';
 import type { Edge, Node } from '@xyflow/react';
 import { Position, useReactFlow } from '@xyflow/react';
 import { Flex, theme } from 'antd';
 import { type MouseEvent, useCallback, useEffect, useState } from 'react';
-
 import { defaultFitViewOptions, NodeEditor } from '@/components/charts/node-editor/node-editor.tsx';
 import { CustomEdgeTypes, CustomNodeTypes } from '@/components/charts/node-editor/node-types.ts';
 import type { Node as GraphNode } from '@/store/api/services/generated/graphApi.ts';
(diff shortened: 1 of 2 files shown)


frontend/src/components/global-explorer/internal-explorer.tsx
```diff
@@ -16,9 +16,49 @@ import { parseEdges } from '../explorer/utils';
 import { Sidebar, type SidebarFilters } from './sidebar/sidebar';
 import { useNodeEditor } from './use-node-editor';
 
-function parseFullNodes(nodes: GraphNode[], setNodeId: (id: string) => void): Node[] {
+const DOMAIN_COLORS = [
+    { bg: 'rgba(59, 130, 246, 0.08)', border: 'rgba(59, 130, 246, 0.3)' }, // blue
```
stijn-janssens: Fine for now, but maybe this should be logged as tech debt as this both hardcodes the colors and will recycle them after 8 domains.

frontend/src/utils/node-parser.helper.test.ts
stijn-janssens, line 1: tests should be placed in the tests folder

backend/tests/app/graph/test_router.py
```diff
@@ -67,3 +68,38 @@ def test_data_products_only_arrow_points_producer_to_consumer(self, client):
         assert edge["target"] == str(consumer.id), (
             "Edge target should be the consumer (dataset reader)"
         )
+
```
stijndehaes: I am unsure what the tests add to be honest, the logic did not change? So unsure why they are required. Can you explain to me the reasoning for adding these?

The includes domain fields tests for example could just be the last 2 asserts that are added to `test_get_graph_data`.

frontend/src/components/global-explorer/internal-explorer.tsx
stijndehaes: We already use antv from ant design, I believe there might be a colorpallette there. I will look into it and respond here again.

frontend/src/utils/node-parser.helper.test.ts
stijndehaes: No in front end testing, it is preferred to place it next to the component

stijn-janssens: :o that feels weird :D But okay! Learned something new

backend/tests/app/graph/test_router.py
```diff
@@ -20,13 +21,17 @@ class TestGraphRouter:
         ],
     )
     def test_get_graph_data(self, client, route):
```
stijndehaes, line 23: This I like a lot more, enhancing an existing test to ensure domain data is included

frontend/src/components/global-explorer/internal-explorer.tsx
stijndehaes: So after some investigation there is no easy way to get the colorpallete out of antv, to ensure however we have consistent styling it might make sense to copy the colors of ant and add them to our theme. That way we can also use them here.
WDYT @stijn-janssens ?

stijndehaes: Figured out how to get the category20 from ant design

PR #2865, opened 2026-03-09
Fix bad auth resolve on create route

stijn-janssens:

### DIFF
--- a/backend/tests/app/authorization/role_assignments/data_product/test_router.py
+++ b/backend/tests/app/authorization/role_assignments/data_product/test_router.py
@@ -46,16 +46,13 @@ class TestDataProductRoleAssignmentsRouter:
         assert data[0]["id"] == str(assignment.id)
 
     @pytest.mark.parametrize(
-        ("permissions", "should_update_casbin"),
+        "permissions",
         [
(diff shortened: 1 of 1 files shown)


backend/tests/app/authorization/role_assignments/data_product/test_router.py
```diff
@@ -100,6 +101,63 @@ def test_create_assignment(
         else:
             mock_data_product_auth_assignment.assert_not_called()
 
+    @pytest.mark.parametrize(
+        ("permissions", "should_update_casbin"),
```
stijndehaes: `should_update_casbin` is basically because I have `DATA_PRODUCT__APPROVE_USER_REQUEST` it is automatically approved? Maybe we should rename it to something better, because this took me a while to understand :D 

perhaps:
- `role_assignment_automatically_approved`?

stijn-janssens: correct! Will rename and merge

stijn-janssens: actually I don't even need that var..

PR #2882, opened 2026-03-11
ADR for provisioner architecture

pascal-knapen:

### DIFF
--- a/docs/adr/0013-provisioner-architecture.md
+++ b/docs/adr/0013-provisioner-architecture.md
@@ -1,67 +1,43 @@
-# Provisioner Architecture: Fluent Builder vs. Abstract Base Class SDK
+# Provisioner Architecture: Backend Event Adapter + Abstract Base Class SDK
 
-## Context and Problem Statement
+## How It Works Today
 
-The current provisioner (`demo/basic/provisioner/provisioner/main.py`) is a flat FastAPI application with no formal abstraction, no typed inputs, and no SDK. New provisioner authors must copy the demo, understand the internal `Router` class, and manually handle raw webhook payloads. There is no documented contract, no type safety, and no helper client for calling back into the Portal API.
+The Portal backend fires a generic webhook after every mutating API call. The payload is a raw forward of the HTTP exchange:
 
-This ADR proposes an SDK approach (`dpp-sdk`) that:
-
-- Provides a clear, typed interface for all provisioner lifecycle events
-- Wraps Portal API calls in a high-level client (`PortalClient`)
-- Exposes a standard FastAPI `app` object that authors serve themselves
-
-A key design principle, inspired by how [Agno](https://docs.agno.com) serves agents via `agent_os.get_app()`, is that the SDK returns a plain FastAPI application rather than owning the server lifecycle. Authors stay in control: they can pass the app to uvicorn, mount it inside an existing FastAPI app, add middleware, or combine it with other routes. The SDK handles routing and deserialisation; the author handles deployment.
-
-## Decision Drivers
-
-* **Developer experience**: new provisioner authors need a clear, discoverable starting point
-* **Type safety**: handler functions receive rich objects, not raw dicts
-* **Coverage**: all mutating Portal events should be addressable via named hooks
-* **Composability**: the SDK produces a standard FastAPI `app` — authors control how it is served
-* **Transport flexibility**: hook implementations must be transport-agnostic so they can be driven by webhooks today and a message queue tomorrow without code changes
-* **Consistency**: align with the base-class pattern established in ADR-0009
-
-## Considered Options
-
-* **Option 1: Fluent Builder API** — a `Provisioner` object configured by chaining `.on_*()` calls, exposing `.get_app()`
-* **Option 2: Abstract Base Class SDK** — a `BaseProvisioner` class that authors subclass, overriding only the hooks they need, exposing `.get_app()`
-
-## Decision Outcome
-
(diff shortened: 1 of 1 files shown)


docs/adr/0013-provisioner-architecture.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Context and Problem Statement
+
+The current provisioner (`demo/basic/provisioner/provisioner/main.py`) is a flat FastAPI application with no formal abstraction, no typed inputs, and no SDK. New provisioner authors must copy the demo, understand the internal `Router` class, and manually handle raw webhook payloads. There is no documented contract, no type safety, and no helper client for calling back into the Portal API.
+
+This ADR proposes an SDK approach (`dpp-sdk`) that:
+
+- Provides a clear, typed interface for all provisioner lifecycle events
+- Wraps Portal API calls in a high-level client (`PortalClient`)
+- Exposes a standard FastAPI `app` object that authors serve themselves
```
stijndehaes: Why to they need this? Is this because we call webhooks? And this FastAPI is the way to serve these?

I don't think I understand it :)

docs/adr/0013-provisioner-architecture.md
```diff
@@ -0,0 +1,264 @@
...
+
+A key design principle, inspired by how [Agno](https://docs.agno.com) serves agents via `agent_os.get_app()`, is that the SDK returns a plain FastAPI application rather than owning the server lifecycle. Authors stay in control: they can pass the app to uvicorn, mount it inside an existing FastAPI app, add middleware, or combine it with other routes. The SDK handles routing and deserialisation; the author handles deployment.
```
stijndehaes: This also means I don't understand this

stijndehaes (changes requested):
So I don't understand the premise or I might be not sure. Can you try to explain it differently to me?
It makes revieweing the rest hard :) 

I think what you want is to make an SDK that allows people to make a FAST API application that implements a all the endpoints needed for our webhook architecture to call. Is that correct?

If that is true I would propose to try to generate this out of an open api spec, that way people could use that in other programming languages too if they want to. No?

docs/adr/0013-provisioner-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+    .get_app()
+)
+```
+
+Good: handlers are plain functions, easier to unit-test in isolation. No class inheritance required.
+Bad: no natural home for shared state; gaps in hook coverage are silent; discoverability depends on docs rather than IDE autocomplete.
+
+## Transport Extensibility
+
+Currently events arrive via **synchronous HTTP webhooks**: the Portal calls the provisioner directly after each mutation. This is simple but fragile — if the provisioner is temporarily unavailable, the event is lost.
```
stijndehaes, line 106: We could also store the events in the database, so you can request all events in the past X days rather easily. But unsure if we should add that

docs/adr/0013-provisioner-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+    .on_create_data_product(handle_create)
+    .on_approve_link(handle_approve)
+    .get_app()
+)
+```
+
+Good: handlers are plain functions, easier to unit-test in isolation. No class inheritance required.
+Bad: no natural home for shared state; gaps in hook coverage are silent; discoverability depends on docs rather than IDE autocomplete.
+
+## Transport Extensibility
```
stijndehaes, line 104: Users can already translate the webhook into a kafka or sqs message if they want to but then they have to implement the SDK themselves.

docs/adr/0013-provisioner-architecture.md
```diff
@@ -0,0 +1,258 @@
...
+- Backend enriches webhook payloads with full domain objects before dispatching to provisioners
+- Delivered as installable Python package `dpp-provisioner-sdk`
+- Demo provisioner (`demo/basic/provisioner/`) rewritten to subclass `BaseProvisioner`
+- Package README shows a minimal from-scratch example
+
+---
+
+## Appendix A: Full Event Catalogue
+
+All hooks receive typed objects rather than raw dicts. The `DataProduct` context is always present; `OutputPort` and `TechnicalAsset` are included where the event is scoped to them.
```
stijndehaes, line 138: Can we make a description of what is available in `DataProduct`, `OutputPort` and `TechnicalAsset`?
Are these full blown objects, or just id references?

docs/adr/0013-provisioner-architecture.md
pascal-knapen: Indeed, that's how Ranger handles it as well.

docs/adr/0013-provisioner-architecture.md
pascal-knapen: Correct

docs/adr/0013-provisioner-architecture.md
pascal-knapen: No sure yet tbh. Still need to discover what makes it easy to use. Was considering keeping them close to the more manageable v2 objects.

docs/adr/0013-provisioner-architecture.md
stijndehaes: I think that might be for another improvement

docs/adr/0013-provisioner-architecture.md
stijndehaes: I think that might be for another improvement

docs/adr/0013-provisioner-architecture.md
stijndehaes: I would propose to start with as small as an object as possible to make it easy to keep consistent? 

We can always add fields later.

pascal-knapen: Agreed

PR #2937, opened 2026-03-19
Add banner for pending requests to home page

vlieven:

## Dataminded theme
Screenshot 2026-03-31 at 15 23 35

## Green theme
Screenshot 2026-03-31 at 14 49 22

### DIFF
--- a/frontend/src/pages/home/components/quick-actions/quick-actions.tsx
+++ b/frontend/src/pages/home/components/quick-actions/quick-actions.tsx
@@ -48,7 +48,7 @@ export function QuickActions() {
     ];
 
     return (
-        <section>
+        <>
             <div style={{ marginBottom: 24 }}>
                 <Title level={3} style={{ margin: 0, marginBottom: 4 }}>
                     {t('What would you like to do today?')}
@@ -86,6 +86,6 @@ export function QuickActions() {
                     </Col>
                 ))}
             </Row>
-        </section>
+        </>
     );
 }


frontend/src/pages/home/components/alert-banner/alert-banner.tsx
```diff
@@ -0,0 +1,64 @@
...
+            classNames={{
+                root: styles.alertRoot,
+                icon: styles.alertIcon,
+                title: styles.alertText,
+            }}
+            title={
+                <Trans t={t} count={actionsCount}>
+                    You have{' '}
+                    <Link strong style={{ color: 'inherit' }} onClick={navigateToRequests}>
+                        {/* @ts-expect-error TS2353 */}
```
stijn-janssens: what is the error?

frontend/src/pages/home/components/alert-banner/alert-banner.tsx
```diff
@@ -0,0 +1,64 @@
...
+            icon={<BellOutlined />}
+            classNames={{
+                root: styles.alertRoot,
+                icon: styles.alertIcon,
+                title: styles.alertText,
+            }}
+            title={
+                <Trans t={t} count={actionsCount}>
+                    You have{' '}
+                    <Link strong style={{ color: 'inherit' }} onClick={navigateToRequests}>
```
stijn-janssens: why have you chosen to make only this text a link and not the entire banner?

frontend/src/styles/_colors.scss
```diff
@@ -177,3 +177,8 @@ $color-text: var(--ant-color-text);
 $base-text-color: var(--ant-color-text-base);
 $base-text-bg-base: var(--ant-color-bg-base);
 $color-white: var(--ant-color-white);
+
+// Alerts
+$color-background-alert: #F5F3FF;
```
stijn-janssens: Should we add this to the theme instead?

vlieven: I tried to, but I could not figure out how to add custom keys to a theme

frontend/src/pages/home/components/alert-banner/alert-banner.tsx
vlieven: Personal preference. Do you want me to make the entire banner a button?

frontend/src/pages/home/components/alert-banner/alert-banner.tsx
vlieven: Typescript does not understand the string interpolation, and claims that {{actionsCount}} is wrong. Will add it to the comment

frontend/src/pages/home/components/alert-banner/alert-banner.tsx
stijn-janssens: Personal preference :D Maybe @katarinamil can make the call here :)

frontend/src/styles/_colors.scss
stijn-janssens: Alright, fine for now. Maybe we can use a non-custom key?
Don't want to overcomplicate anything but also want to make sure that colours are adjustable in a single place. Not adding it to the theme kindof breaks that. But probably we have already sinned against this anyways :D

vlieven: Yeah, I have similar issues with the cards feature. Google and AI also don't provide any good pointers...

frontend/src/pages/home/components/alert-banner/alert-banner.tsx
katarinamil: Let’s go with making the entire banner clickable. It’s more intuitive for the user to click anywhere on the alert to navigate, rather than aiming for the specific text link. Thanks for the catch, @stijn-janssens!

@vlieven I'd rather if possible have the entire banner (not just text) wrapped as a link and not button, as it serves navigation. What's happening on hovering over? Would be nice if that shows it's interactable.

vlieven: Not sure I follow here... If the entire surface should be clickable, semantically it's a button? 
On hover, your cursor turns into a pointer if it's interactable.

vlieven: Also, should I remove the "View requests" action then?

Also, note that what's currently there, is following how ant design components are designed to behave. If you want me to create a banner that functions as a button, but is not styled as one, it breaks that design. Just pointing it out, not saying it can't be done.

katarinamil: to clarify: I'd like something that looks like a clickable card. And when you click it should behave like a semantic link (brings you to a different url), rather than like button (performs action like save/delete etc). Feels weird for me to click on a button in order to be placed to a different url. Is there a way that this could be done without breaking that design system logic?

I'd also like us to keep the View requests CTA, like a visual explanation to what will happen when they click.

vlieven: Ok, applied

frontend/src/pages/home/components/alert-banner/alert-banner.module.scss
```diff
@@ -0,0 +1,21 @@
...
+
+  margin: globals.$spacing-l globals.$spacing-xxl globals.$spacing-s;
+
+  &:hover {
+    cursor: pointer;
+  }
+}
+
+.alert-icon {
+  color: colors.$color-text-alert !important;
```
stijndehaes: Are the importants needed?

frontend/src/pages/home/components/alert-banner/alert-banner.module.scss
```diff
@@ -0,0 +1,21 @@
+@use "@/styles/colors";
+@use "@/styles/globals";
+
+.alert-root {
+  border-color: colors.$color-border-alert;
+  background-color: colors.$color-background-alert;
+
+  margin: globals.$spacing-l globals.$spacing-xxl globals.$spacing-s;
```
stijndehaes: Why the 3 different spacings? Also I am wondering if they are needed?

frontend/src/styles/_colors.scss
stijndehaes: Can't we change the info alert color of the team? According to the docs of ant desing `colorInfo`, `colorInfoBg` and `colorInfoBorder` are used. I think adding these extra colors makes it look out of place

frontend/src/pages/home/home.page.tsx
stijndehaes, line 36: I don't understand why not everything is in the Row system? This would solve the margin bottom added to all the components. Because the gutter becomes the margin between elements

frontend/src/pages/home/components/alert-banner/alert-banner.module.scss
```diff
@@ -0,0 +1,9 @@
+@use "@/styles/globals";
+
+.alert-root {
+  margin-bottom: globals.$spacing-ml;
```
stijndehaes, line 4: I don't think this margin is needed when we fix the home page see other comment

PR #2968, opened 2026-03-24
add basic test setup with tests for people page

nclaeys:


frontend/src/tests/mocks/handlers.ts
```diff
@@ -0,0 +1,12 @@
+// Define handlers that catch the corresponding requests and returns the mock data.
+import { HttpResponse, http } from 'msw';
+
+export const handlers = [
+    http.get('*/api/v2/authz/roles/:scope', () => {
+        return HttpResponse.json({ roles: [] });
+    }),
+
+    http.get('*/api/v2/authz/access/:action', () => {
```
stijndehaes: How would we handle unit tests where things are hidden or disabled when the user has no access to something?

frontend/src/tests/mocks/users.ts
```diff
@@ -0,0 +1,34 @@
+import { HttpResponse, http } from 'msw';
+import type { GetUsersResponse, UsersGet } from '@/store/api/services/generated/usersApi';
+import { server } from '@/tests/mocks/server.ts';
+
+export const mockUsers: UsersGet[] = [
```
stijndehaes, line 5: Wouldn't it be nicer if we have a function to generate a random user? A bit similar to the factories in the backend. I would make this a new ticket btw

frontend/src/tests/pages/people/people-table.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        mockUsersHttp([]);
+        renderWithProviders(<PeoplePage />);
+
+        await waitFor(() => {
+            expect(screen.getByText('People')).toBeInTheDocument();
+        });
+
+        const table = screen.getByTestId('people-table');
+        const description = table.querySelector('.ant-empty-description');
+        expect(description).toHaveTextContent('No data');
```
stijndehaes, line 34: Instead of using test selector can't we just check for the text No Data?

frontend/src/tests/mocks/handlers.ts
nclaeys: You can always extract this in a function to allowAllAuthz which you use by default. For those situations manage the calls explicitly.

frontend/src/tests/mocks/users.ts
nclaeys: I don't know if it makes things easier in the frontend. What is the benefit for having a random user here? You are mocking anyway all request to backend so you have no state persistence issues

frontend/src/tests/pages/people/people-table.test.tsx
nclaeys: I found it not detailed enough. GetByText doesn't work because you have 2 places where no data is displayed so it fails. I found this cleaner, what is wrong with using a testId?

stijndehaes: The idea with unit tests is to use the application like a user would, and to avoid test id's because a user wouldn't.

But I guess in this case you aren't inputting stuff anyway, so it doesn't really matter.

frontend/src/tests/mocks/users.ts
stijndehaes: It makes it easy to generate 20 of them, to check pagination for example. But maybe there are other ways to handle that?

nclaeys: That is true, it can easily be added if we want.

PR #3002, opened 2026-03-26
Rename access_type public to unrestricted.

nclaeys:


frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -73,14 +73,6 @@ const DEBOUNCE = 500;
 
 const getAccessTypeOptions = (t: TFunction) => {
```
stijndehaes, line 73: Why is the exaplanation for private removed?

nclaeys: my bad will rename the correct one.

PR #3003, opened 2026-03-26
Migrate AWS CLI to v2

stijndehaes:


stijn-janssens:
As discussed in person: Logic seems to partially work only. The login page is still opened and a token is fetched, but my test call did not work

stijn-janssens (approved):

Issue #2651, opened 2026-02-10
Support Dagre v2.x


stijn-janssens:
## Summary
Dependabot makes updates to Dagre v2 which breaks our frontend compile.
Please fix so we can stay up to date with latest versions

### Motivation
Security and dependency management

PR #3018, opened 2026-03-27
Upgrade dagre to 2.0.0

stijndehaes:
Also ensure explorers are always imported lazily
this reduces the main bundle size

Fixes: #2651


frontend/src/utils/node-editor.helper.ts
```diff
@@ -53,8 +53,3 @@ export const generateDagreLayout = (
 
     return { nodes, edges };
 };
-
-export const defaultFitViewOptions: FitViewOptions = {
```
stijn-janssens, line 57: Have you tested all of the current functionality? Both internal and full explorer? I think this zoom was needed (it used to be at least)

stijndehaes: This is an unused variable, so it doesn't do anything. No functionallity changed after removing this

PR #3020, opened 2026-03-27
Improve user search by filtering out accents etc

stijndehaes:


frontend/tsconfig.json
```diff
@@ -23,5 +23,5 @@
             "@/*": ["./src/*"]
         }
     },
-    "include": ["src", "vite.config.ts"]
+    "include": ["src", "vite.config.ts", "node_modules/vitest/globals.d.ts"]
```
vlieven: Why is this include needed?

stijndehaes: It's needed to be able to use `expect` in a test

PR #3125, opened 2026-04-09
Add azure platform

nclaeys:


backend/app/database/alembic/versions/2026_04_09_1307-3a883d3ec0f6_extend_platform_services_supported.py
```diff
@@ -0,0 +1,80 @@
...
+        sa.text(
+            """
+            UPDATE platform_services
+            SET technical_info_template = '{bucket_arn}/{suffix}/{path}/*'
+            WHERE name = 'S3'
+            """
+        )
+    )
+
+    azure_id = session.execute(
```
stijn-janssens, line 36: I would not add this in a migration. Adding it in the database means the tiles are rendered in the technical asset form which for some installations should not happen for Azure, or for Databricks for example.
We might want to rework this logic so that platform entries in the database are not correlated to tiles showing up in the UI but for now it does so I would remove this from the migration until rework and instead handle it in the sample data only

frontend/src/components/data-products/data-output-form/data-output-form.component.tsx
```diff
@@ -298,6 +299,12 @@ export function DataOutputForm({ mode, formRef, dataProductId, modalCallbackOnSu
                     showSearch={{ filterOption: selectFilterOptionByLabel }}
                 />
             </Form.Item>
+            <Form.Item<TechnicalAssetsCreateForm>
```
stijn-janssens: I believe this fix is now no longer needed

backend/app/database/alembic/versions/2026_04_09_1307-3a883d3ec0f6_extend_platform_services_supported.py
nclaeys: as discussed, checked it and the enabled_plugins solves this. Not enabled plugins are not visible in the UI.

frontend/src/components/data-products/data-output-form/data-output-form.component.tsx
nclaeys: good catch. Forgot about this.

stijn-janssens (approved):
Love the new integration documentation! 🚀 
PR looks good. As discussed we have some follow up steps but we won't tackle those in this PR.
E.g. we can remove the global platform configs as it is duplicated information

nclaeys:
> Love the new integration documentation! 🚀 PR looks good. As discussed we have some follow up steps but we won't tackle those in this PR. E.g. we can remove the global platform configs as it is duplicated information

Fixed the issues. Was related to incorrect foreign key definitions for the configs.

PR #3135, opened 2026-04-10
Remove old v1 from data products routers

stijn-janssens:
Continuing my quest

Some logic changed in this PR or needs to be verified:
I placed TODOs which need to be checked.
- history for both TA and OP: TA / OP does not need to exist in order to fetch history -> allow history to be fetched for deleted items
- usage endpoint removed, I think it has been fully replaced by query stats
- to be checked: the output ports endpoint does not correctly filter out private output ports and should therefor no longer be used? Did we take this into account correctly @stijndehaes ?


stijndehaes:
`history for both TA and OP: TA / OP does not need to exist in order to fetch history -> allow history to be fetched for deleted items` -> Do we need this? I think we don't there should be a global endpoint to fetch history where you can filter on objects and that should cover that? I would remove the TODO because this would be part of making the audit page no?


`to be checked: the output ports endpoint does not correctly filter out private output ports and should therefor no longer be used? Did we take this into account correctly @stijndehaes ?`
-> we should probably add that filtering, but for me the concept is broken, why is your data product not private? And should we have a check on that?

stijn-janssens:
The deleted objects can already occur in the history page of detailed data products. E.g. create a TA and delete it -> shows up in the history of the parent DP even though the TA itself no longer exists. We had unit test for this on the old endpoint and I think it makes sense to have it on the new ones as well? e.g. `def test_history_event_created_on_data_output_deletion(self, client):`

stijn-janssens:
I agree with the fact that the concept is at this stage not well-designed. I do think we need to take a more drastic approach of either not supporting it right now, or fixing the endpoints. The current "it works a little bit, but weird" is not a good solution

stijndehaes (approved):

PR #3142, opened 2026-04-13
Chore/fix frontend api generation

stijn-janssens:
Frontend API generation is drifting out of sync with actual api spec.


scripts/check_generated_api.sh
```diff
@@ -0,0 +1,19 @@
...
+# Check if the generated API is up to date
+cd frontend
+
+echo "Checking if generated API is up to date..."
+
+# Run generate-api
+npm run generate-api
+
+# Check if there are any changes in the generated files
+if ! git diff --exit-code src/store/api/services/generated/; then
```
stijndehaes: There are a a couple of issues with this.
- I don't think git diff checks for untracked filed
- When running locally it's ok that git is dirty, this is only not OK when we are in CI so this will fail locally.

A check similar to the `check_openapi_spec.sh` script should work better:

```

if [[ -n "${CI}" ]]; then
  if [[ -z "$(git status --porcelain src/store/api/services/generated)" ]];
  then
    exit 0
  else
    echo "Git is dirty"
    git status --porcelain src/store/api/services/generated
    git --no-pager diff
    exit 1
  fi
fi
```

.pre-commit-config.yaml
```diff
@@ -61,6 +61,13 @@ repos:
       language: script
       types: [text]
       files: ^backend\/.*\.py$
+    - id: check-generated-api
+      name: Check generated API is up to date
+      entry: ./scripts/check_generated_api.sh
+      language: script
+      types: [text]
+      files: ^backend\/.*\.py$
```
stijndehaes: I would let this depend on `docs/static/openapi.json` instead of backend, when the openapi changed only then we should run this one

stijndehaes (changes requested):

stijndehaes (approved):

PR #3154, opened 2026-04-14
Chore/remove v1 api 3

stijn-janssens:
Builds further on #3135


backend/app/data_products/output_ports/service.py
```diff
@@ -527,12 +527,22 @@ def is_visible_to_user(self, dataset: DatasetModel, user: User) -> bool:
...
+    def get_output_ports(
+        self, data_product_id: Optional[UUID], user: User
+    ) -> Sequence[OutputPort]:
         query = select(DatasetModel)
         if data_product_id is not None:
             ensure_data_product_exists(data_product_id, self.db)
             query = query.filter(DatasetModel.data_product_id == data_product_id)
-        return self.db.scalars(query).unique().all()
+
+        results = self.db.scalars(query).unique().all()
```
nclaeys, line 538: Can we add a test for this functionality? This way we make sure the problem does not reappear.

backend/app/data_products/service.py
```diff
@@ -110,6 +110,7 @@ def get_rolled_up_tags(self, data_product_id: UUID) -> set[Tag]:
 
         return rolled_up_tags
 
+    @deprecated("Leftover from API v1")
```
nclaeys, line 113: Why do we keep this instead of also removing it?

backend/app/data_products/output_ports/service.py
stijn-janssens: this has been tested and solved in the previous PR #3135 
Unit tests where broken before I added in this change and this changed fixed the logic. So logic is unit tested :)

backend/app/data_products/service.py
stijn-janssens: For now this is still used in the MCP code, which is for a later stage of the cleanup.
Also, as discussed with Stijn on the initial PR, we decided on mainly focussing on the router cleanup and only do the obsolete service logic as a second pass.

nclaeys (approved):

PR #3167, opened 2026-04-15
Chore/update python sdk

stijn-janssens:
Pre-commit hook to keep the Python SDK in sync with openapi spec
- removed v2 filtering as v1 is fully removed


nclaeys (approved):

scripts/check_generated_sdk.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 0
+  else
+    echo "Git is dirty"
+    git status --porcelain .
+    git --no-pager diff
+    exit 1
+  fi
+fi
+
+echo "✅ Generated API files are up to date."
```
stijndehaes: A nicer way to do this could be:

```suggestion

echo "Checking if generated SDK is up to date..."
pushd sdk


  # Install dependencies (including openapi-python-client)
  poetry install --no-interaction --only dev
  
  # Run generate-api
  poetry run openapi-python-client generate \
    --path ./../docs/static/openapi.json \
    --output-path ./sdk/api_client \
    --meta none \
    --overwrite


  if [[ -n "${CI}" ]]; then
    if [[ -z "$(git status --porcelain .)" ]];
    then
      exit 0
    else
      echo "Git is dirty"
      git status --porcelain .
      git --no-pager diff
      exit 1
    fi
  fi

popd
echo "✅ Generated API files are up to date."
```

Pushd enter the directory, and by adding tabs to all command it's clearly visual you are executing them in that context, popd undoes the pushd, which is nice if you want after that command to be executed in the normal dir.

stijndehaes (approved):

scripts/check_generated_sdk.sh
stijn-janssens: TIL!

PR #3169, opened 2026-04-15
Rework search output ports setup

nclaeys:
Remove duplication in sample_data and alembic migrations.

In the meantime fix a lot of issues in our down and up migrations.


stijn-janssens (approved):

stijndehaes (approved):
Approved but had one question, why don't these issues pop up in our tests?

Or can we add a test to figure out these issues?

nclaeys:
> Approved but had one question, why don't these issues pop up in our tests?
> 
> Or can we add a test to figure out these issues?

Because there is no test that performs all down and up migrations.
Now we have 1 that will detect all the issues.

PR #3171, opened 2026-04-15
ADR for emitting v2 events

pascal-knapen:

### DIFF
--- a/docs/adr/0014-enriched-webhook-events.md
+++ b/docs/adr/0014-enriched-webhook-events.md
@@ -29,7 +29,6 @@ We want a v2 event system that delivers named events with full domain objects em
 * **Explicit and discoverable.** Reading a route handler should make it immediately obvious whether a v2 event fires, and for which named event. A grep should return a definitive list of all emitting endpoints.
 * **Testable.** Each emitting endpoint must be testable in isolation without making real outbound webhook calls.
 * **Full domain object in payload.** The event payload must include the complete hydrated Pydantic object, not just an ID.
-* **Consistent with existing patterns.** The codebase already expresses cross-cutting concerns as FastAPI `Depends()` callables returned from factory functions (see `Authorization.enforce()` in `backend/app/core/authz/authorization.py`). The v2 mechanism should follow the same idiom.
 
 ## Considered Options
 
--- a/docs/adr/0014-enriched-webhook-events.md
+++ b/docs/adr/0014-enriched-webhook-events.md
@@ -83,7 +83,6 @@ Like all FastAPI dependencies, the event-emitting dependencies run as part of th
 * **Good, because** the webhook call is co-located with the business operation.
 * **Bad, because** services are transport-agnostic by convention: they accept a `Session` and return domain objects; they do not import HTTP concerns such as `BackgroundTasks` or `Request`.
 * **Bad, because** `BackgroundTasks` is not available inside services; emitting with `asyncio.create_task()` directly would re-introduce the fire-on-failure risk of v1.
-* **Bad, because** each service that emits events becomes harder to unit test since the emitter must be mocked at the service layer.
 
 ---
 
--- a/docs/adr/0014-enriched-webhook-events.md
+++ b/docs/adr/0014-enriched-webhook-events.md
@@ -1,0 +1,395 @@
+# Enriched Webhook Events (v2)
+
+## Context and Problem Statement
+
+The Portal backend fires a webhook after every mutating API call via an HTTP middleware (`backend/app/main.py`). The payload is a raw forward of the HTTP exchange:
+
+```json
+{
+  "method": "POST",
+  "url": "/api/v2/data_products",
+  "query": {},
+  "response": "{\"id\": \"abc-123\"}",
+  "status_code": 200
+}
+```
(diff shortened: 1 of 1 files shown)


docs/adr/0014-enriched-webhook-events.md
```diff
@@ -0,0 +1,394 @@
...
+* **Good, because** DELETE works correctly: the dependency fetches the full object before the handler body deletes it.
+* **Neutral, because** UPDATE and CREATE routes require a yield-based factory (`emit_event_after`) rather than the simpler `emit_event`. CREATE routes additionally require one line in the handler body to store the created object in `request.state`.
+* **Neutral, because** each emitting route requires one additional entry in its `dependencies=[]` list.
+
+### Option C: Service-Layer Emission
+
+* **Good, because** the webhook call is co-located with the business operation.
+* **Bad, because** services are transport-agnostic by convention: they accept a `Session` and return domain objects; they do not import HTTP concerns such as `BackgroundTasks` or `Request`.
+* **Bad, because** `BackgroundTasks` is not available inside services; emitting with `asyncio.create_task()` directly would re-introduce the fire-on-failure risk of v1.
+* **Bad, because** each service that emits events becomes harder to unit test since the emitter must be mocked at the service layer.
```
stijndehaes: Disagree that this is bad, the good is that you must mock it!
And that you have to say, I expected an event to be fired as a developer. Instead of it being a black box.

docs/adr/0014-enriched-webhook-events.md
```diff
@@ -0,0 +1,394 @@
...
+* **Good, because** annotated routes are self-documenting: the `dependencies=[]` list shows both the authorization requirement and the event contract in one place.
+* **Good, because** a project-wide search for `emit_event` returns the complete catalogue of emitting endpoints.
+* **Good, because** it is testable by patching `call_v2_webhook` or using `app.dependency_overrides` — the same mechanisms used for existing v1 webhook tests and auth overrides in `conftest.py`.
+* **Good, because** DELETE works correctly: the dependency fetches the full object before the handler body deletes it.
+* **Neutral, because** UPDATE and CREATE routes require a yield-based factory (`emit_event_after`) rather than the simpler `emit_event`. CREATE routes additionally require one line in the handler body to store the created object in `request.state`.
+* **Neutral, because** each emitting route requires one additional entry in its `dependencies=[]` list.
+
+### Option C: Service-Layer Emission
```
stijndehaes, line 81: This is the option chosen for audit/history events, so choosing this would be in line with that decision.

So the downside to option B is that you choose a different architecture pattern

docs/adr/0014-enriched-webhook-events.md
```diff
@@ -0,0 +1,394 @@
...
+* **Bad, because** the middleware does not have access to a DB session without injecting one separately; enriching the payload requires a round-trip DB query after the transaction has already committed.
+* **Bad, because** the emitting logic is untestable at the individual route level; tests must manipulate the middleware globally.
+
+### Option B: Dependency Annotation on Routes
+
+* **Good, because** it matches the existing `Authorization.enforce()` pattern exactly.
+* **Good, because** success-only semantics are mechanically guaranteed without any explicit status-code check in application code. For DELETE, Starlette discards background tasks when the handler raises `HTTPException`. For UPDATE and CREATE, the yield dependency's `try/except/else` skips the webhook call on any exception.
+* **Good, because** annotated routes are self-documenting: the `dependencies=[]` list shows both the authorization requirement and the event contract in one place.
+* **Good, because** a project-wide search for `emit_event` returns the complete catalogue of emitting endpoints.
```
stijndehaes, line 75: Can we even scrape that programmaticaly to generate docs out of it?

stijndehaes: I would like to have docs that say for which endpoints which events will be send with which object.

docs/adr/0014-enriched-webhook-events.md
stijn-janssens: Not entirely correct, the audit / history events are currently being created and emitted in the routers. So not tied to commits in the database etc.

Not saying that is the correct architecture. But it does not fit 100% with what is being proposed in option C

docs/adr/0014-enriched-webhook-events.md
pascal-knapen: Yes, we probably can

docs/adr/0014-enriched-webhook-events.md
pascal-knapen: So, we should have a ticket to revisit auditing later on and make things more consistent? Will have a look at auditing and how this has been added.

docs/adr/0014-enriched-webhook-events.md
pascal-knapen: True, will remove this as an argument

docs/adr/0014-enriched-webhook-events.md
stijn-janssens: I would indeed revise the auditing. I like this system more. But if possible all should use the same approach

stijndehaes (approved):

PR #3183, opened 2026-04-16
ADR 15 lightweight data products

stijndehaes:

### DIFF
--- a/docs/adr/0015-light-weight-data-products.md
+++ b/docs/adr/0015-light-weight-data-products.md
@@ -1,0 +1,83 @@
+# Lightweight data products technical design
+
+## Context and Problem Statement
+We have decided to create a lightweight version of a data product meant for experimentation purposes and to fix
+personal access.
+The idea is to not have people create a heavy data product, and to make it easy for us to differentiate.
+For example for a full data product we will want to ensure people share their data, have data quality checks, data
+contracts etc...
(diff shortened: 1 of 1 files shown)


stijn-janssens (approved):
I like it!

stijndehaes:
I talked to @pascal-knapen he was a bit worried with the joined approach that the joins would slow down the application, but we currently already do about 15 joins when fetching data products. And it isn't an issue, so it should not be know. The union variant forced by having fully separate tables would have a performance impact. But joined version should be fine

stijndehaes:
After trying it out myself to get a feel I am wondering if this makes sense,  data_product and LWDP share only a little:
- name
- domain
- description

And that is about it, to add people for an LWDP for example we need to make custom roles for this. Because we can't reuse the data product roles, they will be different, so assignment will work differently.

The can possibly share the logic for input ports, so that might make it worth it.

Another reason to use some inheritance in the database is to be able to show them in a single list somewhere.
I will try to verify if this is behavior that we want. 

So for now I'll implement with database inheritance it is rather trivial to split it up again later.


I also noticed we now have a uniqueness constraint `uq_data_product` on (namespace, name), I am unsure how important that is. Because this means you are allow to have the same namespace, as long as the name is differnet or same name, as long as the namespace is different. I don't think that makes sense

PR #3185, opened 2026-04-16
Rework mcp

nclaeys:

### DIFF
--- a/backend/app/mcp/mcp.py
+++ b/backend/app/mcp/mcp.py
@@ -78,7 +78,22 @@ def get_auth_provider() -> Optional[JWTVerifier]:
     return None
 
 
-mcp = FastMCP(name="DataProductPortalMCP", auth=get_auth_provider())
+mcp = FastMCP(
+    name="DataProductPortalMCP",
+    instructions="""
+    Portal for discovering and exploring data products and output ports.
+
+    Recommended discovery flow:
+    1. get_marketplace_overview: understand what's available and get domain IDs
+    2. search_output_ports: find specific output ports (preferred for most searches)
+       Use search_data_products only when the user explicitly asks to find a data product.
+    3. get_*_details: drill into a specific result by UUID
+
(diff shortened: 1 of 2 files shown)


backend/app/mcp/mcp.py
```diff
@@ -105,12 +127,15 @@ def universal_search(
     query: str, entity_types: list[str] = [], limit: int = 10
 ) -> Dict[str, Any]:
     """
-    Universal search across data products, datasets, and data outputs.
+    Search across data products, output ports, technical assets, and domains in a single call.
```
stijndehaes, line 130: Just for my understanding, is this documentation returned via the mcp protocol somehow?

mcp/README.md
```diff
@@ -0,0 +1,7 @@
+# Local development of MCP server
+
+When testing the MCP server locally, keep in mind the following:
+
+- You need to have OIDC authentication set up and running
```
stijndehaes, line 5: I assume it's possible in the future to ensure we don't need that?

stijndehaes (approved):

stijn-janssens (approved):
No rework of the proxy? Assuming that the old way works on most machines? :D

backend/app/mcp/mcp.py
stijn-janssens: I believe it is used in the selection of the correct tool to use yes. I can view this in my available tools in Cursor

nclaeys: Yes it is shown when a client lists the tools of an mcp server.

mcp/README.md
nclaeys: well yes we can. We can find a way but the MCP does a token check so we need to find to fake this for testing.

nclaeys:
> No rework of the proxy? Assuming that the old way works on most machines? :D

Not for now. We can but then we need to publish our dpp-sdk first. If I am the only one struggling it does not make sense. I have a workaround now.

PR #3199, opened 2026-04-17
feature/v2 events

pascal-knapen:


backend/app/authorization/role_assignments/data_product/router.py
```diff
@@ -30,13 +30,35 @@
...
+from app.data_products.service import DataProductService
 from app.database.database import get_db_session
 from app.events.enums import EventReferenceEntity, EventType
 from app.events.schema import CreateEvent
 from app.events.service import EventService
 from app.users.notifications.service import NotificationService
 from app.users.schema import User
 
+_emit_data_product_team_member_added = emit_event_after(
+    "data_product.team_member_added",
```
stijn-janssens, line 44: I miss an event at update time. Worst case you can handle it with remove and create (which still both need to be send) but I would prefer an update.

backend/app/authorization/role_assignments/data_product/router.py
```diff
@@ -157,22 +180,25 @@ def request_data_product_role_assignment(
...
     role_assignment = service.create_assignment(
-        data_product_id=request.data_product_id,
-        user_id=request.user_id,
-        role_id=request.role_id,
+        data_product_id=body.data_product_id,
+        user_id=body.user_id,
+        role_id=body.role_id,
         actor=user,
     )
+    request.state.data_product_id = role_assignment.data_product_id
```
stijn-janssens, line 211: I am not really following. Why does this need to be passed via the request and can't this be passed directly as an argument as you do with the id?

docs/docs/developer-guide/webhook-events.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+### Data Product Events
+
+| Event | HTTP Trigger | Payload |
+|-------|-------------|---------|
+| `data_product.about_updated` | `PUT /v2/data_products/{id}/about` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
+| `data_product.created` | `POST /v2/data_products` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
+| `data_product.deleted` | `DELETE /v2/data_products/{id}` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
+| `data_product.input_port_linked` | `POST /v2/data_products/{id}/link_input_ports` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
```
stijn-janssens, line 42: shouldn't we have input port rejected events as well? I think either we need to add those here or remove the denied and accepted events at TA level. I don't really know of any useful integrations at that point but might be missing some.

docs/docs/developer-guide/webhook-events.md
```diff
@@ -0,0 +1,73 @@
...
+| `data_product.input_port_unlinked` | `DELETE /v2/data_products/{id}/input_ports/{input_port_id}` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
+| `data_product.setting_changed` | `POST /v2/data_products/{id}/settings/{setting_id}` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
+| `data_product.status_updated` | `PUT /v2/data_products/{id}/status` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
+| `data_product.team_member_added` | `POST /v2/authz/role_assignments/data_product` | [`data_product`](/docs/api/#schema/GetDataProductResponse) |
```
stijn-janssens, line 46: as said above, team_member_updated should definitely be here

stijn-janssens (commented):
In general I like the idea a lot!
As discussed: downstream updates are the responsability of the provisioner then? Events will only be send for the current impacted object

backend/app/authorization/role_assignments/data_product/router.py
pascal-knapen: You mean a change is role for the role of a user on a data product?

backend/app/authorization/role_assignments/data_product/router.py
pascal-knapen: You are right. Following the default pattern here. It works but isn't really needed as I could get it from the request in the first place. Let me see if I did this in multiple place.

backend/app/authorization/role_assignments/data_product/router.py
pascal-knapen: Done!

docs/docs/developer-guide/webhook-events.md
pascal-knapen: Done

backend/app/authorization/role_assignments/data_product/router.py
pascal-knapen: Nevermind. Not sure what you means as the id ins't a query param in this case but part of the body. I can access it, but it's more of a hassle to get it. I think this is a cleaner solution.

pascal-knapen:
> In general I like the idea a lot! As discussed: downstream updates are the responsability of the provisioner then? Events will only be send for the current impacted object

Yes, at least in this first version. In the next pull request, I will update the 2 demos and I will probably learn a lot more.

backend/app/authorization/role_assignments/data_product/router.py
```diff
@@ -30,13 +30,45 @@
...
 from app.users.schema import User
 
+_emit_data_product_team_member_added = emit_event_after(
+    "data_product.team_member_added",
+    lambda request, db, **_: {
+        "data_product": GetDataProductResponse.model_validate(
+            DataProductService(db).get_data_product(request.state.data_product_id)
+        )
+    },
+)
```
stijndehaes, line 50: I would have expected this to be a pydantic model itself, instead of a dictionary with a pydantic model inside.

The event type would then just be the pydantic modal name, something like:


```suggestion
_emit_data_product_team_member_added = emit_event_after(
    DataProductTeamMemberAddedEvent
    lambda request, db, **_: DataProductTeamMemberAddedEvent(
         data_product=GetDataProductResponse.model_validate(
            DataProductService(db).get_data_product(request.state.data_product_id)
        )
    )
)
```

I think this might make life easier later in the SDK itself as well.

I am also unsure if adding the full data product is a good idea, I would have expected specialised models here which contain less info. For example I don't think the data product about should be included here

backend/app/authorization/role_assignments/data_product/router.py
stijndehaes: After reading what request.state is it makes sense to use to share things between functions.

But at this point I am wondering, wouldn't it be better if we just set the event in the state in this function by doing that?
And just have a depends on that emits the event?

So something like this:


```
@router.post(
    "",
    dependencies=[
        Depends(
            Authorization.enforce(
                Action.DATA_PRODUCT__CREATE_USER,
                resolver=DataProductResolver,
                object_id="data_product_id",
            )
        ),
        Depends(_emits_event(CreateDataProductAddMemberEvent)),
    ],
)
def create_data_product_role_assignment(
    body: CreateDataProductRoleAssignment,
    request: Request,
    background_tasks: BackgroundTasks,
    user: User = Depends(get_authenticated_user),
) -> DataProductRoleAssignmentResponse:
  
    request.state.event = CreateDataProductAddMemberEvent(data_product=...)
```

That way you emits_event has to do almost nothing and we don't have to do extra round trips to the DB when not needed, but it's very clear for us as developers we have to send and event to the state.

Feel free to call me to discuss this

backend/app/core/webhooks/v2.py
```diff
@@ -0,0 +1,103 @@
...
+        "specversion": "1.0",
+        "id": str(uuid.uuid4()),
+        "source": "data-product-portal",
+        "type": event_type,
+        "time": datetime.now(timezone.utc).isoformat(),
+        "data": data,
+    }
+    try:
+        async with httpx.AsyncClient() as client:
+            resp = await client.post(url, json=body)
```
stijndehaes, line 30: Do we need a timeout here? I would say 5 seconds?

backend/app/core/webhooks/v2.py
```diff
@@ -0,0 +1,103 @@
...
+            if resp.status_code != 200:
+                logger.warning("v2 webhook returned %d", resp.status_code)
+    except Exception as e:
+        logger.warning("v2 webhook failed: %s", e)
+
+
+def emit_event(event_type: str, extract) -> Callable:
```
stijndehaes, line 37:
```suggestion
def emit_event_delete(event_type: str, extract) -> Callable:
```

backend/app/core/webhooks/v2.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+            data = {k: v.model_dump(mode="json") for k, v in result.items()}
+            background_tasks.add_task(call_v2_webhook, event_type, data)
+        except Exception as e:
+            logger.warning("v2 webhook setup failed: %s", e)
+
+    return _dependency
+
+
+def emit_event_after(event_type: str, extract) -> Callable:
```
stijndehaes, line 70:
```suggestion
def emit_event_create_update(event_type: str, extract) -> Callable:
```

backend/app/core/webhooks/v2.py
stijndehaes: Proposed rename to make the method clearer

backend/app/data_products/output_ports/service.py
```diff
@@ -123,6 +121,40 @@ def get_dataset(
...
+        For system/internal callers already authorised at the endpoint level, use
+        get_dataset() instead.
+        """
+        dataset = self._load_dataset(id, data_product_id)
+        if not self.is_visible_to_user(dataset, user):
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail="You do not have access to this private dataset",
+            )
+        return dataset
```
stijndehaes, line 149: I would rewrite this into 1 method, `get_dataset` with a parameter check_visibility.

Or if you want 2 functions, you can just make two functions:
`get_dataset` without visibility check, so just fold `_load_dataset` into this one.
And `get_visible_dataset`, 3 functions are not needed.

backend/app/data_products/output_ports/service.py
```diff
@@ -123,6 +121,40 @@ def get_dataset(
...
+    ) -> DatasetGet:
+        """Fetch a dataset without enforcing consumer-visibility.
+
+        Use this for internal/system callers (e.g. webhook extract lambdas) where the
+        endpoint's Authorization.enforce() dependency has already authorised the request.
+        The visibility gate ("can this user consume this dataset?") is irrelevant once
+        endpoint-level permission is established.
+
+        For user-facing endpoints and MCP tools, use get_visible_dataset() instead.
+        """
```
stijndehaes, line 130:
```suggestion
        """Fetch a dataset without enforcing consumer-visibility.

        Use this for internal/system callers thave have already authorised the request.
        """
```

backend/app/data_products/output_ports/service.py
```diff
@@ -123,6 +121,40 @@ def get_dataset(
...
+    ) -> DatasetGet:
+        """Fetch a dataset, raising 403 if the user cannot see it as a consumer.
+
+        Use this for user-facing endpoints and MCP tools where a specific user identity
+        is available and dataset visibility must be enforced. Non-UNRESTRICTED datasets
+        are hidden from users who are not in the dataset's access list.
+
+        For system/internal callers already authorised at the endpoint level, use
+        get_dataset() instead.
+        """
```
stijndehaes, line 142:
```suggestion
        """Fetch a dataset, raising 403 if the user cannot see it as a consumer.

        Use this for endpoints where dataset visibility must be enforced.

        For system/internal callers already authorised at the endpoint level, use
        get_dataset() instead.
        """
```

backend/tests/app/core/webhooks/test_v2_data_products.py
stijndehaes, line 1: Why not add these to data_products router? A lot of this is copy past from there too.

We can just add checks to existing tests instead of writing complete new tests no?

backend/tests/app/core/webhooks/test_v2_output_ports.py
stijndehaes, line 1: Again can't we add extra checks to existing output port tests?

backend/tests/conftest.py
```diff
@@ -115,8 +116,9 @@ def clear_db(session: scoped_session[Session]) -> None:
 
 
 @pytest.fixture
-def admin() -> UserFactory:
+def admin(session) -> UserFactory:
```
stijndehaes, line 119: Whe these changes? I don't understand it :)

scripts/generate_webhook_docs.py
```diff
@@ -0,0 +1,250 @@
...
+  "data": { ... }
+}
+```
+
+## Events
+
+"""
+
+
+def find_router_files() -> list[Path]:
```
stijndehaes, line 62: This can be simplified quite a lot, if you commit this script to the backend folder, you can just ask the fast api app for all the routes. This is a lot less error prone. 
You can then ask for all their dependencies and see if emit events is part of that. Hopefully when emit_events says it emits a certain class, you can even extract that.

stijndehaes (changes requested):

backend/app/authorization/role_assignments/data_product/router.py
pascal-knapen: Let's discuss. I don't know yet either. For now I return the entities as they exist today in the APIs (as we discussed in the orginal ADR). I did do a quick check for the demo I often need to know some details of the underlying data product. I propose for this pull request to leave it as is and that I add an experimental flag on the documentation. When I then implement the actual provisioner for the demo's I can finetune what I am returning. Wdyt?

backend/app/core/webhooks/v2.py
pascal-knapen: Good point!

backend/app/authorization/role_assignments/data_product/router.py
pascal-knapen: Indeed, let'a have a short call to decide.

scripts/generate_webhook_docs.py
pascal-knapen: I'll have a look and see how easy it is.

pascal-knapen:
- Test using the state to put the entire event
- Maximum events but will refactor after implementing provisioner (alpha feature flag met notice on docs)
- Regeneration 
- Wrapper objects (pydantic)

backend/app/data_products/router.py
```diff
@@ -102,26 +124,32 @@ def create_data_product(
...
         extra_receiver_ids=owners,
     )
     RefreshInfrastructureLambda().trigger()
     if data_product.input_ports is not None:
         request_input_ports_for_data_product(
-            created_data_product.id,
+            created_id,
             data_product.input_ports,
             background_tasks=background_tasks,
+            http_request=request,
```
pascal-knapen, line 143: Unclear why I added this?

backend/app/core/webhooks/events.py
```diff
@@ -0,0 +1,276 @@
...
+
+    Every subclass must define an ``event_type()`` classmethod that returns
+    the CloudEvents ``type`` string.  Omitting it raises ``TypeError`` at
+    class-definition time so mistakes are caught on import.
+    """
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        if "event_type" not in cls.__dict__:
+            raise TypeError(f"{cls.__name__} must define an event_type() classmethod")
```
stijndehaes, line 28: I love this, we can use it to have a dict of all events.
I would also use this to ensure every event type type is unique.

```suggestion
    events = {}
    def __init_subclass__(cls, **kwargs: Any) -> None:
        super().__init_subclass__(**kwargs)
        if "event_type" not in cls.__dict__:
            raise TypeError(f"{cls.__name__} must define an event_type() classmethod")
        if cls.event_type() in cls.events:
            raise Exception("Duplicated event type detected")
        cls.event_type[cls.event_type()] = cls
```

We can later use these registered events in the webhook implementation

backend/app/core/webhooks/v2.py
stijndehaes: I still don't see a timeout here?

backend/app/data_products/service.py
```diff
@@ -93,10 +93,10 @@ def get_rolled_up_tags(self, data_product_id: UUID) -> set[Tag]:
         return rolled_up_tags
 
     def get_data_product(self, id: UUID) -> GetDataProductResponse:
-        data_product = self.db.get(
-            DataProductModel,
-            id,
-            options=[selectinload(DataProductModel.tags)],
+        data_product = self.db.scalar(
```
stijndehaes, line 98: Why is this changed?

backend/app/data_products/output_ports/input_ports/router.py
```diff
@@ -59,21 +69,31 @@ def get_input_ports_for_output_port(
...
 ) -> None:
     data_product_link = DataProductDatasetService(db).approve_output_port_as_input_port(
         data_product_id=data_product_id,
         output_port_id=output_port_id,
-        consuming_data_product_id=request.consuming_data_product_id,
+        consuming_data_product_id=body.consuming_data_product_id,
         actor=authenticated_user,
     )
+    http_request.state.event = OutputPortLinkApprovedEvent(
+        data_product=GetDataProductResponse.model_validate(
```
stijndehaes: This is incorrect, this is not a data product, but an abstract data product. This will fail on explorations

backend/app/data_products/output_ports/input_ports/router.py
```diff
@@ -103,21 +123,31 @@ def approve_output_port_as_input_port(
...
 ) -> None:
     data_product_link = DataProductDatasetService(db).deny_output_port_as_input_port(
         data_product_id=data_product_id,
         output_port_id=output_port_id,
-        consuming_data_product_id=request.consuming_data_product_id,
+        consuming_data_product_id=body.consuming_data_product_id,
         actor=authenticated_user,
     )
+    http_request.state.event = OutputPortLinkDeniedEvent(
+        data_product=GetDataProductResponse.model_validate(
```
stijndehaes: This should be an abstract data product

backend/tests/app/authorization/role_assignments/data_product/test_router.py
```diff
@@ -582,3 +583,95 @@ def modify_data_product_role_assignment(client: TestClient, assignment_id, json)
...
+                    "user_id": str(member.id),
+                    "role_id": str(role.id),
+                },
+            )
+
+        assert response.status_code == 200
+        mock_webhook.assert_awaited_once()
+        event_type, data = mock_webhook.call_args.args
+        assert event_type == "data_product.team_member_added"
+        assert "data_product" in data
```
stijndehaes, line 609: I think we can do this better, and just extend existing tests by creating something like (AI generated):
```
import functools
from unittest.mock import patch, AsyncMock
from contextlib import ExitStack

def assert_webhook_fired(expected_event):
    def decorator(func):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            # We use patch as a context manager inside the wrapper
            with patch("app.core.webhooks.v2.call_v2_webhook") as mock_webhook, \
                 webhook_v2_config():
                
                mock_webhook.return_value = AsyncMock()
                
                # We inject the mock into kwargs in case the test needs 
                # to do extra specific assertions (optional)
                kwargs['mock_webhook'] = mock_webhook
                
                # Execute the actual test
                result = func(*args, **kwargs)

                # Automated Assertions
                mock_webhook.assert_awaited_once()
                event_type, data = mock_webhook.call_args.args
                
                assert event_type == expected_event, f"Expected {expected_event}, got {event_type}"
                
                return result
        return wrapper
    return decorator
```

And using that in existing tests:
```
@pytest.mark.usefixtures("admin")
@assert_webhook_fired("data_product.team_member_added")
def test_team_member_added_fires_event(self, client, mock_webhook):
    # Setup
    dp = DataProductFactory()
    member = UserFactory()
    role = RoleFactory(scope=Scope.DATA_PRODUCT, permissions=[])

    # Action
    response = client.post(
        ENDPOINT,
        json={
            "data_product_id": str(dp.id),
            "user_id": str(member.id),
            "role_id": str(role.id),
        },
    )

    # Verification (The decorator handles the webhook check!)
    assert response.status_code == 200
```

WDYT?

backend/tests/app/data_products/output_ports/test_router.py
```diff
@@ -855,3 +858,213 @@ def get_output_port_history(client, output_port_id, data_product_id):
         return client.get(
             f"{ENDPOINT.format(data_product_id)}/{output_port_id}/history"
         )
+
+
+@pytest.fixture
+def output_port_event_payload():
```
stijndehaes: I think this needs to be indented

stijndehaes (changes requested):
I think this already looks great!
Some things we have to watch out for:
- On input ports we should export abstract data products, not data products
- I would rather enhance existing tests then create a bunch of new ones. If we make new objects I think it makes sence you only make one creation test

backend/app/data_products/service.py
pascal-knapen: Added comment

stijndehaes (approved):

backend/tests/app/authorization/role_assignments/data_product/test_router.py
stijndehaes: Won't do for now

PR #3210, opened 2026-04-20
Initial commit for versioning ADR

stijn-janssens:

### DIFF
--- a/docs/adr/0015-versioning.md
+++ b/docs/adr/0015-versioning.md
@@ -1,0 +1,256 @@
+# Output Port Versioning
+
+## Context and Problem Statement
+
+Producers need to evolve their data products over time. When a Producer changes a table schema or data transformation logic, Consumers currently have no control over when they adopt the change. The Portal has no built-in versioning, so schema changes are immediately visible to all Consumers, potentially breaking their pipelines.
+
+The current workaround is to manually create new technical assets and output ports for the versions, which would show duplicates in the marketplace and create confusion all around.
+
+**Core question: Should we version Output Ports, and if so, what does a "version" represent?**
+
+This ADR focuses narrowly on the versioning mechanism itself. Authorization, notifications, and deprecation workflows are out of scope and are subject to further ADRs.
+
+## Decision Drivers
+
+* Practical feasibility: Solution must be operationally realistic for Producers to maintain
+* Clear versioning trigger: Must be obvious when to create a new version vs update in place
+* Consumer control: Consumers need to choose when to migrate, not be forced
+* Infrastructure alignment: Should work with how databases and storage actually evolve
+* Minimal cognitive overhead: Producers shouldn't need to understand complex versioning rules
+
+## Considered Options
+
+* **Option 1: No versioning** - Producers update Technical Assets in place; Consumers always see latest
+* **Option 2: Branch-based versioning** - Create new Output Port + new Technical Asset only for major versions (breaking changes)
+* **Option 3: Full versioning** - Every change (major/minor/patch) creates new Output Port + new Technical Asset
+
+## Decision Outcome
+
+**Chosen option: Option 2 — Branch-based versioning.**
+
+Version only when necessary: when a change would break existing Consumers. This means major version bumps only (v1 → v2 → v3).
+
+**Rationale:**
+
+The fundamental insight is that **data products evolve differently than software.** In software, you can have many minor versions (v1.0, v1.1, v1.2) because each represents a stable release at a point in time. In data products, there is no "release" - data flows continuously.
+
+Trying to version every schema change creates an unsustainable operational burden. A Producer would need to maintain views for v1.0, v1.1, v1.2, v1.3, etc., all reading from the same evolving table. This doesn't match how databases actually work in practice.
+
+Instead, think of versions as "branches":
+- **v1**: The current stable version, continuously evolving with backward-compatible changes
+- **v2**: A new branch created when a breaking change is necessary
+- Old versions (v1) remain frozen or continue with backported fixes only
+
+This aligns with how APIs are versioned in practice (`/api/v1/`, `/api/v2/`) and how databases are actually managed.
+
+### How it works
+
+**Backward-compatible changes** (no new version):
+- Producer updates the Technical Asset in place
+- Examples: Adding nullable columns, adding new API endpoints, new S3 files
+- All Consumers on v1 automatically see the changes
+- Document in changelog, but no migration required
+
+**Breaking changes** (major version bump):
+- Producer creates new Output Port (v2) with new Technical Asset
+- Examples: Removing columns, renaming columns, changing data types, schema restructuring
+- Consumers on v1 continue using old infrastructure
+- Consumers explicitly request access to v2 when ready to migrate
+- Both v1 and v2 run in parallel during migration period
+
+**When to version:** Only when the change would break existing Consumer queries or code.
+
+**Key principle:** Versions are long-lived branches, not frequent releases. A data product might have only 2-3 versions over its lifetime (v1, v2, maybe v3).
+
+**Important:** Each Output Port version MUST have its own Technical Asset. You cannot have v1 and v2 pointing to the same table/bucket/API - if they did, they would return identical data and versioning would be meaningless. This means versioning always creates infrastructure duplication.
+
+### Confirmation
+
+- Output Ports have an optional `version` field (default: v1)
+- Version is a simple integer: v1, v2, v3 (not semantic versioning)
+- Creating a new version creates a new Output Port record + new Technical Asset
+- Both versions coexist until the old version is explicitly deprecated
+- Handshakes reference a specific Output Port version
+- Backward-compatible changes are made in-place to existing versions
+- Changelog tracks all changes (versioned or not)
+
+## Pros and Cons of the Options
+
+### Option 1: No versioning
+
+* **Good, because** simplest possible solution - no new infrastructure needed
+* **Good, because** matches how most internal databases work today
+* **Good, because** no operational overhead of maintaining multiple versions
+* **Bad, because** Producers can't make breaking changes without breaking Consumers
+* **Bad, because** forces workaround of creating separate Output ports and technical assets for v1/v2
+* **Bad, because** versioning is not managed by portal and multiple versions show up as duplicates in the marketplace
+
+### Option 2: Branch-based versioning (major versions only)
+
+* **Good, because** operationally realistic - only version when necessary
+* **Good, because** matches how APIs are versioned in practice (v1, v2, v3)
+* **Good, because** clear rule: version on breaking change, update in-place otherwise
+* **Good, because** aligns with how databases evolve (backward-compatible schema additions are normal)
+* **Neutral, because** requires Producers to understand what "breaking" means for their Consumers
+* **Bad, because** backward-compatible changes happen without Consumer opt-in
+
+
+### Option 3: Full versioning (all changes create new versions)
+
+* **Good, because** maximum Consumer control - every change is opt-in
+* **Good, because** clean separation - each version is completely independent
+* **Bad, because** operationally unsustainable - would require maintaining dozens of views/endpoints
+* **Bad, because** doesn't match how databases actually evolve
+* **Bad, because** creates massive infrastructure overhead
+* **Bad, because** unclear when to version (is adding an index a new version?)
+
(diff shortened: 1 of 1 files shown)


pascal-knapen (commented):
Good morning Stijn, did a quick review. I am not sure I am completely aligned tbh. Why not do semantic versioning? Data pipelines do have a release cycle like normal software and non-breaking changes could lead to the output port beging bumped to a new minor version. This would allow to keep track of changes. Ideally the user makes it part of their CI/CD process upon release. I do agree that you don't want to update on every RUN of your data pipeline, similar that you don't want to do that for every API call you receive.

I  am considering also the case of O&V and Luminus, where a technical asset is the location where data is being stored and at what level access is granted. In those cases, the technical asset is a schema. When using dbt, I can version models but they would end up in the same schema with a suffix `_v2`. How would you deal with that? Do I then need to created x tables instead of just 1 schema. Might be quite some non-productive click work especially when my data product has like 10 tables. Would be good to describe this case or have a discussion how to solve this.

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+The fundamental insight is that **data products evolve differently than software.** In software, you can have many minor versions (v1.0, v1.1, v1.2) because each represents a stable release at a point in time. In data products, there is no "release" - data flows continuously.
+
+Trying to version every schema change creates an unsustainable operational burden. A Producer would need to maintain views for v1.0, v1.1, v1.2, v1.3, etc., all reading from the same evolving table. This doesn't match how databases actually work in practice.
+
+Instead, think of versions as "branches":
+- **v1**: The current stable version, continuously evolving with backward-compatible changes
+- **v2**: A new branch created when a breaking change is necessary
+- Old versions (v1) remain frozen or continue with backported fixes only
+
+This aligns with how APIs are versioned in practice (`/api/v1/`, `/api/v2/`) and how databases are actually managed.
```
stijndehaes: I understand what you are trying to say, only having a change for a major version has impact and thus needs to create new technical assets etc, and fundamentally changes your output port.

But that doesn't mean there is a value in having a changelog for your v1, just because only the latest version of v1 available doesn't mean you don't want to tell people a new field is available. Or that certain fields are deprecated in favor of something else.

There is value in that.

Now I do agree the MVP is just supporting major versions

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+
+### How it works
+
+**Backward-compatible changes** (no new version):
+- Producer updates the Technical Asset in place
+- Examples: Adding nullable columns, adding new API endpoints, new S3 files
+- All Consumers on v1 automatically see the changes
+- Document in changelog, but no migration required
```
stijndehaes: I notice you add a changelog here, why not just bump the version number then :) From v1 to v1.1

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+**When to version:** Only when the change would break existing Consumer queries or code.
+
+**Key principle:** Versions are long-lived branches, not frequent releases. A data product might have only 2-3 versions over its lifetime (v1, v2, maybe v3).
+
+**Important:** Each Output Port version MUST have its own Technical Asset. You cannot have v1 and v2 pointing to the same table/bucket/API - if they did, they would return identical data and versioning would be meaningless. This means versioning always creates infrastructure duplication.
+
+### Confirmation
+
+- Output Ports have an optional `version` field (default: v1)
+- Version is a simple integer: v1, v2, v3 (not semantic versioning)
```
stijndehaes: I disagree with this, there is still value in having semantic versioning, however only a major version change results in need to create a new technical asset.
Agree that this is an MVP though

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+* **Good, because** matches how most internal databases work today
+* **Good, because** no operational overhead of maintaining multiple versions
+* **Bad, because** Producers can't make breaking changes without breaking Consumers
+* **Bad, because** forces workaround of creating separate Output ports and technical assets for v1/v2
+* **Bad, because** versioning is not managed by portal and multiple versions show up as duplicates in the marketplace
+
+### Option 2: Branch-based versioning (major versions only)
+
+* **Good, because** operationally realistic - only version when necessary
+* **Good, because** matches how APIs are versioned in practice (v1, v2, v3)
```
stijndehaes: API's still have minor and bugfix versions :) it's just that the v1 version always uses the latest v1, and you can't go back to a previous minor of v1

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+- Removing a column
+- Renaming a column
+- Changing a column's data type
+- Removing an API endpoint
+- Changing API response structure
+- Restructuring file formats
+- Removing required fields
+
+**Not breaking (update in-place):**
+- Adding a new nullable column
```
stijndehaes:
```suggestion
- Adding a new column
```

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+- Adding a new API endpoint
+- Adding optional fields to JSON
```
stijndehaes:
```suggestion
- Adding fields to JSON
```

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+- Adding new files to S3 prefix
+- Performance improvements
+- Data quality fixes
+- Documentation updates
+
+**Gray area (Producer decides):**
+- Adding a new non-nullable column (breaks `SELECT *` but not explicit column lists)
```
stijndehaes: This is non breakable no?

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+- Changing column order (breaks `SELECT *`)
```
stijndehaes: This is non breaking

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+- Changing data freshness (e.g., real-time → daily batch)
```
stijndehaes: This is breaking

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+**Change needed:**
+- Add `loyalty_tier` column
+
+**Steps:**
+1. Add column to existing table: `ALTER TABLE customers ADD COLUMN loyalty_tier VARCHAR(20)`
+2. Update Output Port metadata (schema documentation)
+3. No new version needed - all Consumers on v1 automatically have access to new column
+4. Consumers can choose to use new column or ignore it
+
+### Database schema
```
stijndehaes: Is this also an example?

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+
+-- Add unique constraint: can't have duplicate versions for same logical port
+CREATE UNIQUE INDEX idx_datasets_version
+ON datasets(name, data_product_id, version);
+```
+
+No version states (active/deprecated/sunset) needed - just delete old versions when fully migrated.
+
+
+## Implications and Trade-offs
```
stijndehaes: I think we can remove thsi section, it's a bit more of the same

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+---
+
+## Open Questions
+
+1. **What if Consumers disagree on whether a change is breaking?**
+   - A Producer thinks adding a column is non-breaking (it's nullable)
+   - But a Consumer using `SELECT *` will get broken code
+   - Decision: Producer determines breaking-ness, but should be conservative
+   - Recommendation: Producers should avoid `SELECT *` in their documentation
+
```
stijndehaes: Does this break? It won't in spark or pandas for example.

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+2. **How long should old versions be supported?**
+   - No technical enforcement - purely a governance/SLA question
+   - Recommendation: Set deprecation timeline per Data Product (e.g., 6 months notice)
+   - Portal can track which Consumers are on which version to inform deprecation
```
stijndehaes: I think we should support setting deprecation timeline when making a major version bump

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+
+3. **What about data contracts?**
+   - If data contracts are adopted, they could help identify breaking changes
+   - Contract diff tool could flag schema changes as breaking/non-breaking
+   - This is complementary to versioning, not a replacement
+   - See: ADR-0012 (BiToL standard supports data contracts)
```
stijndehaes: Each major version has a differnt contract no?

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+
+4. **Should there be a "latest" alias?**
```
stijndehaes, line 188: Nope I would encourage to not do this, also we are not a technical catalog so what does it even mean?

When people consume an output port they consume every version anyway no?

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,256 @@
...
+   - Deferred: Start without aliases
+   - Forces Consumers to make explicit version choices
+   - Can add later if there's demand for "always use latest stable"
+
+5. **How does this interact with data quality issues?**
```
stijndehaes, line 191: I guess data quality should be posted per major version?

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,259 @@
...
+
+**Patch changes** (1.0.0 → 1.0.1):
+- Data quality fixes, documentation updates, performance improvements
+- Update in place, bump version number on existing Output Port
+- Notify Consumers via changelog
+
+**Minor changes** (1.0.0 → 1.1.0):
+- Backward-compatible additions (new column, new file, new API field)
+- Update in place, bump version number on existing Output Port
+- Notify Consumers via changelog
```
stijndehaes, line 61:
```suggestion
**Minor changes** (1.0.0 → 1.1.0):
- Backward-compatible additions (new column, new file, new API field)
- Update in place, bump version number on existing Output Port
- Notify Consumers via changelog
- deprecations, warn users of fields that will be removed in a new version, and give them an alternative in the changelog
```

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,259 @@
...
+- Consumers confirm when they're ready to switch; only after can the v1 be removed.
+- Both versions coexist during migration period
+
+**When to version:** Only when the change would break existing Consumer queries or code.
+
+**Key principle:** Versions are long-lived branches, not frequent releases. A data product might have only 2-3 major versions over its lifetime.
+
+### Access inheritance for major versions
+
+When a Producer creates v2 of an Output Port, all Consumers with active Handshakes on v1 automatically receive access to v2. They don't need to go through the access request flow again - requiring re-requests would recreate the exact problem we're solving today.
```
stijndehaes:
```suggestion
When a Producer creates v2 of an Output Port, all Consumers with an input port on v1 automatically receive access to v2. They don't need to go through the access request flow again - requiring re-requests would recreate the exact problem we're solving today.
```

docs/adr/0015-versioning.md
```diff
@@ -0,0 +1,259 @@
...
+**Initial state:**
+- Output Port "Customer Data" v1.2.0 → Table `customers` (id, name, email)
+
+**Change needed:**
+- Add `loyalty_tier` column
+
+**Steps:**
+1. Add column to existing table: `ALTER TABLE customers ADD COLUMN loyalty_tier VARCHAR(20)`
+2. Bump version to 1.3.0 in Portal, add changelog entry: "Added optional loyalty_tier column"
+3. Consumers are notified; no migration required - the new column is already available
```
stijndehaes: I think this is redundant, and can be removed for readability

stijndehaes (changes requested):

stijndehaes (approved):

pascal-knapen (approved):
Looks good Stijn. Just something I want to put inside your head for a moment. Sometimes you want to update something but not publish it yet, like we do with the release notes. I see something like that also needed for output ports. Not really versioning, more related to lifecyle

Issue #3211, opened 2026-04-20
Implement Explorations


stijndehaes:
- [x] Basic API with hierarchical joined sql alchemy implementation
- [ ] UX Flows:
   - [ ] existing data product
   - [ ] new data product
   - [ ] existing exploration
   - [ ] new data exploration
- [ ] add to explorer
- [ ] Add input ports support to explorations
- [ ] add RBAC to explorations
- [ ] Add Audit events to explorations
- [ ] Domain model has `data_product_count`, check how it is used, and if we need it for explorations too
- [ ] Add some explorations to samples

stijndehaes:
Feedback Katarina on first iteration of create exploration, with answers that were approved

- Initial designs had no description - less clicks for exploration -> Can do that later, we just make this optional on exploration, but not put it in the form
- we said domain is optional for exploration (it may be a question coming from CEO - what's the domain of that etc.) -> Can't we just remove it from the form then? We can just keep it on the object?
we said tags optional -> Tags are currently unsupported but can be added later when required
- why do we need a namespace? for the basic infra that is still needed to explore right? -> This is kind of a hard requirement in the backend now, but we can think about it. But we can maybe remove it from the form
- We need to handle empty cart better -> Was already on TODO
<img width="3400" height="1978" alt="Image" src="https://github.com/user-attachments/assets/6e6ed53c-743a-4a8f-a6cf-e63684119ecb" />

stijndehaes:
Fix consumers tab
<img width="1705" height="991" alt="Image" src="https://github.com/user-attachments/assets/e42d232b-1460-445b-9d9e-73ef676b7d85" />

PR #3212, opened 2026-04-20
Feature/3211 explorations api

stijndehaes:

Also check out todo list here:
https://github.com/conveyordata/data-product-portal/issues/3211


backend/app/configuration/domains/schema_response.py
```diff
@@ -12,41 +10,13 @@ class BaseDomainGet(ORMModel):
...
-            name=d.name,
-            description=d.description,
-            data_products=d.data_products,
-        )
-
-
-@deprecated("Use GetDomainsItem instead")
-class GetDomainsItemOld(BaseDomainGet):
-    data_product_count: int
+    pass
```
stijndehaes, line 14: I did remove data_products from the domain get api, this is a breaking change, but it wasn't used in the front end.
And I also don't think it was used? @stijn-janssens feel free to correct me

backend/app/explorations/router.py
```diff
@@ -0,0 +1,29 @@
...
+    CreateExplorationResponse,
+    GetExplorationsResponse,
+)
+from app.explorations.service import ExplorationService
+
+route = "/v2/explorations"
+router = APIRouter(tags=["Explorations"], prefix=route)
+
+
+@router.get("", response_model=GetExplorationsResponse)
```
stijndehaes, line 18: For reviewes this is the recommended way to fix our typing issue, add the response model instead of changing the typing of the route.

This ensure fast api knows it needs to return a `GetExplorationsResponse`, but we don't have to add the incorrect typing to ` ExplorationService(db).get_explorations()`, this should fix our mypy issues

backend/app/configuration/domains/schema_response.py
stijn-janssens: Think it's fine

backend/app/configuration/domains/model.py
```diff
@@ -19,11 +19,17 @@ class Domain(Base, BaseORM):
...
-    def data_product_count(self) -> int:
-        return len(self.data_products)
+    abstract_data_products: Mapped[list["AbstractDataProduct"]] = relationship(
+        "AbstractDataProduct", lazy="raise"
+    )
+
+    data_product_count = deferred(
+        select(func.count(literal_column("id")))
+        .select_from(text("abstract_data_products"))
+        .where(text("abstract_data_product_type = 'data_product'"))
```
stijn-janssens, line 29: do we only want to count normal data products here? I would at least have a second count then for the explorations as well.

backend/app/explorations/model.py
```diff
@@ -0,0 +1,16 @@
+from sqlalchemy import ForeignKey
+from sqlalchemy.dialects.postgresql.base import UUID
+from sqlalchemy.orm import Mapped, mapped_column
+
+from app.abstract_data_product.model import AbstractDataProduct
+
+
+class Exploration(AbstractDataProduct):
```
stijn-janssens, line 8: doesn't an exploration need assignments and input ports?

stijn-janssens: or if that is part of different PR then also fine

backend/app/explorations/router.py
```diff
@@ -0,0 +1,28 @@
...
+def get_explorations(
+    db: Session = Depends(get_db_session),
+):
+    return {"explorations": ExplorationService(db).get_explorations()}
+
+
+@router.post("", response_model=CreateExplorationResponse)
```
stijn-janssens: I would already add new AuthorizationActions on these calls. Can be in different PR but since this is the api focussed PR I would do it here

backend/app/explorations/schema_request.py
```diff
@@ -0,0 +1,10 @@
+from uuid import UUID
+
+from app.shared.schema import ORMModel
+
+
+class CreateExplorationRequest(ORMModel):
```
stijn-janssens, line 6: attaching input ports would be a separate call?

backend/app/graph/service.py
```diff
@@ -32,14 +32,15 @@ def get_graph_data(
                         """
                 SELECT
                     data_products.id as id,
-                    data_products.name as name,
-                    data_products.description as description,
+                    adp.name as name,
```
stijn-janssens, line 35: does this make sense? At least allow it to be toggled to show exploration products in the explorer or not. Can become a lot of useless clutter fast

backend/sample_data.sql
```diff
@@ -367,35 +389,52 @@ begin
...
+    It improves trial design and treatment personalization.', operations_and_logistics_domain_id, '2025-10-28 18:29:24.833167', NULL, NULL);
+    INSERT INTO public.data_products (id, about, status, type_id, lifecycle_id, usage) VALUES ('58b837a5-33d0-41cf-bf95-eb9af846f4d0', NULL, 'ACTIVE', '90ab1128-329f-47dd-9420-c9681bfc68c4', data_product_lifecycle_id, NULL);
+    INSERT INTO public.abstract_data_products (id, name, namespace, abstract_data_product_type, description, domain_id, created_on, updated_on, deleted_at) VALUES ('ccdc13fa-4a1a-4dde-ad1c-efa0d58eafb7', 'R&D Portfolio Prioritization Model', 'rd-portfolio-prioritization-model', 'data_products', 'Ranks R&D projects based on scientific potential, ROI, and strategic fit.
+    It provides leadership with an objective decision framework for resource allocation.', operations_and_logistics_domain_id, '2025-10-28 18:34:32.308251', NULL, NULL);
+    INSERT INTO public.data_products (id, about, status, type_id, lifecycle_id, usage) VALUES ('ccdc13fa-4a1a-4dde-ad1c-efa0d58eafb7', NULL, 'ACTIVE', 'f1672c38-ad1a-401a-8dd3-e0b026ab1416', data_product_lifecycle_id, NULL);
```
stijn-janssens, line 437: Maybe we also should add a couple of exploration products in the sample data?

backend/app/configuration/domains/model.py
stijndehaes: I think we want that for the future indeed :) But do we show this anywere?

I will add it as a todo

backend/app/explorations/model.py
stijndehaes: Correct! I already had RBAC, but added input ports on todo list

backend/app/explorations/schema_request.py
stijndehaes: Similar to create data product yes

backend/app/graph/service.py
stijndehaes: This is a mistake, I meant to filter out explorations and only keep data products. Had it on my todo to add explorations later

backend/sample_data.sql
stijndehaes: Currently they don't really work, will be added later

backend/app/configuration/domains/model.py
stijn-janssens: maybe we used to in the domain nodes. I have no clue right now

backend/app/graph/service.py
stijndehaes: Nevermind it is correctm but I added a test to check it anyway

stijndehaes: Added a test

backend/app/explorations/router.py
stijndehaes: Added RBAC support

stijn-janssens (approved):

PR #3226, opened 2026-04-21
Existing data product flow

stijndehaes:
In support of: #3211

Here is a small video of what has been built:
https://www.youtube.com/watch?v=-F-8eDBSHAY


stijndehaes:
https://www.youtube.com/watch?v=-F-8eDBSHAY

frontend/src/pages/cart-explorations/components/cart-overview.component.tsx
stijndehaes, line 1: For reviewers this is basically a copy of the old one, with some logic extracted into `useCartOverlapCheck`

frontend/src/tests/mocks/data_product.ts
stijndehaes, line 1: Logic moved to dataProduct.ts

frontend/src/tests/translation-glossary.test.ts
```diff
@@ -23,6 +23,10 @@ const GLOSSARY_TERMS: GlossaryRule[] = [
         expected: 'Technical Asset',
         pattern: /techincal asset/i,
```
stijn-janssens, line 24: this can't be right :P

frontend/src/routes.tsx
```diff
@@ -96,7 +97,8 @@ const router = createBrowserRouter([
                         path: ApplicationPaths.MarketplaceCart,
                         children: [
                             {
-                                element: <Cart />,
+                                element:
+                                    localStorage.getItem('explorations') === 'true' ? <ExplorationsCart /> : <Cart />,
```
stijn-janssens, line 101: localstorage now functions as your feature flag? Is that the best way of doing it? Shouldn't we use the config.js for this instead?

frontend/src/hooks/use-form-persist.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const hasPersistedData = useRef(localStorage.getItem(storageKey) !== null);
+    const [isRestored, setIsRestored] = useState(!hasPersistedData.current);
+
+    useEffect(() => {
+        const savedData = localStorage.getItem(storageKey);
+        if (savedData) {
+            try {
+                form.setFieldsValue(JSON.parse(savedData));
+            } catch (error) {
+                console.error('Failed to parse form data from localStorage', error);
```
stijn-janssens, line 14: Might be an edge case, but if users store their form and we update portal to an incompatible form and they load it in again they will see an error suddenly. Is that the behaviour we want?

frontend/src/pages/cart-explorations/components/card-selection.tsx
```diff
@@ -0,0 +1,51 @@
...
+    value: T;
+};
+
+type Props<T> = {
+    selectedChoice?: T;
+    setSelectedChoice: (t: T) => void;
+    options: CardSelectionOptions<T>[];
+};
+
+export const CardSelection = <T,>({ options, setSelectedChoice, selectedChoice }: Props<T>) => {
```
stijn-janssens, line 22: it's sad that antd does not support this out of the box. I think we can reuse this component in the technical asset form as well to select platforms and services?

stijn-janssens (commented):
Good PR, only some open comments or issues with previous code

frontend/src/tests/translation-glossary.test.ts
stijndehaes: https://github.com/conveyordata/data-product-portal/issues/3237

frontend/src/hooks/use-form-persist.tsx
stijndehaes: They won't see an error since this is just logged to console, we just will ignore the data

frontend/src/pages/cart-explorations/components/card-selection.tsx
stijndehaes: Probably, should I create a good beginner issue out of it? Maybe it's a bit too hard, though it's now rather specific

frontend/src/routes.tsx
stijndehaes: It's what we do on Conveyor, using localstorage allows us to easily test in any environment, we just tell people to run the following snippet in the console:

```
localStorage.setItem("explorations", true)
```

But then agian open to do it in another way. Advantage of config.js is that we can easily enabled it for a whole environment, but not for single users. I don't mind either way

stijn-janssens: fine to test it out like this for now. Maybe Posthog also has some feature flags / AB testing we can use? But not worth it to investigate rn

frontend/src/pages/cart-explorations/components/card-selection.tsx
stijn-janssens: Create an issue indeed, but I would not label it good beginner 😅

stijn-janssens (approved):

frontend/src/pages/cart-explorations/components/card-selection.tsx
stijndehaes: https://github.com/conveyordata/data-product-portal/issues/3238

frontend/src/routes.tsx
stijndehaes: Don't think that is useful, since that is only used when you want to measure stuff

PR #3239, opened 2026-04-22
Added new data product form in cart

stijndehaes:
Also refactored the old data product form
so most items are reused.

In support of: #3211

https://github.com/user-attachments/assets/cd6cce8b-0190-45df-b870-8fb21b38e2f7


frontend/src/pages/cart-explorations/components/new-data-product-form.tsx
```diff
@@ -0,0 +1,127 @@
...
+        [
+            requestDatasetAccessForDataProduct,
+            posthog,
+            createDataProduct,
+            navigate,
+            clearStorage,
+            cartOutputPorts,
+            dispatch,
+        ],
+    );
```
stijndehaes, line 87: Idea for the future, and question for reviewer.

Should we have a create data product that includes the links in one go? That way it either fails or succeeds. Now users can get into an inbetween state where the data product was created but the links failed.

stijn-janssens: I think this is indeed makes sense. If we don't do it as a single call and a failure happens then people need to switch tabs to "existing product" and rerequest access there. That feels very clunky. Ofcourse it is an edge case so would maybe leave it out for now but document it.

stijn-janssens (approved):

frontend/src/pages/cart-explorations/components/new-data-product-form.tsx
stijndehaes: Added to todo

PR #3253, opened 2026-04-23
Added input ports api

stijndehaes:
Also fixed the email sending and extracted
it to the abstract service

In support of: #3211

### DIFF
--- a/backend/app/abstract_data_product/schema_response.py
+++ b/backend/app/abstract_data_product/schema_response.py
@@ -1,4 +1,10 @@
+from uuid import UUID
+
+from pydantic import Field
+
 from app.abstract_data_product.model import AbstractDataProductType
+from app.data_products.output_ports.input_ports.schema import InputPortBase
+from app.data_products.output_ports.schema import OutputPort
 from app.shared.schema import ORMModel
 
 
@@ -6,3 +12,8 @@ class AbstractDataProductInfo(ORMModel):
     name: str
     namespace: str
     abstract_data_product_type: AbstractDataProductType
+
+
+class InputPort(InputPortBase):
+    output_port_id: UUID = Field(validation_alias="dataset_id")
+    output_port: OutputPort = Field(validation_alias="dataset")
--- a/backend/app/abstract_data_product/service.py
+++ b/backend/app/abstract_data_product/service.py
@@ -1,0 +1,180 @@
+from copy import deepcopy
+from datetime import datetime
+from typing import Sequence
+
+import pytz
+from fastapi import BackgroundTasks, HTTPException
+from sqlalchemy import UUID, select
+from sqlalchemy.orm import Session, selectinload
+from starlette import status
+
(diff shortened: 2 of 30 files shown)


backend/app/abstract_data_product/schema_response.py
```diff
@@ -1,8 +1,19 @@
...
 from app.shared.schema import ORMModel
 
 
 class AbstractDataProductInfo(ORMModel):
     name: str
     namespace: str
     abstract_data_product_type: AbstractDataProductType
+
+
+class InputPort(InputPortBase):
```
stijn-janssens, line 17: shouldn't this be placed in the input ports folder?

backend/app/abstract_data_product/service.py
```diff
@@ -0,0 +1,180 @@
...
+                    self.db
+                ).users_with_authz_action(
+                    dataset_link.dataset_id,
+                    Action.OUTPUT_PORT__APPROVE_DATAPRODUCT_ACCESS_REQUEST,
+                )
+                background_tasks.add_task(
+                    email.send_dataset_link_email(
+                        dataset_link.consuming_abstract_data_product,
+                        dataset_link.dataset,
+                        requester=deepcopy(actor),
```
stijn-janssens, line 177: I would filter out the actor from the approvers in this case

stijn-janssens: it seems this also does not happen in the original code? But I want to avoid sending an email to yourself if you are both requester and an approver

stijn-janssens (commented):
I have a generic question on this PR.
We add stuff to the abstract data product regarding input ports. First of all I'm not sure that is the correct spot, I would place it in a subfolder maybe, but I don't see any changes on the data product input port folder. Is this normal?
Do we still need that spot? How come only data products need input ports and explorations do not?

For me it would make sense to have an input port folder on abstract level and only have it there. Or have it in both subclasses.

backend/app/abstract_data_product/service.py
stijndehaes: Should I make a ticket out of this one?

stijn-janssens (approved):

stijndehaes:
> I have a generic question on this PR. We add stuff to the abstract data product regarding input ports. First of all I'm not sure that is the correct spot, I would place it in a subfolder maybe, but I don't see any changes on the data product input port folder. Is this normal? Do we still need that spot? How come only data products need input ports and explorations do not?
> 
> For me it would make sense to have an input port folder on abstract level and only have it there. Or have it in both subclasses.

The conclusion of this remark is a confusion between input ports of your data product, and of your output port. This graph explains it:
```
Input port -> Data product -> output port -> input port
```

On abstract data product we look at input port on the left side, in output ports we look at it from the other side, who are my consumers (show on the right), perhaps we should rename the right side. I'll make a ticket out of that.

PR #3267, opened 2026-04-24
Fix rbac migration to not depend on code

stijndehaes:
Before you were enticed to change the function
`initialize_prototype_roles` which shouldn't be
changed since it was part of migration.

However it was required by some tests to function, and other tests created the prototype roles themselves.

Changed tests to not clean up prototype roles
and reused them.

There are still some issues with the migration since the role object is still used in part of it. But
that is for another PR

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -10,7 +10,6 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.routing import APIRoute
 from starlette.middleware.base import BaseHTTPMiddleware
 
-from app.authorization.roles.service import RoleService
 from app.authorization.service import AuthorizationService
 from app.core.auth.device_flows.background_tasks import cleanup_device_flow_table_task
 from app.core.auth.jwt import oidc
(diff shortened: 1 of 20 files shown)


backend/app/main.py
```diff
@@ -65,10 +64,10 @@ async def log_middleware(request: Request, call_next):
 
 @asynccontextmanager
 async def lifespan(_: FastAPI):
-    db = next(database.get_db_session())
-    resync = RoleService(db).initialize_prototype_roles()
-    if resync or settings.AUTHORIZER_STARTUP_SYNC:
-        AuthorizationService(db).reload_enforcer()
+    with database.SessionLocal() as db:
```
stijndehaes, line 67: So this fixes an issue where a transaction would stay open when running `reload_enforcer`, this resulted in a lock on the roles table, resulting in my migration for a unique index failing.

This fixes it by running it in a proper transaction that is closed

stijn-janssens (approved):

PR #3277, opened 2026-04-27
Added Exploration to the UI

stijndehaes:
First version showing the input ports in a tab.
We will later improve upon this

In support of: #3211

https://github.com/user-attachments/assets/6089a875-5d13-450b-b00c-2da1b274ccce


stijn-janssens:
Couple of remarks based on the UI film alone, not looked at the code yet:
- I would prefer a different icon for the data products and the explorations.
- Does this only show my explorations or all? I would have a switch as with data products

backend/app/explorations/model.py
```diff
@@ -14,3 +22,17 @@ class Exploration(AbstractDataProduct):
...
+
+
+def ensure_exploration_exists(
+    id: UUID, db: Session, authenticated_user: "User", **kwargs
+) -> Exploration:
+    exp: Exploration = ensure_exists(id, db, Exploration, **kwargs)
+    if exp.owner_id != authenticated_user.id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Exploration access denied: you are not the owner",
```
stijn-janssens: This i find weird personally, what if someone requests access to your output port for an exploration? Then I would want to check out that exploration and see what else is in there. Is there a reason we make this so strict?

backend/app/explorations/service.py
```diff
@@ -32,14 +35,20 @@ def create_exploration(
...
         self.db.add(model)
         self.db.flush()
         return model
 
-    def get_explorations(self) -> Sequence[ExplorationModel]:
-        query = select(ExplorationModel).order_by(asc(ExplorationModel.name))
+    def get_explorations(self, authenticated_user: User) -> Sequence[ExplorationModel]:
+        query = (
+            select(ExplorationModel)
+            .where(ExplorationModel.owner_id == authenticated_user.id)
```
stijn-janssens: this answers my initial question, but I am not a big fan. This needs to be "auditable" by higher level people as well

stijn-janssens (commented):
My only remark is about the decision to only show 'my explorations' If this is a well-thought true decision I'm fine to stand aside with my opinion

backend/app/explorations/model.py
stijndehaes: So changed it to our default now :)

stijndehaes: I will use the owner in edits

stijn-janssens (approved):

PR #3278, opened 2026-04-27
Exclude requester from approver email recipients (closes #3265)

### LINKED ISSUE #3265: Do not send approvers an email when you are the approver
## Summary

In the current code we send emails, when a request is done to an output port. But if you are the approved we also send you an email, which is quite useless.

https://github.com/conveyordata/data-product-portal/pull/3253#discussion_r3137236083

### Motivation
Sending less spam to users

mvanhorn:
Closes #3265.

## Summary

When a user requested an output-port link, dataset access, or data-product role and was themselves one of the approvers, they got the resulting "Action Required" email about their own request. This filters the requester out of the approver-email recipients at the three call sites that triggered it.

## Changes

- \`backend/app/data_products/output_port_technical_assets_link/router.py\` — replaced the all-or-nothing \`if authenticated_user not in approvers\` guard with a \`other_approvers\` filter so the email still goes to the rest of the approver set when the requester happens to be one of them.
- \`backend/app/abstract_data_product/service.py\` — the input-port path added in #3253 sent to the full approver set; now filters \`actor\` out before mailing.
- \`backend/app/authorization/role_assignments/data_product/router.py\` (\`request_data_product_role_assignment\`) — same filter for the role-request flow. The sibling \`create_data_product_role_assignment\` path already short-circuits via the auto-approval branch (\`is_admin or user.id in approvers\`), so it didn't need a change.

When the filtered list is empty (requester was the sole approver), no email is sent — that preserves the existing auto-approval behavior covered by \`test_dataset_link_auto_approval_no_email_sent\`.

## Verification

- \`python3 -m py_compile\` on all three files passes.
- I don't have a local poetry/pytest env, so the existing CI suite is the test of record. The single-approver test referenced above asserts \`mock_send_email.assert_not_called()\`, which still holds (filtered list is empty -> branch not taken). The manual-approval test asserts \`assert_called_once()\`, also still holds (requester has no approve permission, so they're not in \`approvers\` to begin with).

### DIFF
--- a/backend/app/abstract_data_product/service.py
+++ b/backend/app/abstract_data_product/service.py
@@ -170,11 +170,15 @@ class AbstractDataProductService:
                     dataset_link.dataset_id,
                     Action.OUTPUT_PORT__APPROVE_DATAPRODUCT_ACCESS_REQUEST,
                 )
-                background_tasks.add_task(
-                    email.send_dataset_link_email(
-                        dataset_link.consuming_abstract_data_product,
-                        dataset_link.dataset,
-                        requester=deepcopy(actor),
-                        approvers=[deepcopy(approver) for approver in approvers],
+                other_approvers = [a for a in approvers if a != actor]
+                if other_approvers:
+                    background_tasks.add_task(
+                        email.send_dataset_link_email(
+                            dataset_link.consuming_abstract_data_product,
+                            dataset_link.dataset,
+                            requester=deepcopy(actor),
+                            approvers=[
+                                deepcopy(approver) for approver in other_approvers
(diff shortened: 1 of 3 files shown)


stijn-janssens:
We are aware that the email sending functionality is currently not being tested enough, but could you please make sure to write some unit tests to ensure that this issue will not occur again in the future. Thanks a lot!

stijn-janssens:
Hi @mvanhorn, I will close this PR because it has been inactive for a long period.
Feel free to open a new one and implement my previous comments! Thanks for your input

PR #3279, opened 2026-04-27
Feature/mcp athena query access

stijn-janssens:
The system relies heavily on certain patterns that are used in the underlying infrastructure. Most of these are hardcoded and expected to be there. e.g. workgroups in Athena, Athena output paths in buckets, database suffixes with double underscores, ...


backend/app/settings.py
```diff
@@ -22,6 +22,9 @@ class Settings(BaseSettings):
     AWS_DEFAULT_PROFILE: Optional[str] = None
     PORTAL_API_KEY: Optional[str] = None
 
+    # AWS Athena Configuration (company-wide settings)
```
stijndehaes: I am unsure if this is a good idea? But I also haven't analysed it

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+    TWO DISTINCT MODES OF OPERATION
+    ═══════════════════════════════════════════════════════════════════════
+
+    MODE 1: DISCOVERY (Metadata Only - Fast, No Credentials Required)
+    ──────────────────────────────────────────────────────────────────────
+    Use this when the user wants to:
+    ✓ Find what data exists: "What sales datasets are available?"
+    ✓ Get metadata: "Who owns the customer data?", "What's in the sales_db?"
+    ✓ Explore the catalog: "Show me all data products in the finance domain"
+    ✓ Check ownership/descriptions: "What does the revenue_monthly table contain?"
```
stijndehaes:
```suggestion
    - Find what data exists: "What sales datasets are available?"
    - Get metadata: "Who owns the customer data?", "What's in the sales_db?"
    - Explore the catalog: "Show me all data products in the finance domain"
    - Check ownership/descriptions: "What does the revenue_monthly table contain?"
```

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+
+    Tools for Discovery (no credentials needed):
+    • search_output_ports(query) - find datasets
+    • search_data_products(query) - find data products
+    • get_output_port_details(id) - metadata about a dataset
+    • get_data_product_details(id) - details about a data product
+    • get_data_product_analytics(id) - what output ports a data product has
+    • get_marketplace_overview() - high-level statistics
+    • get_environments() - available environments
```
stijndehaes:
```suggestion
    - search_output_ports(query) - find datasets
    - search_data_products(query) - find data products
    - get_output_port_details(id) - metadata about a dataset
    - get_data_product_details(id) - details about a data product
    - get_data_product_analytics(id) - what output ports a data product has
    - get_marketplace_overview() - high-level statistics
    - get_environments() - available environments
```

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+
+    MODE 1: DISCOVERY (Metadata Only - Fast, No Credentials Required)
+    ──────────────────────────────────────────────────────────────────────
+    Use this when the user wants to:
+    ✓ Find what data exists: "What sales datasets are available?"
+    ✓ Get metadata: "Who owns the customer data?", "What's in the sales_db?"
+    ✓ Explore the catalog: "Show me all data products in the finance domain"
+    ✓ Check ownership/descriptions: "What does the revenue_monthly table contain?"
+
+    Tools for Discovery (no credentials needed):
```
stijndehaes: Could we make this a dynamic list somehow?

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+    • get_environments() - available environments
+
+    MODE 2: DATA QUERIES (Actual Data - Requires AWS Credentials)
+    ──────────────────────────────────────────────────────────────────────
+    Use this when the user wants to:
+    ✓ See actual data: "Show me sales from last month"
+    ✓ Run analytics: "What are the top 10 customers by revenue?"
+    ✓ Query tables: "SELECT * FROM sales_db.transactions WHERE date > '2026-01-01'"
+    ✓ Get row-level data: "How many orders were placed yesterday?"
+    ✓ Find records in data: "Which users are in the sales data?"
```
stijndehaes:
```suggestion
    - See actual data: "Show me sales from last month"
    - Run analytics: "What are the top 10 customers by revenue?"
    - Query tables: "SELECT * FROM sales_db.transactions WHERE date > '2026-01-01'"
    - Get row-level data: "How many orders were placed yesterday?"
    - Find records in data: "Which users are in the sales data?"
```

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+
+    B. If ALL consuming products fail, try the owner data product:
+       get_aws_credentials(owner_namespace, environment)
+
+    C. If EVERYTHING fails → Tell user:
+       "No access to this data. Request access from [owner] or these
+        consuming data products: [list names from data_product_links]"
+
+    ⚠️  REMEMBER: Once you find working credentials, use that SAME namespace
+        for query_athena in the following steps!
```
stijndehaes: This is pretty hardcoded for AWS and athena, can we make this more broad?

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+      ✓ CORRECT: SELECT * FROM "datalake_experimentation_sales-data__sales"."users"
+      ✗ WRONG:   SELECT * FROM users (missing database)
+      ✗ WRONG:   SELECT * FROM datalake_experimentation_sales-data__sales.users (no quotes)
+
+    Step 8: GET RESULTS
+    ───────────────────
+    get_athena_query_results(query_execution_id, data_product_namespace, environment)
+    → If status = 'RUNNING': Wait 3-5 seconds, retry
+    → If status = 'SUCCEEDED': Return formatted results to user
+    → If status = 'FAILED': Show error message
```
stijndehaes: Again this is fairly direct for your use case? Should we have that in the general MCP?

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+       → database="mcp-athena-query-test"
+       → data_product_links: [{data_product: {namespace: "mcp-test-sales-overview"}}]
+    3. get_environments() → Use 'dev'
+    4. Try consuming product FIRST:
+       get_aws_credentials("mcp-test-sales-overview", "dev")
+       → ✓ Success! Use this namespace for all queries
+    6. query_athena("mcp-test-sales-overview", "dev",
+         'SELECT * FROM "datalake_experimentation_mcp-athena-query-test__sales"."users" LIMIT 100')
+       → Note: Quotes around database/table names with hyphens!
+    7. get_athena_query_results(...) → Returns user data
```
stijndehaes: Are these examples necessary to get better results?

backend/app/mcp/mcp.py
```diff
@@ -72,14 +88,218 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+    • Don't use unqualified table names (missing database prefix)
+    • Don't forget quotes around database/table names with hyphens or underscores
+
+    ═══════════════════════════════════════════════════════════════════════
+    MCP COMPATIBILITY
+    ═══════════════════════════════════════════════════════════════════════
+
+    This flow is designed for MCP (Model Context Protocol) servers:
+    • Discovery queries are synchronous and fast (< 1 second)
+    • Data queries are asynchronous (Athena takes 3-30 seconds)
```
stijndehaes: Should this be here?

backend/app/mcp/mcp.py
```diff
@@ -90,16 +310,20 @@ def get_auth_provider() -> Optional[JWTVerifier]:
...
+        )
+        return {
+            "id": user.id,
+            "external_id": user.external_id,
+            "first_name": user.first_name,
+            "email": user.email,
+            "last_name": user.last_name,
+        }
+    finally:
+        db.close()
```
stijndehaes: with database.SessionLocal() as db:
  user = get_authenticated_user(
      token=JWTToken(sub="", token=f"Bearer {token}"), db=db
  )
  return {
      "id": user.id,
      "external_id": user.external_id,
      "first_name": user.first_name,
      "email": user.email,
      "last_name": user.last_name,
  }

backend/app/mcp/mcp.py
```diff
@@ -111,6 +335,288 @@ async def get_current_user(ctx: Context) -> dict[str, Any]:
...
+
+
+@mcp.tool
+def get_database_prefix(environment: str) -> str:
+    prefix = settings.AWS_ATHENA_PREFIX
+    return f"{prefix}_{environment}_"
+
+
+@mcp.tool
+def get_glue_database(environment: str, technical_asset_id: str) -> str:
```
stijndehaes: We now have glue and athena, this means not all technical asset types are supported?

backend/app/mcp/mcp.py
```diff
@@ -111,6 +335,288 @@ async def get_current_user(ctx: Context) -> dict[str, Any]:
...
+
+        # Unknown status
+        result["message"] = f"Unexpected query status: {status}"
+        return result
+
+    except Exception as e:
+        return {
+            "error": f"Failed to get query results: {str(e)}",
+            "query_execution_id": query_execution_id,
+        }
```
stijndehaes: This is super specific should this not be part of the technical asset plugin?

stijndehaes (changes requested):
I have some problems with this PR. While I know the idea is to have this as an experiment there are some flaws with the approach:
- we are adding query options for certain technical assets, should we not add these to the plugins?
- This will be a poor experience when technical assets are not supported, or what even happens at the moment?
- We added a setting called `AWS_ATHENA_PREFIX` only used in MCP, we do not even have athena technical assets at the moment right?

I wouldn't mind having a call to see how we can fix this :) 

Maybe not a bad option to make this a separate thing?

.github/workflows/docker-build-main.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - 'main'
+      - feature/mcp-athena-query-access
```
stijndehaes:
```suggestion
```

backend/app/mcp/loader.py
```diff
@@ -0,0 +1,57 @@
+"""Plugin loader: auto-discovers and registers all MCP plugin tools.
```
stijndehaes: Can't we add a method to the `AssetProviderPlugin` to list all MCP tools instead?
Then we can add a method to the `PluginService` to register all tools of all registered plugins no?

backend/app/settings.py
stijndehaes: Still unsure if this is a good idea, shouldn't the technical asset fully define where it's located? So maybe the glue technical asset should have an optional field that specifies the athena DB

backend/pyproject.toml
```diff
@@ -14,6 +14,7 @@ authors = ["Stijn Janssens <stijn.janssens@dataminded.com>"]
 readme = "README.md"
 
 [tool.poetry.dependencies]
+data-product-portal-sdk = { path = "../sdk" }
```
stijndehaes: Why is this added?

stijndehaes (changes requested):

backend/pyproject.toml
stijn-janssens: It's there because I want to fetch the AWS credentials via the SDK. If I would call the service directly I'd skip the authentication checks, allowing the credentials to be given out always, not conditionally on whether you have access to the product or not. By including the sdk I can just call the sdk api method with correct auth.

Do you have a proposal on how to implement it cleaner in another way?

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,454 @@
...
+            return {
+                "error": "You don't have access to this data product or it doesn't exist."
+            }
+        return {
+            "AccessKeyId": result.access_key_id,
+            "SecretAccessKey": result.secret_access_key,
+            "SessionToken": result.session_token,
+        }
+    finally:
+        db.close()
```
stijndehaes: Please rewrite this to use:
`db: Session = Depends(get_db_session),`
and also depends for fetching the token, this in itself could be a dependency no?

backend/pyproject.toml
stijndehaes: Ah I understand, my problem with that is that we did not use that pattern for all the other calls.
I don't think it's a bad pattern to be honest and maybe one we should use everywhere. 
An improvement would be to use `httpx.ASGITransport` btw this reduces network overhead

The other option we could do is to add authorization to that endpoint, similarly to the API endpoints. WDYT

backend/app/database/database.py
```diff
@@ -28,9 +29,11 @@ def ensure_exists(
 
 
 def get_url(async_: bool = False) -> str:
+    user = quote_plus(settings.POSTGRES_USER)
+    password = quote_plus(settings.POSTGRES_PASSWORD)
     return (
-        f"postgresql{'+asyncpg' if async_ else ''}://{settings.POSTGRES_USER}:"
-        f"{settings.POSTGRES_PASSWORD}@{settings.POSTGRES_SERVER}:"
+        f"postgresql{'+asyncpg' if async_ else ''}://{user}:"
```
stijndehaes, line 35: Why are these changes needed? More out of interest this does not block approval

backend/app/mcp/mcp.py
```diff
@@ -1,23 +1,29 @@
-from contextlib import asynccontextmanager
-from typing import Any, Optional, Sequence
+"""Data Product Portal MCP server.
```
stijndehaes: Is this comment useful?

backend/app/mcp/mcp.py
```diff
@@ -354,37 +281,76 @@ def search_data_products(
...
+            ],
+            "count": len(filtered_data_products),
+            "filters_applied": {
+                "query": query,
+                "domain_id": domain_id,
+                "status": status,
+            },
+        }
+
+    return _execute(db)
```
stijndehaes: What is the reason to change this like it happened here?

I mean the pattern:
```
def search_data_products(...):
    def _execute(db):
         ...
    return _execute(db)
```

backend/app/mcp/mcp.py
```diff
@@ -454,8 +423,20 @@ def get_data_product_details(
     the data product it belongs to, and owner contact information.
     Use after search_output_ports to get complete information about a specific dataset.
 
+    CRITICAL FOR DATA QUERIES: This returns data_product_links[], which contains the consuming
```
stijndehaes: Is this correct? Doesn't it make more sense to go the other way around? For a certain data product you check it's input ports, and that is how you can figure out what you have access to? 

Anyway maybe in a future iteration it would be nice if you can just requests all output ports you have access to no?

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,504 @@
...
+    """
+
+    try:
+        authorize_data_product_read_integrations(
+            data_product_namespace=data_product_namespace,
+            authorized_user=authorized_user,
+            db=db,
+        )
+    except (ValueError, PermissionError) as e:
+        return {"error": str(e)}
```
stijndehaes: Is this actually the pattern we want with the MCP server? Doesn't it make more sense to handle error translate by the MCP server or is that not possible? I have been wondering this before.

According to my research this is not what we want it is recommended to either:
- Just let the exception be bubbled up
- Translate it to a ToolError

Otherwise the LLM might thing the MCP call did succeed, while it didn't

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,504 @@
...
+        if not isinstance(result, AWSCredentials):
+            return {"error": "Invalid credentials format"}
+        return {
+            "AccessKeyId": result.AccessKeyId,
+            "SecretAccessKey": result.SecretAccessKey,
+            "SessionToken": result.SessionToken,
+        }
+
+    @mcp.tool
+    def get_glue_database(
```
stijndehaes, line 159: I am going to be honest, I tried to read this code, but it's super long and confusing, I don't actually know what is happening, but don't we return this info when you go to a technical asset page and you look at the technical information? So can't we reuse that code?

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,504 @@
...
+    ────────────────────
+    get_athena_query_results(query_execution_id, data_product_namespace, env)
+    → RUNNING → wait 3-5 s and retry
+    → SUCCEEDED → return formatted rows
+    → FAILED    → show error
+"""
+
+
+def register_tools(mcp: FastMCP) -> None:
+    @mcp.tool
```
stijndehaes, line 127: Is this tool required since you call fetch aws credentials in `query_athena` and `list_glue_tables`?

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,504 @@
...
+                        "description": t.get("Description", ""),
+                        "table_type": t.get("TableType", ""),
+                        "created_at": str(t.get("CreateTime", "")),
+                        "updated_at": str(t.get("UpdateTime", "")),
+                    }
+                    for t in response.get("TableList", [])
+                )
+                next_token = response.get("NextToken")
+                if not next_token:
+                    break
```
stijndehaes: This code can be simplified by using a paginator:
https://docs.aws.amazon.com/boto3/latest/reference/services/glue/paginator/GetTables.html

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,504 @@
...
+            bucket: S3 bucket for results (from get_glue_database). Optional.
+            workgroup: Athena workgroup (from get_glue_database). Falls back to server default if absent.
+        Returns:
+            {'query_execution_id': '...', ...} or {'error': '...'}.
+        """
+        creds = _fetch_aws_credentials(data_product_namespace, env, authorized_user, db)
+        if isinstance(creds, dict) and "error" in creds:
+            return creds
+        if not isinstance(creds, AWSCredentials):
+            return {"error": "Invalid credentials format"}
```
stijndehaes: So if we just raise toolerrors, or raise other error, we don't need to do these things

backend/app/technical_asset_configuration/glue/mcp_tools.py
```diff
@@ -0,0 +1,504 @@
...
+            response = client.start_query_execution(**kwargs)
+            result = {
+                "query_execution_id": response["QueryExecutionId"],
+                "data_product_namespace": data_product_namespace,
+                "environment": env,
+                "query": query,
+                "status": "Query submitted. Use get_athena_query_results to poll for results.",
+            }
+            if resolved_workgroup:
+                result["workgroup"] = resolved_workgroup
```
stijndehaes:
```suggestion
            kwargs: Dict[str, Any] = {"QueryString": query}
            if workgroup:
                kwargs["WorkGroup"] = workgroup
            if bucket:
                kwargs["ResultConfiguration"] = {
                    "OutputLocation": f"s3://{bucket}/athena-results"
                }
            response = client.start_query_execution(**kwargs)
            result = {
                "query_execution_id": response["QueryExecutionId"],
                "data_product_namespace": data_product_namespace,
                "environment": env,
                "query": query,
                "status": "Query submitted. Use get_athena_query_results to poll for results.",
            }
            if workgroup:
                result["workgroup"] = workgroup
```

stijndehaes (changes requested):

backend/app/database/database.py
stijn-janssens: FastMCP get url requires url encodings

backend/app/technical_asset_configuration/glue/mcp_tools.py
stijn-janssens: this tool is also used to check access up front, before trying to query athena or list glue tables, to see if we have the right namespace to query from

backend/app/mcp/mcp.py
stijn-janssens: Yes it also makes sense, but you rarely specify from which data product you would want to query data. You specify which data you would like to see, from which the mcp server can easily look up descriptions and find the owner data product, but then it needs to check if you have access through a consuming data product

backend/app/technical_asset_configuration/glue/mcp_tools.py
stijndehaes: Do we need to fetch credentials for that? Wouldn't it be better if you have an endpoint that just tells you which data products you have access to?

backend/app/mcp/mcp.py
stijndehaes: mmmm I understand, but does that mean we let the AI combine that from multiple input ports of data products you have access to? Is that not something we should not encourage?

backend/app/technical_asset_configuration/glue/mcp_tools.py
stijndehaes: Make future improvement to make this a seperate tool call to check access, or just list of data products/output ports you have access to

stijndehaes (approved):
Only thing to fix is remove the translation of error to tool error, as it does not bring any value:
https://gofastmcp.com/servers/tools#error-handling

Issue #3266, opened 2026-04-24
Improve naming of input ports in output ports


stijndehaes:
## Summary
See discussion here:
https://github.com/conveyordata/data-product-portal/pull/3253#issuecomment-4312781929

It is quite confusing, and it would help to find a new name of the input ports folder in output ports. Something like consumptions or something

### Motivation
Clears up confusion

PR #3293, opened 2026-04-28
Fixed an issue with SDK clash

stijndehaes:
The refactor of input ports resulted in
a clash in the SDK with 2 things called InputPort

In support of: #3211

### DIFF
--- a/backend/app/data_products/output_ports/input_ports/router.py
+++ b/backend/app/data_products/output_ports/input_ports/router.py
@@ -18,7 +18,7 @@ from app.data_products.output_ports.input_ports.schema_request import (
 )
 from app.data_products.output_ports.input_ports.schema_response import (
     GetInputPortsForOutputPortResponse,
-    InputPort,
+    OutputPortInputPort,
 )
 from app.data_products.output_ports.input_ports.service import DataProductDatasetService
 from app.data_products.output_ports.service import OutputPortService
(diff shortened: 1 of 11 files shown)


pascal-knapen (approved):
Noticed it as well and was thinking of similar fix tbh.

backend/app/data_products/output_ports/input_ports/router.py
```diff
@@ -41,7 +41,7 @@ def get_input_ports_for_output_port(
 ) -> GetInputPortsForOutputPortResponse:
     return GetInputPortsForOutputPortResponse(
         input_ports=[
-            InputPort.model_validate(x)
+            OutputPortInputPort.model_validate(x)
```
stijn-janssens, line 44: Not sure I agree on the new naming.
The way I currently saw it (but I might be wrong)
Input Port = the link between output port and data product, with a justification. But only has the output port metadata

OutputPortInputPort = input port extended with information about the consuming data product? I would call it DataProductInputPort if that is the case.

stijndehaes: We want to fix it here:

https://github.com/conveyordata/data-product-portal/issues/3266

But for now we fix the SDK, we want to improve this in the future

stijn-janssens (approved):

PR #3297, opened 2026-04-28
Simplify exploration creation

stijndehaes:
Reduced the form to minimal setup currently
required.
So description is copied from justification
and namespace is automatically generated.

In support of: #3211

Screenshot 2026-04-28 at 15 12 15

### DIFF
--- a/backend/app/data_products/service.py
+++ b/backend/app/data_products/service.py
@@ -21,10 +21,6 @@ from app.configuration.tags.model import Tag as TagModel
 from app.configuration.tags.model import ensure_tag_exists
 from app.configuration.tags.schema import Tag
 from app.core.aws.get_url import _get_data_product_role_arn
-from app.core.namespace.validation import (
-    NamespaceValidator,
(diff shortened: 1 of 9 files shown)


stijn-janssens (changes requested):
If I understand this correctly this will be the behavior:
- fill in form with non-unique name (resulting in invalid namespace)
- successfully submit form
- get error message
I would prefer the namespace to be checked upfront and show an error on the name.

stijndehaes:
> If I understand this correctly this will be the behavior:
> 
> * fill in form with non-unique name (resulting in invalid namespace)
> * successfully submit form
> * get error message
>   I would prefer the namespace to be checked upfront and show an error on the name.

You are right, I made this a bit too simple. I also just realised that my namespace validation might be wrong, the namespace has to be unique across data product and exploration in the DB at the moment, but that isn't properly checked.

stijndehaes:
Closed in favor of: https://github.com/conveyordata/data-product-portal/pull/3320

PR #3304, opened 2026-04-29
Added exploration icon

stijndehaes:
Also put icons into folders to make it a bit easier to see which belong together and what their purpose is.

Also changed outline icons to have the same viewport and ensured they fill the viewport. That way they have a consistent size when put next to ant design icons that fill the viewport.

In support of: #3211


stijndehaes:
Exploration icon in tab
image
Exploration border icon at the top, will be used in explorer too
image

Consumers icon is just the data product and exploration icon next to each other
image

stijn-janssens (approved):

PR #3307, opened 2026-04-29
Added sticky cart summary

stijndehaes:
In support of: #3211

- [X] When updating the UI please provide screenshots or videos to the reviewers.

https://github.com/user-attachments/assets/03b8ee87-586c-43cf-b982-62429b47acc1

Worst case when there are many items causing a vertical overflow:

https://github.com/user-attachments/assets/bc8863d4-d6fd-4c7d-942f-d15f05f12254

### DIFF
--- a/frontend/src/pages/cart-explorations/cart.page.tsx
+++ b/frontend/src/pages/cart-explorations/cart.page.tsx
@@ -181,7 +181,7 @@ function ExplorationsCart() {
                     {form}
                 </Flex>
             </Col>
-            <Col span={8}>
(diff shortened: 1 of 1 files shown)


stijn-janssens (approved):

stijn-janssens:
I don't like the overflow but I don't think it is a realistic case / have no alternative in mind

stijndehaes:
> I don't like the overflow but I don't think it is a realistic case / have no alternative in mind

Agree that scrolling is imperfect at that point, but don't think it will happen in real life. I just wanted to ensure that if it happens at least things aren't broken

PR #3319, opened 2026-04-30
Show required by default in all forms

stijndehaes:
It was confusing that it sometimes shows
up and sometimes doesn't. It's better for
users to always show it that way they know
things are required.

Also removed autocomplete from the exploration
form since it was useless.

In support of: #3211


stijn-janssens (approved):
What is the impact of this?

PR #3428, opened 2026-05-15
Migrate data product portal in single image

stijndehaes:
This simplifies deployment


Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+COPY backend/requirements-poetry.txt .
+RUN pip install -r requirements-poetry.txt --require-hashes
+
+COPY backend/poetry.lock backend/pyproject.toml backend/alembic.ini backend/sample_data.sql /
+RUN poetry install --no-root
+
+# ---------------------------------------------------------------------------
+# Stage 3 – final runtime image
+# ---------------------------------------------------------------------------
+FROM --platform=${PLATFORM} python:3.13.13-slim-bookworm@sha256:cfdfc1c32835000e04bf476b7679121da79b4a1e21e2a352985b40b3c275c0c8
```
vlieven: We should define this image only once, we currently have this in the backend Dockerfile, I would reuse the pattern:

ARG PLATFORM=linux/amd64
FROM --platform=${PLATFORM} python:3.13.13-slim-bookworm@sha256:cfdfc1c32835000e04bf476b7679121da79b4a1e21e2a352985b40b3c275c0c8 AS base-image

FROM base-image AS build-stage
// builder stuff

FROM base-image
// final stuff

backend/app/main.py
```diff
@@ -181,6 +176,18 @@ def use_route_names_as_operation_ids(app: FastAPI) -> None:
 
 
 use_route_names_as_operation_ids(app)
+
+if settings.SERVE_FRONTEND:
```
vlieven, line 180: In which scenario do we not want to serve the frontend?

backend/app/main.py
```diff
@@ -181,6 +176,18 @@ def use_route_names_as_operation_ids(app: FastAPI) -> None:
...
+    _frontend_dir = Path(__file__).parent.parent / settings.FRONTEND_DIST_DIR
```
vlieven: Can't we just define `settings.FRONTEND_DIST_DIR` as the correct location, and not mess with the `Path().parent.parent`?

backend/app/main.py
```diff
@@ -155,12 +156,6 @@ async def send_response_to_webhook(request: Request, call_next):
     return response
 
 
-# K8S health and liveness check
-@app.get("/", include_in_schema=False)
-def root():
-    return {"message": "Hello World"}
```
vlieven, line 161: Are we not running a health check?

Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+COPY --from=python-build /usr/local /usr/local
+COPY --from=python-build alembic.ini .
+COPY --from=python-build sample_data.sql .
+
+# Copy backend application
+COPY backend/app ./app
+COPY backend/templates ./templates
+COPY backend/alembic.ini ./alembic.ini
+COPY backend/VERSION ./VERSION
+COPY backend/sample_data.sql ./sample_data.sql
```
vlieven, line 61: We're doing a lot of different COPY steps here? Also, we are copying `alembic.ini` and `sample_data.sql` twice and and overwriting them

backend/app/main.py
stijndehaes: When we run locally

backend/app/main.py
stijndehaes: The root now returns index.html so that is used as health check :)

Dockerfile
stijndehaes: Removed duplicated copies, and the lot's of different copies are similar to the old dockerfile

Dockerfile
stijndehaes: Agree

backend/app/main.py
vlieven: I can see there are scenario's where it is indeed not necessary to serve the frontend locally, but I don't really see how it hurts to just have it anyway. I think the extra setting adds a bit of unnecessary complexity, but not a deal breaker if you think this is better

backend/app/main.py
vlieven: This was not resolved 😅

vlieven (approved):

PR #3443, opened 2026-05-18
Remove deploys from releases

stijndehaes:
We want our demo environment to be stable
so not automatic releases. And our dev env
will allready have received this version


nclaeys:
I do not get why you do not want demo to have the latest released version?

In my opinion demo should be up to date with the latest released version in order to show new features. As a release should be done on a stable version this should not result in any downtime. When you remove it the gap between what we are working on and what is visible on demo grows even bigger than it is now.

stijndehaes:
> not get why you do not want demo to have the latest released version?
> 
> In my opinion demo should be up to date with the latest released version in order to show new features. As a release should be done on a stable version this should not result in any downtime. When you remove it the gap between what we are working on and what is visible on demo grows even bigger than it is now.

I agree that demo should have the latest release that is possible.

However I have rewritten the other side of this release process because of issues with the previous version, making it non compatible with this sadly. But it fixes other issues (mainly commits to main not triggering an automated release).

I'll see what can be done to restore the release to demo, but for now this does not work

vlieven:
@stijndehaes Should we close or merge this PR? Do we need a follow-up PR to restore the automated demo deployments in another way?

PR #3460, opened 2026-05-19
Create backend implementation for data contract ingestion/retrieval 

nclaeys:


stijn-janssens (commented):
Is / will there be in the future any link between technical assets and the output port schemas? Can we enforce this in portal in the future? Or do we want this to be solely the users end responsibility?
The last option does not seem favourable to me, but that is what this initial implementation offers ( I believe ). Does it make sense to already tie it to the technical asset concepts we have? Or is that overkill for now

nclaeys:
> Is / will there be in the future any link between technical assets and the output port schemas? Can we enforce this in portal in the future? Or do we want this to be solely the users end responsibility? The last option does not seem favourable to me, but that is what this initial implementation offers ( I believe ). Does it make sense to already tie it to the technical asset concepts we have? Or is that overkill for now

Well at the moment there is no link between schemas and technical assets. I do not follow why this will not be favorable. As the output port exposes it's information using a standard data contract spec, I do not really see what you want to do to link them? 

In the future you can ofcourse make the link more explicit if you also want to offer clear access to the underlying assets, but that was not the scope of this feature as far as I know.

stijn-janssens (approved):

PR #3484, opened 2026-05-21
Add OpenAPI export flag to exclude deprecated routes

RobbertDM:
Introduces a `--exclude-deprecated` flag to `app.open_api_export` that filters out routes marked as deprecated.
The openapi spec that rolls out (`docs/static/openapi-no-deprecated.json`) can be used for SDK generation.

- Add `_route_filter` keeping only non-deprecated `/api/v2/*` routes
- Some Taskfile and .tool-versions additions for local setup
- Also there is this `check_access` function that did not play nicely with the SDK generation. This PR surrounds a `Query()` around it so that it is explicitly a query parameter. Then the SDK generation is correct.


stijn-janssens:
Discussed in person:
This feels to me like it defeats the point of deprecating an API. We like to keep everything in sync throughout all the interfaces, allowing downstream users to migrate over time.
To be checked: Potentially the tool could not handle the deprecated routes, please investigate :)

stijn-janssens (approved):
I updated your code by removing the non deprecated openapi spec generation, but I liked all of the other additions. Ping me if you do not agree :)

PR #2235, opened 2025-12-06
initial version of go sdk and terraform provider

RobbertDM:
Hey guys, no hard feelings if you close this PR, it was generated by Opus so I did not spend that much time myself. Altough I believe releasing a tf provider would be valuable - even with just latest/latest compatibility. If this PR is closed, what I could do is release it on the tf registry under my own name as a "community-contributed" tf provider. Maybe that is the best way forward anyways?

- [ ] Update the release notes document if needed in the [release notes](../docs/docs/release-notes.md)
- [ ] When updating the UI please provide screenshots or videos to the reviewers

PR #3485, opened 2026-05-21
go sdk

RobbertDM:
This does nothing but commit the files generated by https://github.com/doordash-oss/oapi-codegen-dd 
This is what made my [original PR](https://github.com/conveyordata/data-product-portal/pull/2235) so scary big, but it's just artifacts that I think you would want to commit (so you can also see the diff easily if you run against new versions of the API).


stijn-janssens:
Hi @RobbertDM,

A couple of things:
- We discussed in person about the non-deprecated version of openapi. I am not a big fan of that personally.
- We were working in parallel on fixing the golang CLI for the v2 version of the API. In doing that we regenerated a golang sdk, which looks to be more complete than the one you generate in this PR. For example I seemed to missing the /deviceflow routes
- We are not planning to actively maintain and release a golang sdk for now, other than the CLI commands which we support.

For those reasons I'll close your PR for now but feel free to ping me if you have questions or if you disagree with this decision

PR #3539, opened 2026-06-01
Docs/event payload scoping spec

pascal-knapen:


docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+The provisioner creates an IAM role for the exploration so the owner can assume it to access S3 locations they have been granted read access to via input ports. No S3 prefix is created — explorations do not produce or store data.
+
+### Exploration Created
+
+The provisioner needs the exploration's namespace and the owner's identity.
+
+1. Create IAM role `dpp-{namespace}` with the owner as the sole principal in the trust policy.
+2. Update the exploration status from `Draft` to `Ready`.
+
+> Explorations do not own storage — no S3 prefix or technical asset is created.
```
vlieven, line 285: Don't they require some sort of scratch space that is not shared?

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+
+The provisioner needs the data product's namespace, domain, type, and the list of initial team members (with their roles).
+
+1. Resolve the domain bucket name from the domain (by convention: `{domain-slug}`).
+2. Create the IAM role `dpp-{namespace}`:
+   - Trust policy: include all initial team members as principals immediately. Creation always includes at least one owner; do not wait for `team_member_added` events to populate the trust policy.
+   - Permission policy: read/write access to `s3://{domain-bucket}/{namespace}/*`.
+   - If the data product type confers additional permissions (e.g. cross-account access for a shared type), apply them now.
+3. Create the S3 prefix by uploading a placeholder or relying on the first write.
+4. Call the Portal API to register the subfolder as a `default` technical asset on this data product. This triggers a `technical_asset.created` event — that handler will update the IAM role. Make the handler idempotent so re-running it on an already-existing prefix is safe.
```
stijndehaes:
```suggestion
4. Call the Portal API to register the subfolder as a `default` technical asset on this data product. This triggers a `technical_asset.created` event. That handler should be idempotent and see we already have access.
```

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+5. Update the data product status from `Draft` to `Ready`.
+
+> **Note**: The data product type may influence the base permissions attached to the IAM role. The exact mapping is platform-specific and should be documented per type. Tags and status could similarly influence permissions, but their semantics are not yet defined.
+
+---
+
+### Data Product Updated
+
+The provisioner needs the data product's namespace, the old and new domain, the new type, and the list of current technical assets (with their S3 configuration).
```
stijndehaes, line 108:
```suggestion
The provisioner needs the data product's namespace, the new domain, the new type, and the list of current technical assets (with their S3 configuration).
```

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+
+1. Identify the new domain bucket from the new domain name.
+2. Call the Portal API to create a new `default` technical asset pointing to `s3://{new-domain-bucket}/{namespace}/`. Wait for the `technical_asset.created` event to confirm the new prefix is provisioned and the IAM role has access.
+3. Copy all data from `s3://{old-domain-bucket}/{namespace}/` to `s3://{new-domain-bucket}/{namespace}/`.
+4. For each output port that had the old technical asset linked:
+   a. Link the new technical asset to that output port.
+   b. For each data product with an approved input port on that output port, grant read access to the new S3 location on their IAM role.
+5. Remove read/write access to the old S3 prefix from `dpp-{namespace}`.
+6. For each consumer data product identified in step 4b, revoke their access to the old S3 location.
+7. Mark the old technical asset as archived via the Portal API.
```
stijndehaes, line 128: I don't know if this is something you should support as a team.

An alternative is that we do not copy the new data, or delete the old prefix. But that that is manual, we could send out an email or something instead with the procedure?

The copy could also take hours or days.

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+
+> **Warning**: This sequence involves multiple Portal API calls that each trigger further events. Design all handlers to be idempotent.
```
stijndehaes: You might need to put your events in a queue, so that new events for this data product can't be handled while there are ongoing changes.

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+The provisioner needs the data product's namespace, its settings, the full S3 configuration of each technical asset, and — critically — for each technical asset, the list of consumer data products that had approved read access via any output port exposing that asset. This consumer list is gone from the database by the time the event fires, so it must be available at event dispatch time.
+
+1. For each technical asset in the payload:
+   a. Revoke read access to this TA's S3 location from the IAM role of every consumer data product that had approved access.
+   b. If `read-only-role` is enabled on any of those consumers, also revoke from `dpp-{consumer-namespace}-ro`.
+2. If the `read-only-role` setting was enabled on this data product: delete `dpp-{namespace}-ro`.
+3. Delete the IAM role `dpp-{namespace}`.
+4. Archive or delete S3 objects under `s3://{domain-bucket}/{namespace}/`.
+
+> **Recovery note**: Restoring a deleted data product requires both S3 object recovery (possible via bucket versioning) and database recovery. Enable versioning on all domain buckets. There is no automated restore path — restoring the data product record requires a database recovery operation.
```
stijndehaes:
```suggestion
> **Recovery note**: Restoring a deleted data product requires both S3 object recovery (possible via bucket versioning). Or moving it to a trash folder on s3 that cleans up files after 30 days.
There is no automated restore path — restoring the data product record requires a manual s3 copy or restoring of versioned files
```

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+1. Create IAM role `dpp-{namespace}-ro`.
+2. Grant read-only access to all S3 locations of the data product's own technical assets (`s3:GetObject`, `s3:ListBucket` on each `s3://{bucket}/{prefix}/*`).
+3. Grant read-only access to all S3 locations of every approved input port (i.e. all TAs the data product already has approved access to via its input ports).
+4. Apply the same trust policy as the main role so all current team members can also assume the read-only role.
+
+**`read-only-role` disabled:**
+
+1. Delete `dpp-{namespace}-ro`.
+
+> Warn any external consumers (e.g. BI tools) that were granted access to this role before deleting it.
```
stijndehaes: Should we take into account how to warn them? We can send out an email to users who have access?

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+2. Remove the principal from the trust policy of `dpp-{namespace}`.
+3. If `read-only-role` is enabled: also remove the principal from the trust policy of `dpp-{namespace}-ro`.
+
+---
+
+### Team Member Updated (role changed)
+
+Since both `owner` and `developer` have integration access, a role change does not require any change to the IAM trust policy. No action required.
+
+> **Note**: If future roles are introduced without integration access, the provisioner would need the user's email, their old role, and their new role to decide whether to add or remove the principal from the trust policy. The old role is not available when the update event fires — it would need to be tracked externally or included alongside the new role.
```
stijndehaes:
```suggestion
> **Note**: If future roles are introduced without integration access, the provisioner would need the user's email, and their new role. They can then read the roles of the data product to decided whether to add or remove the principal from the trust policy. 
The old role is not needed since the provisioner can just check all roles
```

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+
+Only type and domain changes have infrastructure impact.
+
+**Type changed:**
+
+1. Update the IAM role `dpp-{namespace}` to reflect the permission set associated with the new type.
+
+**Domain changed:**
+
+This is the most disruptive change. Proceed carefully:
```
stijndehaes, line 118:
```suggestion
This is the most disruptive change. To figure out we are in this case, we need to check the bucket of the new domain, if there is no folder for our current data product we know we need to check all other domains to see if our data product has a folder there. If they do we know we need to migrate it
```

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+
+No action required. The link was PENDING; no access was ever granted.
+
+---
+
+### Input Port Removed
+
+Fires when a data product's access to an output port is removed. This can happen to both approved and pending links.
+
+The provisioner needs the requesting data product's namespace and settings, the link status at the time of removal (APPROVED or PENDING), and the S3 configuration of the technical assets behind the output port. Without the status the provisioner cannot tell whether access was ever granted and therefore whether there is anything to revoke.
```
stijndehaes:
```suggestion
The provisioner needs the requesting data product's namespace and settings, the link status at the time of removal (APPROVED or PENDING), and the S3 configuration of the technical assets behind the output port. Without the status the provisioner would need to check the AWS to tell whether access was ever granted and therefore whether there is anything to revoke.
```

docs/adr/reference/use-case-aws-s3-provisioner.md
stijndehaes: I do believe there are use cases where that makes sense indeed, but ok for me to just make things a bit easier and ignore it 😄

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+### Output Port Deleted
+
+> **Important**: Technical assets are owned by the data product, not by the output port. Deleting an output port does **not** delete the technical assets. It removes the links between the output port and its TAs. The data product retains all its technical assets.
+
+The provisioner needs the S3 configuration of the technical assets that were linked to this output port, and the list of consumer data products (with their namespace and settings) that had approved input port access. Both are gone from the database after deletion, so they must be available at event dispatch time.
+
+1. For each consumer data product with an approved input port on this output port:
+   a. For each TA that was linked to this output port, check whether the consumer still has access to the same TA via another approved input port.
+   b. For any TA no longer accessible via another path: revoke read access from `dpp-{consumer-namespace}`.
+   c. If `read-only-role` is enabled on the consumer: also revoke from `dpp-{consumer-namespace}-ro`.
```
stijndehaes, line 367: Maybe we should just not allow users to delete output ports without first removing all consumers?

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+## Technical Asset Events
+
+### Technical Asset Created
+
+The provisioner needs the owning data product's namespace, settings, the technical mapping (default or custom), and the S3 configuration (bucket, prefix or location).
+
+**Default (portal-managed subfolder):**
+
+1. Create the S3 prefix (upload a placeholder object if needed).
+2. Grant read/write access to `s3://{bucket}/{prefix}/*` on `dpp-{namespace}`.
```
stijndehaes: What is the relation between this and the `s3://{domain}/{namespace}` location we created before? 

Maybe we don't need that standard space for data products?

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+---
+
+## Information Requirements Summary
+
+This table captures information the provisioner needs that is not trivially available at event dispatch time. It is intended as input for event design.
+
+| Event | Information needed | Why |
+|---|---|---|
+| `data_product.created` | Initial team members with their roles | Trust policy must be populated at creation, not via subsequent events |
+| `data_product.deleted` | Per-TA: list of consumers with approved access | Object is already gone; provisioner must revoke consumer access |
```
stijndehaes, line 492: We maybe should just require that first all consumers must be removed? That you can't delete a data product with consumers, I think this makes sense

docs/adr/reference/use-case-aws-s3-provisioner.md
```diff
@@ -0,0 +1,504 @@
...
+| `data_product.input_port_linked` | Link status (APPROVED / PENDING) | Provisioner must distinguish unrestricted (act now) from restricted (wait) |
+| `data_product.input_port_unlinked` | Link status at removal time | Provisioner must know whether access was ever granted |
```
stijndehaes, line 494:
```suggestion
| `data_product.input_port_unlinked` | Link status at removal time | Provisioner must know whether access was ever granted or read the status of AWS |
```

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+
+Finalizers (Option 4) are conceptually clean — Kubernetes uses the same pattern to give controllers a chance to clean up before an object is removed — but they invert the operational model. Portal deletes become asynchronous and block on external systems: UI affordances like "delete this data product" no longer reflect an immediate state change, partial failures are easy (one provisioner stuck, the object stuck with it), and the portal must take on finalizer registration, retries, timeouts, abandonment policies, and dead-letter handling. Beyond the delete case, non-delete events would still need SDK round-trips — the finalizer trick only buys delete semantics. Moving from fire-and-forget events to two-phase commit deletes is a significantly larger architectural step than this problem warrants. We may revisit this if we later need strong delete-ordering guarantees across multiple provisioners.
+
+Dedicated models give provisioners a stable, right-sized contract that survives internal refactors, keeps deletes fire-and-forget, and lets each event carry exactly the state needed for the provisioner's branching logic.
```
stijndehaes: I don't thinks this is a good idea, the problem is that when a data product is deleted, you now need to pass:
- a list of output ports and the list of their consumsers
- a list of technical assets

Basically everything that is attached to a data product. We might not know up front what is needed.
By using either soft deletes or finalizers this is not an issue, the data is still there and can be queried, right up until the moment it's not needed.

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+
+Finalizers (Option 4) are conceptually clean — Kubernetes uses the same pattern to give controllers a chance to clean up before an object is removed — but they invert the operational model. Portal deletes become asynchronous and block on external systems: UI affordances like "delete this data product" no longer reflect an immediate state change, partial failures are easy (one provisioner stuck, the object stuck with it), and the portal must take on finalizer registration, retries, timeouts, abandonment policies, and dead-letter handling. Beyond the delete case, non-delete events would still need SDK round-trips — the finalizer trick only buys delete semantics. Moving from fire-and-forget events to two-phase commit deletes is a significantly larger architectural step than this problem warrants. We may revisit this if we later need strong delete-ordering guarantees across multiple provisioners.
```
stijndehaes:
```suggestion
Finalizers (Option 4) are conceptually clean — Kubernetes uses the same pattern to give controllers a chance to clean up before an object is removed — but they invert the operational model. Portal deletes become asynchronous and block on external systems: UI affordances like "delete this data product" no longer reflect an immediate state change, partial failures are easy (one provisioner stuck, the object stuck with it). Beyond the delete case, non-delete events would still need SDK round-trips — the finalizer trick only buys delete semantics. Moving from fire-and-forget events to two-phase commit deletes is a significantly larger architectural step than this problem warrants.
```

There are no issues with finalizer registration, retries, timeouts etc. The finalizers are just a list of strings on an object that you manipulate through the API.

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+- **Bad, because** delete events are still broken for cascade-deleted children (consumers of a TA, TAs of a deleted output port).
+
+### Option 3: Dedicated per-event payload models
+
+- **Good, because** event schema is decoupled from the API layer — internal refactors do not break consumers.
+- **Good, because** delete events can embed full state at dispatch time, solving teardown without soft deletes or finalizers.
+- **Good, because** `output_port.link_approved` and `technical_asset.linked` can embed both sides (requesting DP + output port + relevant TAs + existing consumers), eliminating round-trips on the hot path.
+- **Good, because** branching state (link status APPROVED vs PENDING, TA status transitions) can be carried explicitly on the event.
+- **Neutral, because** requires upfront scoping of each event (see Appendix A) and a worked use case to validate the scope.
+- **Bad, because** two model hierarchies (API responses + event payloads) must be maintained in parallel; care is needed to keep them from drifting back into coupling.
```
stijndehaes, line 109:
```suggestion
- **Bad, because** two model hierarchies (API responses + event payloads) must be maintained in parallel; care is needed to keep them from drifting back into coupling.
- 
- **Bad, because** Payloads might still not include what you need in a provisioner, it is very hard to product. So you might still need to use the SDK to fetch extra information. Which is possible for update, but not for delete
```

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Option 4: ID-only events + finalizers (Kubernetes-style)
+
+- **Good, because** the event contract stays tiny — only IDs travel on the wire.
+- **Good, because** the delete-cascade problem dissolves cleanly: deletes are explicitly blocked until every provisioner signs off, so nothing is gone before teardown finishes.
+- **Good, because** the pattern is well understood from Kubernetes; provisioners that already think in terms of controllers map onto it directly.
+- **Bad, because** portal deletes become asynchronous and block on external systems, changing the UX contract of every delete affordance in the product.
+- **Bad, because** the portal must take ownership of finalizer registration, retries, timeouts, abandonment, and dead-letter handling — significant new infrastructure for one half of the problem.
```
stijndehaes:
```suggestion
```

This is not an issue, finalizer registration is just an API call, and there are no retries, timeouts etc to be handled, unless I misunderstand stomething

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+- **Bad, because** non-delete events (the majority) still need SDK round-trips; the finalizer trick only addresses delete semantics.
```
stijndehaes:
```suggestion
- **Bad, because** events still need SDK round-trips; the finalizer trick only addresses delete semantics.
```

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+- **Bad, because** moving from fire-and-forget events to two-phase-commit deletes is a much larger architectural shift than this ADR is scoped to make.
```
stijndehaes, line 119:
```suggestion
```

I think you are overestimating the change? But I might be wrong

stijndehaes (changes requested):

docs/adr/0019-event-payload-scoping.md
```diff
@@ -0,0 +1,171 @@
...
+- Event payloads should not grow stale when API response models are refactored.
+- Payloads should be minimal — include what provisioners act on, not display metadata (`about`, `usage`, free-text descriptions).
+
+## Considered Options
+
+- **Option 1: ID-only events + soft deletes** — events carry only entity IDs; deleted objects stay queryable via a `deleted_at` flag; provisioners hydrate via the SDK.
+- **Option 2: Full API response objects embedded** — continue embedding `GetDataProductResponse` and friends directly (current approach).
+- **Option 3: Dedicated per-event payload models** — define a small set of purpose-built payload models independent of the API response layer.
+- **Option 4: ID-only events + finalizers (Kubernetes-style)** — events carry only IDs; deletes go through a finalizer protocol where the portal marks an object for deletion, waits for every registered provisioner to run teardown and remove its finalizer, and only then completes the delete.
+
```
stijndehaes, line 31:
```suggestion
- **Option 5**: To facilitate easy deletes or migrations for domain for example, we could guide people to keep the state of their latest "apply". For example if the provisioner created an s3 bucket or location for a data product, it should keep that in a state object (dynamodb or ssm parameter), that way when the domain changes you can check the state object, see the name of the bucket and see it needs changing. This also helps with a delete since you know keep track of every infra component for an object
- **Option 6**: to facicilate deletes a delete product should also trigger delete events for all Technical assets and Output ports and links. 

```

stijndehaes:
I have a couple of cases we also need to take into account:
- What is the privisioner is offline for a time, and missed events
- What if there is a bug in the provisioner, and you need to change something in all provisioned IAM roles

The way that a k8s operator handles this is basically the same for both, when your operator comes online (after downtime, or after a new deploy) it will get a list of all resources it manages and triggers an update event for all of them.
That way your control loop triggers and does all the updates needed, this will ensure your control loops gets triggered for any missing events, or just for every resource and you can fix the bug in your code.

Maybe we should do something like k8s, or we figure out another way.

Other options are:
- Some retry or DLQ mechanism for webhooks
- Some reply mechanism, that sends update events for every resource from our side with latest state
- We could also guide our users to have 2 mechanisms: webhook mechanism for real time event listening, batch mechanism for reconciliation this could work by using list data product, list output ports etc to trigger reconciliation


Probably we need a combination

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: ?

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: You suggest we scan all buckets for the previous location instead of getting the old value?

docs/adr/reference/use-case-aws-s3-provisioner.md
stijndehaes: You do not need the old domain, you can always figure this out.

If the s3 path you expect does not exist on the location you can check the other domains to see if they had that s3 path, and kick of the migration.

And alternative is that we always send the before state of your whole data product.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: ?  `and see we already have access.` Not clear what that means. Clarify

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: This will need to happen at some customers regardeless so I rather have it documented.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: I would add a warning on the challenges rather then listing a solution.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: I made the mistake of solutiong. We should not do this in this document. Why remove database recovery?

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Part of Portal or Provisioner? I had the feeling Katarina wants to add these things into Portal natively.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Indeed, would be similar to a DP anyway.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Maybe, maybe not. Would leave it documented for now until we support this.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: I made a design decision for the platform. You get a default space. You can register subfolders in case that is what you want to share instead of the root folder.  I think stap 2 needs to be removed as we should already have this kind of access. 

This was done deliberate as our customers platforms aren't perfect either. I think the design still holds.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Why remove linked?

docs/adr/reference/use-case-aws-s3-provisioner.md
vlieven: The challenge here will be more eventual consistency than just idempotency no?

docs/adr/0019-event-payload-scoping.md
vlieven: Or a product cannot be deleted before all assets that are linked to it are removed first?

docs/adr/0019-event-payload-scoping.md
pascal-knapen: I think you can argue for each one of these solutions tbh.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: and see if the policy is already up-to-date.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Not to self: Update remark that we should think about concurrent updates.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Remove this.

docs/adr/reference/use-case-aws-s3-provisioner.md
pascal-knapen: Needs to be domein namespace

stijndehaes (approved):

PR #3541, opened 2026-06-01
Chore/downgrade fastmcp

stijn-janssens:

I don't know how this passes the nightlies


stijndehaes:
@stijn-janssens why do you want to downgrade?

stijndehaes:
A problem with poetry can be solved by:
```
poetry env remove --all
poetry install --with test
```

PR #3543, opened 2026-06-03
Added oauth proxy for MCP

stijndehaes:
The Oauth proxy is needed to ensure that oauth
providers that don't provider support for Dynamic
Client Registration (DCR) work with Portal MCP.

Currently this is made the default since most
Ouath providers do no support it.

Confirmed to work with:
- VSCode
- Github copilot, both the app and jetbrains plugin
- Cursor


vlieven (approved):

stijn-janssens (approved):
If cognito redirects need to be adjusted please add it as a breaking change in the release notes

PR #3571, opened 2026-06-05
Remove unused poethepoet

stijndehaes:


vlieven:
This is not exactly unused (I use it for instance). It's fine for me to replace the commands with a taskfile for more consistency, but then the corresponding taskfile should be added :)

PR #3588, opened 2026-06-09
Add ORM event tracking mixin with exploration events

stijn-janssens:


backend/app/core/webhooks/events.py
```diff
@@ -160,6 +161,45 @@ def event_type(cls) -> str:
...
+    exploration: ExplorationPayload
+
+
+class ExplorationUpdatedEvent(V2Event):
+    @classmethod
+    def event_type(cls) -> str:
+        return "exploration.updated"
+
+    old: ExplorationPayload
+    new: ExplorationPayload
```
vlieven:
```suggestion
    before: ExplorationPayload
    after: ExplorationPayload
```

backend/app/core/webhooks/events.py
```diff
@@ -160,6 +161,45 @@ def event_type(cls) -> str:
...
+    domain_id: UUID
+    owner_id: UUID
+
+
+class ExplorationCreatedEvent(V2Event):
+    @classmethod
+    def event_type(cls) -> str:
+        return "exploration.created"
+
+    exploration: ExplorationPayload
```
vlieven:
```suggestion
    after: ExplorationPayload
```

backend/app/core/webhooks/events.py
```diff
@@ -160,6 +161,45 @@ def event_type(cls) -> str:
...
+    old: ExplorationPayload
+    new: ExplorationPayload
+
+
+class ExplorationDeletedEvent(V2Event):
+    @classmethod
+    def event_type(cls) -> str:
+        return "exploration.deleted"
+
+    exploration: ExplorationPayload
```
vlieven:
```suggestion
    before: ExplorationPayload
```

backend/app/core/context.py
```diff
@@ -0,0 +1,22 @@
+from contextvars import ContextVar
+
+from pydantic import BaseModel
+
+# None = no active request context; events queued outside a request are silently dropped.
```
vlieven: When would events be queued outside a request? This sounds like an error/bug, should this be silently dropped instead of raising an error so we can fix it?

stijn-janssens:
Some potential issues with this approach:
- If a database change happens but after the transaction is committed the request fails, then the events are dropped and no one is notified.
- Same (ofcourse) if the webhook delivery fails. But that is expected.
Should we do the filtering on < 4xx events? Normally on status > 4xx nothing should happen in the DB, so I think it's fine.
Solution could be the outbox pattern but I think we stored that for later?

backend/app/database/event_mixin.py
```diff
@@ -0,0 +1,72 @@
...
+    create_event_class: type | None = None
+    update_event_class: type | None = None
+    delete_event_class: type | None = None
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        sql_event.listen(cls, "after_insert", _track_insert)
+        sql_event.listen(cls, "after_update", _track_update)
+        sql_event.listen(cls, "after_delete", _track_delete)
+
```
vlieven:
```suggestion
        class MyModel(Base, EventTrackedMixin, create_event=MyCreatedEvent, update_event=MyUpdatedEvent, delete_event=MyDeletedEvent):

            def to_event(self) -> MyPayload:
                # Only access scalar columns — no relationship traversal.
                return MyPayload(id=self.id, name=self.name, ...)

    Events are flushed to the webhook after a successful HTTP response by
    the ``dispatch_queued_events`` middleware in ``app/main.py``.
    """

    create_event_class: type | None = None
    update_event_class: type | None = None
    delete_event_class: type | None = None

    def __init_subclass__(cls, create_event=None, update_event=None, delete_event=None, **kwargs):
        super().__init_subclass__(**kwargs)
        
        if create_event: cls.create_event_class = create_event
        if update_event: cls.update_event_class = update_event
        if delete_event: cls.delete_event_class = delete_event
        
        sql_event.listen(cls, "after_insert", _track_insert)
        sql_event.listen(cls, "after_update", _track_update)
        sql_event.listen(cls, "after_delete", _track_delete)

```

backend/app/database/event_mixin.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _track_delete(mapper, connection, target) -> None:
+    cls = getattr(target, "delete_event_class", None)
+    if cls is None:
+        return
+    from app.core.context import queue_event
+
+    payload = target.to_event()
+    field_name = next(iter(cls.model_fields))
```
vlieven: These can also be static methods inside the mixin, that might make it more clear it's a unit

backend/app/database/event_mixin.py
```diff
@@ -0,0 +1,72 @@
...
+                # Only access scalar columns — no relationship traversal.
+                return MyPayload(id=self.id, name=self.name, ...)
+
+    Events are flushed to the webhook after a successful HTTP response by
+    the ``dispatch_queued_events`` middleware in ``app/main.py``.
+    """
+
+    create_event_class: type | None = None
+    update_event_class: type | None = None
+    delete_event_class: type | None = None
```
vlieven: Are there cases where we want to not have these events? Or do we want to enforce that all three are added when the Mixin is used?

backend/app/main.py
```diff
@@ -161,6 +163,22 @@ async def send_response_to_webhook(
...
+    try:
+        response = await call_next(request)
+    finally:
+        events = pop_events()
+        _pending_events.reset(token)
+    if response.status_code < 400 and settings.WEBHOOK_V2_URL:
+        for event in events:
+            asyncio.create_task(
+                call_v2_webhook(type(event).event_type(), event.model_dump(mode="json"))
+            )
```
vlieven: Do we want a task per event? or a single task that emits all the events in a loop?

backend/app/core/context.py
```diff
@@ -0,0 +1,22 @@
...
+    lst = _pending_events.get()
+    if lst is not None:
+        lst.append(event)
+
+
+def pop_events() -> list[BaseModel]:
+    lst = _pending_events.get()
+    if lst is None:
+        return []
+    _pending_events.set([])
```
vlieven: Should this be an empty list, or None?

```suggestion
    _pending_events.set(None)
```

vlieven (commented):
Mostly comments on style/structure and silently failing vs. explicitly failing when the setup is not what we expect

vlieven:
Discussed during call: we do need to take ordering / late arriving events into account somehow/somewhere. Probably not in this PR

backend/app/main.py
```diff
@@ -165,20 +165,22 @@ async def send_response_to_webhook(
...
         events = pop_events()
-        _pending_events.reset(token)
-    if response.status_code < 400 and settings.WEBHOOK_V2_URL:
-        for event in events:
-            asyncio.create_task(
-                call_v2_webhook(type(event).event_type(), event.model_dump(mode="json"))
-            )
+        close_event_context(token)
+    if response.status_code < 400 and settings.WEBHOOK_V2_URL and events:
+        asyncio.create_task(_emit_all_events(events))
```
vlieven, line 175: Should this not use [BackgroundTasks](https://fastapi.tiangolo.com/tutorial/background-tasks/#using-backgroundtasks) instead of asyncio directly?

backend/app/main.py
```diff
@@ -165,20 +165,22 @@ async def send_response_to_webhook(
...
     return response
 
 
+async def _emit_all_events(events: list) -> None:
+    for event in events:
+        await call_v2_webhook(type(event).event_type(), event.model_dump(mode="json"))
```
vlieven, line 184: This is now essentially a sequential loop (you await for each call to finish before sending the next event). Good for keeping sequential order, bad for speed, maybe also good for not DoSing the receiver.
For now, I'd put a comment that this has sequential behavior, so it is documented if this ever is an issue.

Note: this can be fixed by using `asyncio.gather`, but then everything runs concurrently, and we lose ordering. To be tackled at some point in the future

stijn-janssens: Well, we can't really guarantee ordering now either right. As mentioned above. However, by using gather we also lose ordering within a single request, so that makes the problem worse or more frequently occurring, and definitely something we should solve. While for now I think the ordering issue is probably a rare edge case? But good comment, I shall document it

backend/app/main.py
stijn-janssens: I don't think so, BackgroundTasks are used in route handlers. This is middleware. Both the previous webhook middleware as the lifespan is using asyncio tasks directly. BackgroundTasks is not available in middleware

vlieven (approved):

PR #3611, opened 2026-06-12
Input port approved action and total consumption daily push

stijn-janssens:

### DIFF
--- a/backend/app/abstract_data_product/service.py
+++ b/backend/app/abstract_data_product/service.py
@@ -102,7 +102,7 @@ class AbstractDataProductService:
                     "data_product_id": str(output_port.data_product_id),
                     "output_port_id": str(output_port_id),
                     "consuming_data_product_id": str(adp.id),
-                    "type": str(adp.abstract_data_product_type),
+                    "type": str(adp.abstract_data_product_type.value),
                 },
             )
 
--- a/backend/app/abstract_data_product/service.py
+++ b/backend/app/abstract_data_product/service.py
@@ -17,6 +17,7 @@ from app.authorization.role_assignments.output_port.service import (
     RoleAssignmentService as OutputPortRoleAssignmentService,
 )
 from app.core.authz import Action
+from app.core.logging.posthog_analytics import get_posthog_client
 from app.data_products import email
 from app.data_products.output_ports.enums import OutputPortAccessType
 from app.data_products.output_ports.input_ports.model import (
@@ -31,6 +32,7 @@ from app.users.model import User
 class AbstractDataProductService:
     def __init__(self, db: Session):
         self.db = db
+        self.posthog = get_posthog_client()
 
     def get_input_ports(self, data_product_id: UUID) -> Sequence[InputPortModel]:
         ensure_abstract_data_product_exists(data_product_id, self.db)
@@ -91,6 +93,18 @@ class AbstractDataProductService:
             if output_port.access_type != OutputPortAccessType.UNRESTRICTED
             else DecisionStatus.APPROVED
         )
+
+        if approval_status == DecisionStatus.APPROVED and self.posthog:
+            self.posthog.capture(
+                distinct_id=actor.id,
+                event="Input Port Approved",
+                properties={
+                    "data_product_id": str(output_port.data_product_id),
+                    "output_port_id": str(output_port_id),
+                    "consuming_data_product_id": str(adp.id),
+                    "type": str(adp.abstract_data_product_type),
+                },
+            )
 
         dataset_link = InputPortModel(
             dataset_id=output_port_id,
--- a/backend/app/core/logging/posthog_analytics.py
+++ b/backend/app/core/logging/posthog_analytics.py
@@ -1,7 +1,16 @@
+import asyncio
+from datetime import datetime, timedelta
 from typing import Optional
 
+import pytz
 from posthog import Posthog
+from sqlalchemy import func, select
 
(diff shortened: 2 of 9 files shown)


backend/app/abstract_data_product/service.py
```diff
@@ -31,6 +32,7 @@
 class AbstractDataProductService:
     def __init__(self, db: Session):
         self.db = db
+        self.posthog = get_posthog_client()
```
stijndehaes, line 35: Shouldn't this be passed as a dependency like db?

backend/app/abstract_data_product/service.py
```diff
@@ -92,6 +94,18 @@ def _add_single_input_port(
             else DecisionStatus.APPROVED
         )
 
+        if approval_status == DecisionStatus.APPROVED and self.posthog:
```
stijndehaes, line 97: Instead of checking if posthog is not none we could also have a dummy or blackhole implementation that just does nothing.
That way we don't have to check, might be less error prone?

But maybe not worth the effort for now? Let me know what you think

backend/app/core/logging/posthog_analytics.py
```diff
@@ -1,15 +1,118 @@
...
 
+from app.abstract_data_product.model import AbstractDataProduct, AbstractDataProductType
+from app.authorization.role_assignments.enums import DecisionStatus
+from app.core.logging import logger
+from app.data_products.output_ports.input_ports.model import InputPort as InputPortModel
+from app.database.database import SessionLocal
 from app.settings import settings
 
-posthog_client = None
+_ADP_TYPE_DISPLAY_NAME: dict[AbstractDataProductType, str] = {
```
stijndehaes: We can just use the value of the enum instead, the __str__ returns the exact same as this map

backend/app/core/logging/posthog_analytics.py
```diff
@@ -1,15 +1,118 @@
...
+
+    Consumption (approved input ports) is a key success metric for the portal —
+    it measures how actively data products are being used by other teams.
+    """
+    while True:
+        await asyncio.sleep(_seconds_until_next_midnight_utc())
+        try:
+            posthog = get_posthog_client()
+            if not posthog:
+                continue
```
stijndehaes:
```suggestion
                return
```

No need to keep this thread running I guess :)

backend/app/core/logging/posthog_analytics.py
```diff
@@ -1,15 +1,118 @@
...
+
+def _seconds_until_next_midnight_utc() -> float:
+    now = datetime.now(tz=pytz.utc)
+    next_midnight = (now + timedelta(days=1)).replace(
+        hour=0, minute=0, second=0, microsecond=0
+    )
+    return (next_midnight - now).total_seconds()
+
+
+async def report_consumption_metrics_task() -> None:
```
stijndehaes, line 60: One problem might arise from this, if people run multiple containers for an installation we get the same number twice. But I guess that is not really an issue in Posthog?

backend/app/core/logging/posthog_analytics.py
```diff
@@ -1,15 +1,118 @@
...
     )
 
 
-def get_posthog_client() -> Optional[Posthog]:
+def get_posthog_client() -> Optional[PortalPosthog]:
     return posthog_client
+
+
+def _seconds_until_next_midnight_utc() -> float:
+    now = datetime.now(tz=pytz.utc)
```
stijndehaes, line 53: Really smart to only use datetime.now() once in this calculation, using it more then once would have resulted in issues :)

backend/app/core/logging/posthog_analytics.py
```diff
@@ -1,15 +1,118 @@
...
+                distinct_id="system",
+                event="Daily Consumption Metrics",
+                properties={
+                    "total_approved_input_ports": total,
+                    **{
+                        f"approved_input_ports_{adp_type}": count
+                        for adp_type, count in counts_by_type.items()
+                    },
+                },
+            )
```
stijndehaes, line 108:
```suggestion
            )
            await asyncio.sleep(1)
```

I would add a simple sleep here just to ensure that when this calculation is really short that we don't accidentily retriger it.

backend/tests/factories/__init__.py
```diff
@@ -28,6 +31,19 @@
 from .theme_settings import ThemeSettingsFactory
 from .user import UserFactory
 
+_fakes_with_unique = [data_product_fake, dataset_fake, exploration_fake]
+
+
+def reset_unique_fakers() -> None:
```
stijndehaes, line 37: Isn't this a problem that you might get data products with the same name in the db then?

stijndehaes (changes requested):

backend/tests/factories/__init__.py
stijn-janssens: so the reset is only called when we clear the full db, so I think that should not be a problem.
The reason why I had to add the reset is because I added extra tests which actually depleted the unique words that are available, since we used uniqueness across the full test suite

backend/app/abstract_data_product/service.py
stijn-janssens: What would be the advantage? You'd need to remember to inject the dependency in each route instead of having posthog by default in the service if needed?

stijndehaes: Currently we dont use dependencies properly so no advantage, will do so later

stijndehaes (approved):

PR #3648, opened 2026-06-17
Feature/deletion finalizers

stijn-janssens:
Added the option to work with a list of finalizers (strings) that can block deletion


backend/app/abstract_data_product/background_tasks.py
```diff
@@ -0,0 +1,48 @@
...
+from app.abstract_data_product.model import AbstractDataProduct
+from app.core.logging import logger
+from app.data_products.status import DataProductStatus
+from app.database.database import SessionLocal
+
+CHECK_INTERVAL_SECONDS = 600  # run every 10 minutes
+STUCK_THRESHOLD_SECONDS = 3600  # warn after 1 hour in DELETING
+
+
+async def check_stuck_deletions() -> None:
```
stijndehaes, line 15: I would be nice if we could surface this to admins somehow. But something for the future this is good enough :)

backend/app/abstract_data_product/model.py
```diff
@@ -42,6 +43,18 @@ class AbstractDataProduct(Base, BaseORM):
         nullable=False,
     )
 
+    status: Mapped[DataProductStatus] = mapped_column(
+        SAEnum(
+            DataProductStatus,
```
stijndehaes: Should we rename this to something else? AbstractDataProductStatus?

backend/app/abstract_data_product/service.py
```diff
@@ -194,3 +206,49 @@ def send_input_port_requested_emails_to_output_port_owners(
...
+        """Add a finalizer to the abstract data product.
+
+        Finalizers block deletion until they are all removed.
+        """
+        adp = ensure_abstract_data_product_exists(id, self.db)
+        if finalizer in (adp.finalizers or []):
+            raise HTTPException(
+                status_code=status.HTTP_409_CONFLICT,
+                detail=f"Finalizer '{finalizer}' already exists",
+            )
```
stijndehaes: We could also just do nothing instead, now we have to catch this error on the provisioner side and ignore it. What is your preference?

backend/app/abstract_data_product/service.py
```diff
@@ -194,3 +206,49 @@ def send_input_port_requested_emails_to_output_port_owners(
                         approvers=[deepcopy(approver) for approver in approvers],
                     )
                 )
+
+    def add_finalizer(self, id: UUID, finalizer: str) -> AbstractDataProduct:
+        """Add a finalizer to the abstract data product.
+
+        Finalizers block deletion until they are all removed.
+        """
+        adp = ensure_abstract_data_product_exists(id, self.db)
```
stijndehaes, line 215: SHould we check here if the adp is already in deleting state?

stijndehaes (changes requested):
I think we should also either:
- Filter out deleting data product and explorations from list calls 
- Or show that they are in deleting state in the UI

Now users don't really get good feedback. Or I don't see any UI changes that make me think that they do, might have missed it though

backend/app/data_products/router.py
```diff
@@ -204,6 +222,41 @@ def remove_data_product(
     )
 
 
+@router.post(
+    "/{id}/finalizers",
+    dependencies=[
+        Depends(
+            Authorization.enforce(Action.DATA_PRODUCT__DELETE, DataProductResolver)
```
stijndehaes, line 229: Does delete make sense here or should we have seperate auth?

stijndehaes (approved):

PR #3649, opened 2026-06-17
ADR 19 and 20 updated

stijndehaes:

### DIFF
--- a/docs/adr/0019-event-payload-scoping.md
+++ b/docs/adr/0019-event-payload-scoping.md
@@ -31,16 +31,14 @@ A worked example for an AWS S3-backed platform — covering every event in the c
 
 ## Decision Outcome
 
-**Chosen approach:** *Option 3 (dedicated per-event payload models) combined with Option 6 (cascade event emission on delete).*
+**Chosen approach:** *Option 1 (ID-only events + soft deletes).* and *Option 4: ID-only events + finalizers (Kubernetes-style)*
+
+They are honestly very similar and just have small differences, we will take the best of both approaches.
 
 Option 2 is rejected outright: it accidentally solves the delete problem but couples the event contract to internal API shapes and ships a lot of display metadata the provisioner never touches.
-
-Options 1 and 4 are kept on the table as evolution paths, not as the starting point:
-
-- **Option 1** dissolves the delete-cascade problem by leaving objects addressable after their delete event, but requires a cross-cutting `deleted_at` schema change and forces an SDK round-trip on every event — including hot paths like `link_approved` and `technical_asset.linked` where two-sided hydration is exactly what the provisioner needs.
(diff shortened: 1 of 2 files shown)


stijn-janssens (approved):

docs/adr/0009-plugin-data-output-configs.md
```diff
@@ -83,8 +83,7 @@ Having a table schema per plugin allows each integration to map its terminology
 For now:
 * Plugins are Python packages installed into the portal environment
 * Adding a plugin requires:
-
-  * Installing the package (e.g. `pip install`)
+  * Installing the package (e.g. `pip install`) in the
```
vlieven: Do you mean this?

```suggestion
  * Installing the package (e.g. `pip install`) in the portal container image
```

docs/adr/0019-event-payload-scoping.md
```diff
@@ -86,9 +84,12 @@ Today, events are emitted from the router layer. A future refactor will move eve
 
 ### Option 1: ID-only events + soft deletes
 
+We will implement event loops in the SDK that only implement data product and reconciler [see here](./reference/id-only-event-implementation.md).
```
vlieven, line 87:
```suggestion
We will implement [event loops](./reference/id-only-event-implementation.md) in the SDK that only implement Data Product and Exploration reconcilers.
```

docs/adr/0019-event-payload-scoping.md
```diff
@@ -86,9 +84,12 @@ Today, events are emitted from the router layer. A future refactor will move eve
...
+Output port events trigger the data product event loop for example.
```
vlieven, line 88:
```suggestion
Output Port events trigger the Data Product event loop for example.
```

vlieven (commented):
Added some suggestions regarding what I believe are missing words

docs/adr/0009-plugin-data-output-configs.md
stijndehaes: This file shouldn't have changed, I'll revert that

docs/adr/0019-event-payload-scoping.md
```diff
@@ -31,16 +31,14 @@ A worked example for an AWS S3-backed platform — covering every event in the c
...
-- **Option 4** is conceptually clean — Kubernetes uses the same pattern to give controllers a chance to clean up before an object is removed — but it inverts the operational model. Portal deletes become asynchronous and block on external systems: UI affordances like "delete this data product" no longer reflect an immediate state change, partial failures are easy (one provisioner stuck, the object stuck with it). Beyond the delete case, non-delete events would still need SDK round-trips. Moving from fire-and-forget events to two-phase-commit deletes is a significantly larger architectural shift than this problem warrants today.
+They are honestly very similar and just have small differences, we will take the best of both approaches.
 
+Option 2 is rejected outright: it accidentally solves the delete problem but couples the event contract to internal API shapes and ships a lot of display metadata the provisioner never touches.
+Option 3 has been rejected after trying implementation options, it doesn't handle certain edge cases nicely like missing events.
```
pascal-knapen, line 39: Can you make this more concrete "doesn't handle certain cases nicely like missing events" as this feels like the major pivotal point compared to the previous chosen option.

pascal-knapen (dismissed):
Needs some more clarification

docs/adr/0019-event-payload-scoping.md
stijndehaes: These are explained in the bad for Option 3 below, do they make sense?

PR #3650, opened 2026-06-17
Updated implementation for adr 19,20 change

stijndehaes:


sdk/example/provisioner.py
```diff
@@ -0,0 +1,171 @@
...
+        self._workspace = workspace
+
+    def _path_for(self, exploration_id: UUID) -> Path:
+        return self._workspace / str(exploration_id) / "manifest.json"
+
+    async def reconcile(self, exploration_id: UUID):
+        logger.info("Reconciling exploration %s", exploration_id)
+
+        response = await get_exploration.asyncio(id=exploration_id, client=self._client)
+        if response is None:
```
stijn-janssens, line 91: I miss the deletion event here. If we work with finalizers a delete will never result in the response being None. Instead the event deletion will happen, but the id will still return the object, the provisioner should remove the object and the finalizer and then portal will delete it. Resulting in a response None, but the provisioner should not do anything on that event. Or am I misinterpreting something?

sdk/example/provisioner.py
```diff
@@ -0,0 +1,171 @@
...
+                "Exploration %s already absent, nothing to deprovision", exploration_id
+            )
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    client = _build_client()
+    manager = ReconcileManager(
+        ExplorationProvisioner(client),
+        default_delay=30.0,  # debounce: wait 30s before picking an exploration up
```
stijn-janssens, line 147: why so high? in the readme you talked about 2 secs

sdk/example/provisioner.py
```diff
@@ -0,0 +1,171 @@
...
+        await client.get_async_httpx_client().aclose()
+
+
+app = FastAPI(lifespan=lifespan)
+
+
+@app.post("/webhook")
+async def webhook(request: Request) -> dict[str, Any]:
+    """Receive portal CloudEvents and enqueue a reconcile for the exploration."""
+    return await request.app.state.handler.dispatch_routing(request)
```
stijn-janssens, line 166: how and where does the reconciler determine whether it is an exploration or product event that needs to be reconciled? Will we do both a get_exploration and get_product on each ID?

sdk/example/provisioner.py
stijndehaes: You are completely correct, but finalisers don't exist in this branch yet. So can't use it :)

Shall we update with a new PR later?

sdk/example/provisioner.py
stijndehaes: It's just an example, but it kind of depends on how expensive your loop is.
The more expensive it is, the higher I would make this :) 

I propose to add this to user documentation later? I want to write a nice doc later, or maybe a small guide

sdk/example/provisioner.py
stijn-janssens: I am fine with the introduction of finalizers later, but I would expect a reaction to event types here? No?
Just assuming that a None return means a delete and a not none return means a refresh feels a bit strange?

sdk/example/provisioner.py
stijndehaes: So currently we only support an exploration loop, but later when more events are added we will support also data product loop.

Which means that for data product events we call your data product loop.

And for input ports for example we have to know the type the input port is connected to so we can call the correct exploration or data product to be updated. This is WIP in another branch

sdk/example/tests/test_provisioner.py
```diff
@@ -0,0 +1,171 @@
...
+def patch_get_exploration(
+    monkeypatch,
+):
+    """Patch the portal lookup; returns a setter for the value reconcile will see."""
+    state: dict[uuid.UUID, GetExplorationResponse] = {}
+
+    async def fake_get(id, client):  # noqa: A002 - matches generated signature
+        return state.get(id)
+
+    monkeypatch.setattr(prov_module.get_exploration, "asyncio", fake_get)
```
vlieven, line 54: You can also use asyncio in testing. But this can be cleaned up later.
https://pypi.org/project/pytest-asyncio/

Edit: you actually have installed this later, so I don't think this is even needed?

sdk/example/provisioner.py
vlieven: Yes, I agree. Otherwise, what is the point of having the different event types?

backend/app/core/webhooks/events.py
```diff
@@ -38,33 +38,21 @@ def event_type(cls) -> str:
 
 class ExplorationPayload(BaseModel):
```
stijn-janssens, line 39: Shouldn't this just become "EventPayload" ? As every event will just be the ID?

sdk/example/tests/test_provisioner.py
stijndehaes: I think you might mis understand what this does. It replaces the asyncio function in the `get_exploration` python file with our mock here. 

Or maybe I mistunderstand the comment

sdk/example/provisioner.py
stijndehaes: So this deleted event will actually be useless later to be honest, since it will only be called after the exploration is deleted. When the exploration is in deleting phase because of the finaliser you will just get an updated event. And the get exploration will work, and you can check the status and finalisers to see if you have to do work. 
Eventually when none is returned here the loop doesn't need to do anything, so the deleted event type is useless

stijn-janssens (approved):

PR #3662, opened 2026-06-18
Data product reconciler

stijndehaes:
Added support for data product reconciler,
in a future PR we will add events for:
- output port
- technical asset
- ...

That trigger updates on the data product

### DIFF
--- a/backend/app/core/webhooks/events.py
+++ b/backend/app/core/webhooks/events.py
@@ -31,23 +31,23 @@ class V2Event(BaseModel):
         raise NotImplementedError
 
 
-class ExplorationPayload(BaseModel):
+class ExplorationEventPayload(BaseModel):
     id: UUID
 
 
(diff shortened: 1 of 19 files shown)


backend/app/core/webhooks/events.py
```diff
@@ -36,23 +29,45 @@ def event_type(cls) -> str:
...
 
-class ExplorationDeletedEvent(V2Event, ExplorationPayload):
+class ExplorationDeletedEvent(V2Event, ExplorationEventPayload):
     @classmethod
     def event_type(cls) -> str:
         return "exploration.deleted"
+
+
+class DataProductEventPayload(BaseModel):
+    id: UUID
```
stijn-janssens, line 81: I would still advocate for this to be the GenericEventPayload or something like that. This has no real use

stijn-janssens (approved):

PR #3678, opened 2026-06-19
More events for data product reconciler

stijndehaes:
Added events for:
- technical assets
- output ports
- connections between technical assets and output ports


backend/app/data_products/output_port_technical_assets_link/model.py
```diff
@@ -65,3 +67,13 @@ class DataOutputDatasetAssociation(Base, BaseORM):
             "data_output_id", "dataset_id", name="unique_data_output_dataset"
         ),
     )
+
+    def to_event(self) -> OutputPortTechnicalAssetLinkEvent:
+        return OutputPortTechnicalAssetLinkEvent(
+            id=self.id,
+            data_product_id=self.dataset.data_product_id
+            if self.data_output is None
```
stijn-janssens, line 75: when is data output none in this case?

stijn-janssens (approved):

backend/app/data_products/output_port_technical_assets_link/model.py
stijndehaes: to_event is triggered in a after_update, after_insert, after_delete trigger in sql alchemy, when the data_output has not been loaded in memory by sql alchemy it is None, but we are not allowed to fetch it here because we are in these triggers. Normally you would access it and sql alchemy would fetch it. So we just use the dataset which hopefully is loaded. If it isn't this crashes and we have a test to fix :)

PR #3702, opened 2026-06-23
Claude desktop support

stijndehaes:
There was an issue with the static file
serving, where /mcp was routed to static
files insteaf of the mcp server.
By adding the redirect from /mcp to /mcp/
this is fixed


vlieven:
The redirect from /mcp to /mcp/ feels weird, I would expect it to work the other way around. However, if this is the way to do it, fine by me.

vlieven (approved):

PR #3706, opened 2026-06-23
Move agents demo to use the reconciler-style provider

vlieven:
- [ ] Updated sample data so the new feature can be easily tested.


stijn-janssens (approved):
Fine to not have explorations I guess for this demo?

vlieven:
> Fine to not have explorations I guess for this demo?

The demo is not using explorations, this PR simply migrates the provider.
If we want to extend the demo for explorations, the provider will need to be extended too, but that can happen then I believe.

PR #3726, opened 2026-06-26
Add api key auth to sdk

stijn-janssens:

### DIFF
--- a/sdk/sdk/auth.py
+++ b/sdk/sdk/auth.py
@@ -21,6 +21,7 @@ class PortalAuth:
         self.timeout = timeout
         self.auth_mode = os.getenv("PORTAL_AUTH_MODE", "")
         self.base_url = os.getenv("PORTAL_BASE_URL", "").rstrip("/")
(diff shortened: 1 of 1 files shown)


sdk/sdk/auth.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, timeout: int = 10):
         self.timeout = timeout
         self.auth_mode = os.getenv("PORTAL_AUTH_MODE", "")
         self.base_url = os.getenv("PORTAL_BASE_URL", "").rstrip("/")
+        self.api_key = os.getenv("PORTAL_API_KEY", "")
```
vlieven, line 24: Where is this API Key defined? Also, we have a auth_mode flag, that we only check to use "client_credentials". I think the entire auth flow needs revising?

stijn-janssens: I currently just see it as if api key is present then api key. Disregarding the mode.
But fine to change it.
Api key is set in the settings / env file

vlieven (approved):

PR #3727, opened 2026-06-26
Fix authentication for finalizers on explorations

stijn-janssens:

### DIFF
--- a/backend/app/explorations/router.py
+++ b/backend/app/explorations/router.py
@@ -165,21 +165,18 @@ def remove_input_port_from_exploration(
         },
     },
     dependencies=[
-        Depends(Authorization.enforce(Action.GLOBAL__CREATE_EXPLORATION, EmptyResolver))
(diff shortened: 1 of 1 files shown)


vlieven:
What behavior are you trying to correct for? The existing implementation looks more correct than the changes you've proposed?

stijn-janssens:
I was testing out some functionality and the existing authorization does not work for admins. Could be that my solution is also not ideal

stijndehaes:
> I was testing out some functionality and the existing authorization does not work for admins. Could be that my solution is also not ideal

Proposal:
- let's just make a new action for this something similar to data product. This is only used by admin anyway at the moment, and they will have that right. An owner of a data product will probably not write a provision so it should be fine that only admin can do it

stijndehaes:
made this as a replacement, let me know what you think: https://github.com/conveyordata/data-product-portal/pull/4002

PR #3728, opened 2026-06-26
fix: exclude the requester from approver email recipients (#3265)

### LINKED ISSUE #3265: Do not send approvers an email when you are the approver

mvanhorn:
Closes #3265. Resubmit of #3278 (closed for inactivity) with the unit tests requested.

When a user requests access and is themselves an approver, they were emailed their own request. This filters the requesting actor out of the approvers list before sending, and skips the email entirely when no other approvers remain, across the three request paths (input-port access, data-product role assignment, output-port technical asset link).

Adds unit tests for both cases (requester-only -> no email; other approvers -> email excludes requester) on all three paths. Verified locally: 6 passed.

### DIFF
--- a/backend/app/abstract_data_product/service.py
+++ b/backend/app/abstract_data_product/service.py
@@ -198,14 +198,18 @@ class AbstractDataProductService:
                     dataset_link.dataset_id,
                     Action.OUTPUT_PORT__APPROVE_DATAPRODUCT_ACCESS_REQUEST,
                 )
-                background_tasks.add_task(
-                    email.send_dataset_link_email(
-                        dataset_link.consuming_abstract_data_product,
-                        dataset_link.dataset,
-                        requester=deepcopy(actor),
-                        approvers=[deepcopy(approver) for approver in approvers],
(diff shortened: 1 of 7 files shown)


mvanhorn:
Fixed the failing backend test. `test_dataset_link_manual_approval_email_sent` set the requester as the only approver, so the new requester-exclusion logic correctly produced no email (Called 0 times). Added a distinct DATASET-scoped approver to the fixture so a non-requester recipient remains and the email fires - matching how the router resolves approvers via `users_with_authz_action(dataset_id, OUTPUT_PORT__APPROVE_TECHNICAL_ASSET_LINK_REQUEST)`.

stijn-janssens (approved):
Thanks a lot for your contribution @mvanhorn

PR #3753, opened 2026-07-01
Update requirements

stijn-janssens:


vlieven (approved):
What changed that this suddenly works?

stijn-janssens:
Every OPTIONS preflight request returned a 500 Internal Server Error when traces were enabled.
The crash was:
`AttributeError: '_IncludedRouter' object has no attribute 'path'`

It happened inside opentelemetry's `_get_route_details` function. FastAPI uses `_IncludedRouter` nodes to represent routes added via `include_router()` and it got handled badly. Was fixed in a newer release of opentelemetry

Issue #3754, opened 2026-07-01
Time bound access for input ports


stijn-janssens:
## Summary
Time bound access for input ports, together with necessary settings and renewal flows etc etc

### Motivation
Allows for stricter regulation of access

TODO:
- [ ] Update release notes and documentation
- [ ] Check events for provisioner
- [x] get_user_requests and pending actions
- [ ] Should we rename approved input port status to active
- [ ] When a renewal is approved, should the start date of the renewal be the end date of the previous requests, currently it just starts today
- [ ] Implement `current_request` on `InputPort` using a sql query with deferred property instead of using the current methods
- [ ] Remove requested_on from inputportRequest, created_on should be the exact same value and meaning. We can easily make `requested_on` an SQL Alchemy alias of `created_on`
- [ ] Remove the status from the  input_port table in the database and only keep it as a computed. field
- [ ] Currently we use the sample_data.sql seed as a base to work from, but we also use it in tests backend/tests/app/search_output_ports/test_router.py for example to test the embedding efficiency. Is it a good idea to use the same for that, what if we add things/remove things because of new features, ... That doesn't make the test very much idempotent. There is something to say to keep it "real" however, how real is the sample_data.sql anyway
- [x] https://github.com/conveyordata/data-product-portal/pull/3867#discussion_r3603047407
- [ ] https://github.com/conveyordata/data-product-portal/pull/3867/changes/BASE..edd4da9b6651a781bbec4a6d75d484546b126e97#r3603096428 discuss with Katarina if we want this. This allows us further move that an input port could go from approved -> expired -> approved -> revoked -> approved. We keep a single input_port and update it with new requests. Instead of creating new input ports, and lose the history of these requests -> As discussed with katarina we will refactor the remove into revoke. This will no longer delete the intput port, but revoke the current access. As a result a new state is introduced for the access request AND the input_port. A new field will be required to have a revoked at, revoked by. This also changes the state machine. 
- [ ] Don't show input ports that are not reelvant anymore. For example when it is expired already for more then 30 days, don't show it, and add a toggle to show inactive requests. Or only show active requests. Or a history button. In addition when you first havea request and it is pending, and you cancell it, (the only cancelled state) don't show it in the UI this is unneeded and doesn't provide any extra information.
- [ ] Because an input port is never deleted it will always show in the exploration, it is currently impossible to hide that an remove that link.

HannesDS:
[...]

PR #3773, opened 2026-07-03
Feature/3754 admin settings timebound

stijn-janssens:
Added the admin settings for creating time bound access policies.

Part of #3754 

UI
Scherm­afbeelding 2026-07-03 om 12 39 18


backend/app/access_durations/model.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+    access_duration_type: Mapped[AccessDurationType] = mapped_column(
+        SAEnum(
+            AccessDurationType,
+            values_callable=lambda enum: [e.value for e in enum],
+            native_enum=False,
+            validate_strings=True,
+        ),
+        nullable=False,
+    )
```
vlieven, line 37: ~It's ok to use enums for the python types, but we don't use enums in the database. They cannot be mutated after they are created, so if we ever want to add a new data_product_type or access_duration_type, we will be in deep 💩.~

I see the migration does use string types, so this is probably fine. Is there a benefit to having the SA model as an enum?

frontend/src/pages/settings/components/settings-tabs/access-policy-tab/access-policy-tab.component.tsx
```diff
@@ -0,0 +1,9 @@
+import { AccessPolicyForm } from '../components/access-policy-form/access-policy-form.component';
+
+export function AccessPolicyTab() {
+    return (
+        <div>
+            <AccessPolicyForm />
+        </div>
+    );
+}
```
vlieven, line 204: Why is this wrapped in an extra diff?
Why is this an additional component? It's a function directly returning another function..?
Or is this intended to be extended soon?

frontend/src/pages/settings/components/settings-tabs/components/access-policy-form/access-policy-form.component.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    }).unwrap(),
+                ),
+            );
+            dispatchMessage({ content: t('Access duration settings saved successfully'), type: 'success' });
+        } catch (_e) {
+            dispatchMessage({ content: t('Failed to save access duration settings'), type: 'error' });
+        }
+    };
+
+    const columns: TableColumnsType<ConsumerPolicy> = [
```
vlieven: Why does every `<Flex />` component in this list use a different `gap` value?

frontend/src/pages/settings/components/settings-tabs/components/access-policy-form/access-policy-form.component.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    <Flex align="center" gap={12}>
+                        <Checkbox
+                            checked={record.alternativeAllowed}
+                            onChange={(e) => {
+                                const checked = e.target.checked;
+                                const alternativeIsTimeBound = record.durationType === AccessDurationType.Permanent;
+                                updatePolicy(record.key, {
+                                    alternativeAllowed: checked,
+                                    alternativeDays:
+                                        checked && alternativeIsTimeBound ? (record.alternativeDays ?? 30) : null,
```
vlieven: Can you rewrite this line? The type of this is now `false | number | null`?

vlieven (changes requested):

backend/app/access_durations/model.py
stijn-janssens: We used this pattern elsewhere in the code and I just copied it. I have not looked up any benefits :D

frontend/src/pages/settings/components/settings-tabs/components/access-policy-form/access-policy-form.component.tsx
stijn-janssens: consistency 😉

backend/app/access_durations/model.py
stijn-janssens: FYI
Since `native_enum=False` is set, there's no PostgreSQL `ENUM` type created — it's just `VARCHAR` in the DB. So the migration concern is already solved.

The benefit of keeping `SAEnum(native_enum=False)` over plain `String`:

1. **Automatic coercion on read** — SQLAlchemy maps the raw `"data_products"` string back to `AbstractDataProductType.DATA_PRODUCT` automatically, so your service code works with Python enums, not raw strings
2. **`validate_strings=True`** — if something tries to write `"invalid_type"` it's caught at the application layer before hitting the DB
3. **Type-safe Mapped annotations** — `Mapped[AbstractDataProductType]` works correctly with mypy/pyright

The downside is just the verbose `values_callable` lambda. The alternative — `mapped_column(String)` with `Mapped[str]` — gives you raw strings everywhere and requires manual casting. Not worth it.

So the current setup is the right call: `native_enum=False` for DB flexibility, `SAEnum` for Python-side type safety.

vlieven (approved):

PR #3774, opened 2026-07-03
Fix/explorers

stijn-janssens:
All explorers are fully bugged

### DIFF
--- a/frontend/src/components/charts/node-editor/node-editor.module.scss
+++ b/frontend/src/components/charts/node-editor/node-editor.module.scss
@@ -2,5 +2,4 @@
 
 .node-editor {
     flex: 1;
-    height: 100%;
 }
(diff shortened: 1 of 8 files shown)


vlieven:
What does fully bugged mean?
Is it because of an update to xyflow?
Can we write a test to catch this?

stijn-janssens:
fully bugged means the global explorer crashed because of a syntax error? In the latest reactflow, hence the downgrade.
The internal explorers were not rendered because suddenly a height was expected.
Not sure on how you would catch this because I believe the element was available on the page. But i'll ask my good friend AI

frontend/src/components/charts/node-editor/node-editor.module.scss
```diff
@@ -1 +1,6 @@
 @use "@/styles/globals";
+
+.node-editor {
+    flex: 1;
```
stijndehaes, line 4: I don't think we need flex 1 here?

stijndehaes (approved):

PR #3775, opened 2026-07-03
fix: Add empty-state message for data product settings with no fields

HannesDS:

Navigating to the data-products page and opening the settings tab didn't show anything if there are no settings present. Added an Empty component when there are no settings and when the page is not loading. After the page, the tab shows as follows:

image


stijn-janssens:
I would personally just hide the settings tab if no settings are present. Please check if this is the case both for data products and output ports

stijn-janssens:
Changed my mind. Empty state might be better. Because otherwise we could get into issues if we add more things to the settings tab and it doesn't render for some people

stijn-janssens (approved):

PR #3780, opened 2026-07-06
Feature/3754 timebound output port

stijn-janssens:
UI changes only, backend to come later

In support of #3754 

Scherm­afbeelding 2026-07-06 om 10 37 55

Edit screen:
Scherm­afbeelding 2026-07-06 om 11 19 24


frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -57,6 +67,136 @@ import { useGetDatasetOwnerIds } from '@/utils/dataset-user-role.helper.ts';
 import { selectFilterOptionByLabel, selectFilterOptionByLabelAndValue } from '@/utils/form.helper.ts';
 import styles from './dataset-form.module.scss';
 
+const PRODUCT_TYPE_LABELS: Record<string, string> = {
+    [AbstractDataProductType.DataProducts]: 'Data Products',
+    [AbstractDataProductType.Explorations]: 'Explorations',
+};
+
+function AccessDurationSection({ type, rows }: { type: AbstractDataProductType; rows: AccessDuration[] }) {
```
stijndehaes: Normally we always make a props type why not for these sections? Also propose to rename type and rows aws variables.
```suggestion

type AccessDurationSectionProps = {
  abstractProductType: AbstractDataProductType;
  accessDurations: AccessDuration[];
};
function AccessDurationSection({ abstractProductType, accessDurations }: AccessDurationSectionProps) {
```

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -57,6 +67,136 @@ import { useGetDatasetOwnerIds } from '@/utils/dataset-user-role.helper.ts';
...
+            disabled: !hasPermanent,
+        },
+    ];
+
+    return (
+        <Flex vertical gap={8} style={{ marginLeft: 20 }}>
+            <Typography.Text type="secondary">
+                {t('{{type}} Access Duration', { type: PRODUCT_TYPE_LABELS[type] })}
+            </Typography.Text>
+            <Flex vertical gap={12}>
```
stijndehaes:
```suggestion
            <Flex vertical gap={"small"}>
```

Prefer using small, medium etc over raw numbers

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -57,6 +67,136 @@ import { useGetDatasetOwnerIds } from '@/utils/dataset-user-role.helper.ts';
...
+            ) : (
+                t('Permanent')
+            ),
+            value: AccessDurationType.Permanent,
+            disabled: !hasPermanent,
+        },
+    ];
+
+    return (
+        <Flex vertical gap={8} style={{ marginLeft: 20 }}>
```
stijndehaes:
```suggestion
        <Flex vertical gap={"small"}>
```

Prefer this over numbers, I also don't like the marginleft, but I might be wrong, can you add a screenshot?

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -57,6 +67,136 @@ import { useGetDatasetOwnerIds } from '@/utils/dataset-user-role.helper.ts';
...
+                )}
+            </Flex>
+        </Flex>
+    );
+}
+
+function AccessDurationInfo({ mode }: { mode: 'create' | 'edit' }) {
+    const { t } = useTranslation();
+    const { data: allDurations = [] } = useGetAllAccessDurationsQuery();
+    const { data: enabledState } = useIsTimeBoundAccessEnabledQuery();
```
stijndehaes, line 151: Very confusing that the backend does not exist yet, `useIsTimeBoundAccessEnabledQuery` is pretty clear.

But what does `useGetAllAccessDurationsQuery` mean? Can an admin configure the allowed durations? Is this just `timebound` vs `permanent`? If so why do we need the `useIsTimeBoundAccessEnabledQuery`? Because this already says if it's enabled no?

stijndehaes (changes requested):

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
stijn-janssens: The queries that are called in this PR already exist. Only the new functionality (extra values to persist etc) is not yet implented.
useGetAllAccessDurations in this case fetches the state of the global wide platform admin settings for access durations. The enabled is just a general feature flag. In theory they can be unrelated (i.e. later disablement of the feature, in practice they probably aren't)

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
stijn-janssens: Scherm­afbeelding 2026-07-06 om 17 18 10
Agree with small. The margin is used for slight indentation. I like it personally, it's made after the first design. It breaks the monotonicity of the form a little bit and for me helps in focussing my eye, groups this together. But happy to change if not for you

stijndehaes: I do think it's also weird now with the warning being not indented.

Actually now that I think about it, it's rather weird that you have a single form item with 2 choices in it.

Shouldn't every duration selection just be a new form item?

stijndehaes (approved):

PR #3781, opened 2026-07-06
Feature/3754 timebound output port backend

stijn-janssens:
Backend implementation of #3780 


backend/app/data_products/output_ports/schema_request.py
```diff
@@ -23,15 +25,6 @@ class CreateOutputPortRequest(DatasetUpdate):
     owners: Annotated[list[UUID], MinLen(1)]
 
 
-@deprecated("Use CreateOutputPortRequest instead")
```
stijndehaes, line 26: ❤️

backend/app/data_products/output_ports/schema_request.py
```diff
@@ -14,6 +14,8 @@ class DatasetUpdate(ORMModel):
     namespace: str
     description: str
     access_type: OutputPortAccessType
+    data_product_access_duration_type: AccessDurationType
```
stijndehaes, line 17: Should we make the default permanent in the API? That way old SDK or API users aren't broken immediately?

stijndehaes (approved):

backend/app/data_products/output_ports/schema_request.py
stijn-janssens: I find this a (broader) difficult question. At some point I would like to have this required and not work with optional or defaults. Might as well do it now then, no?

Curious to find out your opinion on this

backend/app/access_durations/service.py
```diff
@@ -92,6 +96,37 @@ def upsert_access_duration(
...
         return self.get_access_durations_by_type(abstract_data_product_type)
+
+    def _cascade_output_ports(
+        self,
+        abstract_data_product_type: AbstractDataProductType,
+        allowed_types: set[AccessDurationType],
+        new_default_type: AccessDurationType,
+    ) -> None:
+        """Move output ports off an access duration type no longer offered."""
+        if abstract_data_product_type == AbstractDataProductType.DATA_PRODUCT:
```
stijndehaes: I would replace this with a switch/match statement, but not a big deal

backend/app/access_durations/service.py
```diff
@@ -92,6 +96,37 @@ def upsert_access_duration(
...
+            column = DatasetModel.data_product_access_duration_type
+        elif abstract_data_product_type == AbstractDataProductType.EXPLORATION:
+            column = DatasetModel.exploration_access_duration_type
+        else:
+            return
```
stijndehaes: Should we not raise an exception instead? Things will break anyway if we add another type, but we can make it clearer

frontend/src/components/datasets/dataset-form/dataset-form.component.tsx
```diff
@@ -117,12 +126,18 @@ function AccessDurationSection({ accessDurations }: AccessDurationProps) {
     return (
         <Flex vertical gap={'small'}>
             <Flex vertical gap={'small'}>
-                <Radio.Group
-                    value={selected}
-                    onChange={canToggle ? (e) => setSelected(e.target.value) : undefined}
-                    options={options}
-                    optionType="button"
-                />
+                <Form.Item
```
stijndehaes: We shouldn't be adding a form.item here since `AccessDurationSection` itself is already in a form.item

frontend/src/store/api/services/tags/accessDurationsTags.ts
```diff
@@ -0,0 +1,17 @@
...
+
+type EndpointDefinitions = Parameters<typeof api.enhanceEndpoints>[0]['endpoints'];
+
+export const accessDurationsTags = {
+    getAllAccessDurations: {
+        providesTags: [{ type: TagTypes.AccessDuration, id: STATIC_TAG_ID.LIST }],
+    },
+    updateAccessDuration: {
+        invalidatesTags: () => [
+            { type: TagTypes.OutputPort },
```
stijndehaes, line 12: Should InputPorts also be invalidated?

stijndehaes (changes requested):

frontend/src/store/api/services/tags/accessDurationsTags.ts
stijn-janssens: no, changing the admin access durations only has impact on the output port. They can change from being permanent by default to time bound by default, or no longer allowed to be time bound at all.
Input ports are not impacted, they keep their current way of working as per design

stijndehaes (approved):

PR #3796, opened 2026-07-07
Feature/3754 time bound input port

stijn-janssens:
In support of #3754: Adds expiration dates and automated expiry to input ports
- [x] Updated sample data so the new feat

Scherm­afbeelding 2026-07-07 om 15 46 03
ure can be easily tested.


backend/app/data_products/output_ports/input_ports/service.py
```diff
@@ -102,6 +102,11 @@ def approve_output_port_as_input_port(
                 },
             )
 
+        if current_link.requested_duration_days:
+            current_link.expires_on = datetime.now(tz=pytz.utc) + timedelta(
```
vlieven, line 106: I think you don't need pytz anymore, and it can be imported from the datetime module now
(`from datetime import timezone`, `tz=timezone.utc`)

backend/app/main.py
```diff
@@ -81,11 +84,13 @@ async def lifespan(_: FastAPI):
     device_flow_cleanup_task = asyncio.create_task(cleanup_device_flow_table_task())
     consumption_metrics_task = asyncio.create_task(report_consumption_metrics_task())
     stuck_deletion_task = asyncio.create_task(check_stuck_deletions())
+    expire_input_ports_task = asyncio.create_task(expire_input_ports())
     yield
     admin_task.cancel()
     device_flow_cleanup_task.cancel()
     consumption_metrics_task.cancel()
     stuck_deletion_task.cancel()
+    expire_input_ports_task.cancel()
```
vlieven, line 93: Just FYI: I'd like to do a spike to review this structure. It seems a bit brittle. Also, I think you're running synchronous DB calls behind an async routine, which might block the event loop, depending on how fastapi handles this.

This fix should not be part of the PR though. I'll check with the PO if we want to plan this in. Feel free to ignore all of this in the meantime :)

vlieven (approved):
If you rebase the PR to resolve the merge conflicts and drop the `pytz` dependency, this is good to go for me

PR #3821, opened 2026-07-10
feature: add initial proposal for adr 0021

HannesDS:

Introduce ADR for database schema regarding input_ports and input_port_requests. Raised while developing feature #3754.
It proposes to split the input_ports table into two, input_ports and input_port_requests. More details can be found in the ADR itself.

### DIFF
--- a/docs/adr/0021-input-ports-and-input-port-requests.md
+++ b/docs/adr/0021-input-ports-and-input-port-requests.md
@@ -70,7 +70,9 @@ Keep (identity only):
 | `id` | PK |
 | `dataset_id` → datasets | the output port |
 | `consuming_abstract_data_product_id` → abstract_data_products | the consuming data product / exploration |
+| `created_by` | owner, the one we need to notify for renewals | 
 | `created_on`, `updated_on` | audit |
+
 
 Add:
 
--- a/docs/adr/0021-input-ports-and-input-port-requests.md
+++ b/docs/adr/0021-input-ports-and-input-port-requests.md
@@ -75,7 +75,6 @@ Add:
 | Add | Meaning |
 | --- | --- |
 | link access indicator — reused `status`, or a new `is_active` | how the link signals "has access" for counts / graph / ordering / filtering; derived from its requests, not stored. Two options — see Open Questions. |
-| `UniqueConstraint(dataset_id, consuming_abstract_data_product_id)` | one link per (output port, consumer), so a re-request adds a request to the existing link instead of creating a duplicate link. |
 
 Remove (moved to `input_port_requests`, or dropped as now computed):
 
@@ -101,7 +100,7 @@ Requests reference the link one-directionally via `input_port_id`; the link hold
 | Column | Meaning |
 | --- | --- |
 | `id` | PK |
-| `input_port_id` → input_ports (ON DELETE CASCADE) | the link this request belongs to |
+| `input_port_id` → input_ports | the link this request belongs to |
 | `status` | PENDING → APPROVED / DENIED |
 | `justification` | the consumer's reason for this request; on a renewal the previous request's justification is reused |
 | `requested_duration_days` (nullable) | requested window length; NULL = permanent |
@@ -131,14 +130,14 @@ All paths stay on `/api/v2`; changes are additive.
 
 * `GET /data_products/{id}/input_ports` and `GET /explorations/{id}/input_ports` — list a consumer's input ports. Response adds `is_expired`, `is_expiring_soon`, the effective end date (`valid_until`) and the request history; grant fields come from the active request, not the flat link.
 * `POST /data_products/{id}/input_ports` and `POST /explorations/{id}/input_ports` — request access; also the single entry point for renewal. If a link already exists for that (consumer, output port), it adds a new request instead of failing as "already exists" or mutating the old grant; on a renewal the previous justification is reused. Blocked only if a request is already PENDING on that link.
-* `DELETE /data_products/{id}/input_ports/{output_port_id}` (and exploration equivalent) — unlink; deletes the link, and its requests are removed by the FK cascade.
+* `DELETE /data_products/{id}/input_ports/{output_port_id}` (and exploration equivalent) — unlink;
 
 **Producer side** (an output port owner managing consumers):
 
 * `GET /data_products/{dp}/output_ports/{op}/input_ports` — list the consumers of an output port; same additive response changes as the consumer list.
 * `POST .../input_ports/approve` and `POST .../input_ports/deny` — decide the link's single PENDING request; sets `status` and `decided_by/on`, and on approve sets `valid_from/valid_until` from the requested duration.
 * `POST .../input_ports/renew` — producer-initiated extension; adds a new APPROVED request (`decided_by` = the producer) rather than overwriting the current grant's dates.
-* `POST .../input_ports/remove` — revoke; deletes the link (FK cascade removes its requests).
+* `POST .../input_ports/remove` — revoke; deletes the link
 
 **Approver queue:**
 
@@ -146,22 +145,22 @@ All paths stay on `/api/v2`; changes are additive.
 
 ### Edge cases and outcomes
 
-| # | Scenario | Outcome |
-| --- | --- | --- |
-| 1 | Unrestricted OP, permanent | Request auto-approved, `valid_until=NULL` → has access, no expiry |
-| 2 | Unrestricted OP, time-bound | Auto-approved, `valid_from=now`, `valid_until=now+days` → access until it lapses |
-| 3 | Restricted OP, time-bound, request | Request PENDING (`requested_duration_days` set) → no access; on approve → `valid_from/until` set |
-| 4 | Restricted OP request denied | Request DENIED; the link stays, shown as denied; no access |
-| 5 | Re-request after denial | New request on the same link — no duplicate link |
-| 6 | Active grant + renewal denied | R1 still active → has access; R2 shown as denied. Both visible in history |
-| 7 | Active grant + renewal approved | R2 approved with a later `valid_until`; effective end = latest window; R1 historical |
-| 8 | Renewal after expiry | Old grant lapsed → no access; renewal request → approve → has access again |
+| # | Scenario | Outcome                                                                                              |
+| --- | --- |------------------------------------------------------------------------------------------------------|
+| 1 | Unrestricted OP, permanent | Request auto-approved, `valid_until=NULL` → has access, no expiry                                    |
+| 2 | Unrestricted OP, time-bound | Auto-approved, `valid_from=now`, `valid_until=now+days` → access until it lapses                     |
+| 3 | Restricted OP, time-bound, request | Request PENDING (`requested_duration_days` set) → no access; on approve → `valid_from/until` set     |
+| 4 | Restricted OP request denied | Request DENIED; the link stays, shown as denied; no access                                           |
+| 5 | Re-request after denial | New request on the same link — no duplicate link                                                     |
+| 6 | Active grant + renewal denied | R1 still active → has access; R2 shown as denied. Both visible in history                            |
+| 7 | Active grant + renewal approved | R2 approved with a later `valid_until`; effective end = latest window; R1 historical                 |
+| 8 | Renewal after expiry | Old grant lapsed → no access; renewal request → approve → has access again                           |
 | 9 | Expiring soon | Backend sets `is_expiring_soon` from `valid_until`; UI shows renew button + warning; no state change |
-| 10 | Permanent grant | `valid_until=NULL` → no expiry, no extend |
-| 11 | Producer extend | Producer adds a new approved request (`decided_by` = producer) |
-| 12 | Second request while one pending | Blocked (one pending per link) → 400; DB partial-unique-index backstop |
-| 13 | Remove / unlink | Delete the link; requests removed by FK cascade |
-| 14 | Grant lapses while renewal still pending | No access until re-approval (a natural gap — replaces today's "renewing → PENDING") |
+| 10 | Permanent grant | `valid_until=NULL` → no expiry, no extend                                                            |
+| 11 | Producer extend | Producer adds a new approved request (`decided_by` = producer)                                       |
+| 12 | Second request while one pending | Blocked (one pending per link) → 400; DB partial-unique-index backstop                               |
+| 13 | Remove / unlink | Delete the link -> Maybe w never want to delete the ink, just put it on inactive to keep the audit   |
+| 14 | Grant lapses while renewal still pending | No access until re-approval (a natural gap — replaces today's "renewing → PENDING")                  |
 
 ## Open Questions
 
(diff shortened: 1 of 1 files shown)


docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+| `requested_duration_days` | moved to the request |
+
+### `input_port_requests` — the new table (one row per request)
+
+Requests reference the link one-directionally via `input_port_id`; the link holds no request list.
+
+| Column | Meaning |
+| --- | --- |
+| `id` | PK |
+| `input_port_id` → input_ports (ON DELETE CASCADE) | the link this request belongs to |
```
stijn-janssens: Are we sure this needs to be ON DELETE CASCADE? If we see the requests as an audit table maybe we never want to remove rows? Also not if the full input port link is removed, especially not if we can later on readd it and want to see that it had access previously?

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+| `status` | PENDING → APPROVED / DENIED |
```
stijn-janssens: no expired?

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+* **Bad, because** renewals overwrite the grant in place, so there is no history.
+* **Bad, because** denials and renewals cannot be audited, and a denial justification cannot be added.
+* **Bad, because** it needs an `EXPIRED` status maintained by a background task and an `is_renewing` flag to track in-flight renewals.
+
+### Option 2: Split into link + requests
+
+* **Good, because** every request/renewal/denial is its own row — full history and audit trail.
+* **Good, because** an active grant can coexist with a pending or denied renewal.
+* **Good, because** access and expiry are computed live, removing the background task and its edge cases.
+* **Good, because** a denial justification fits naturally on the request row.
```
stijndehaes, line 63: * **
```suggestion
* **Good, because** a denial justification fits naturally on the request row.
* **Good, because** bugs are easier to fix and figure out since we have an audit trail, reducing the mental burden for everyone
```

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Pros and Cons of the Options
+
+### Option 1: Keep the single table
+
+* **Good, because** no schema split and no migration of existing rows.
+* **Neutral, because** it keeps the model familiar to the current code.
+* **Bad, because** renewals overwrite the grant in place, so there is no history.
+* **Bad, because** denials and renewals cannot be audited, and a denial justification cannot be added.
+* **Bad, because** it needs an `EXPIRED` status maintained by a background task and an `is_renewing` flag to track in-flight renewals.
```
stijndehaes:
```suggestion
* **Bad, because** it needs an `EXPIRED` status maintained by a background task and an `is_renewing` flag to track in-flight renewals.
* **Bad, because** it's easy for bugs to slip in and hard to figure them out since we will miss an audit trail
```

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+### `input_ports` — the link (output port ↔ consumer)
+
+Keep (identity only):
+
+| Column | Meaning |
+| --- | --- |
+| `id` | PK |
+| `dataset_id` → datasets | the output port |
+| `consuming_abstract_data_product_id` → abstract_data_products | the consuming data product / exploration |
+| `created_on`, `updated_on` | audit |
```
stijndehaes, line 81:
```suggestion
| `created_by` | owner, the one we need to notify for renewals | 
| `created_on`, `updated_on` | audit |

```

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+The frontend composes its display badge from these plus the access indicator; there is no separate stored status field for the display.
+
+### Endpoint changes
+
+All paths stay on `/api/v2`; changes are additive.
+
+**Consumer side** (a data product or exploration managing what it consumes):
+
+* `GET /data_products/{id}/input_ports` and `GET /explorations/{id}/input_ports` — list a consumer's input ports. Response adds `is_expired`, `is_expiring_soon`, the effective end date (`valid_until`) and the request history; grant fields come from the active request, not the flat link.
+* `POST /data_products/{id}/input_ports` and `POST /explorations/{id}/input_ports` — request access; also the single entry point for renewal. If a link already exists for that (consumer, output port), it adds a new request instead of failing as "already exists" or mutating the old grant; on a renewal the previous justification is reused. Blocked only if a request is already PENDING on that link.
```
stijn-janssens: Don't we only want this if access is not permanent?

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+| `dataset_id` → datasets | the output port |
+| `consuming_abstract_data_product_id` → abstract_data_products | the consuming data product / exploration |
+| `created_on`, `updated_on` | audit |
+
+Add:
+
+| Add | Meaning |
+| --- | --- |
+| link access indicator — reused `status`, or a new `is_active` | how the link signals "has access" for counts / graph / ordering / filtering; derived from its requests, not stored. Two options — see Open Questions. |
+| `UniqueConstraint(dataset_id, consuming_abstract_data_product_id)` | one link per (output port, consumer), so a re-request adds a request to the existing link instead of creating a duplicate link. |
```
stijndehaes: I don't think we want that for now, I think we want in a denial scenario that people have to start over for now no?

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+
+Remove (moved to `input_port_requests`, or dropped as now computed):
+
+| Removed column | What happens to it |
+| --- | --- |
+| `justification` | moved to the request (each request has its own reason; also allows a denial justification later) |
```
stijndehaes:
```suggestion
| `justification` | moved to the request, can be copied over for renewals (each request has its own reason; also allows a denial justification later) |
```

docs/adr/0021-input-ports-and-input-port-requests.md
stijndehaes: Or we need another way to figure out whom to notify

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+| `requested_on`, `requested_by_id` | moved to the request |
+| `approved_on`, `approved_by_id` | moved to the request as `decided_on` / `decided_by_id` |
+| `denied_on`, `denied_by_id` | moved to the request as `decided_on` / `decided_by_id` |
+| `renewed_on`, `renewed_by_id` | dropped — a renewal is its own request; its `decided_on` / `decided_by` is the "renewed by/on" |
+| `is_renewing` | dropped — a renewal in flight is a PENDING request on the link |
+| `expired_on` | dropped — expiry is computed from `valid_until` |
+| `expires_on` | moved to the request as `valid_until` |
+| `total_range_start` | moved to the request as `valid_from` |
+| `total_range_end` | dropped — same value as `valid_until`; the overall span is derived from the request history |
+| `requested_duration_days` | moved to the request |
```
stijndehaes, line 103: Should there be a value that notes if this is a permanent requests, or a time bound one?

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+  * **Bad, because** `status = approved` no longer means "has access right now" — an approved-but-expired request is still `approved`, so a check that needs live access must also look at `valid_until`.
+  * **Bad, because** the link's `status` mirrors the latest request, so a link with an active grant and a newer pending renewal reads `pending`; "still has access while a renewal is in flight" is only visible via `valid_until`.
+
+* **Option B — a new `is_active` boolean.** Computed "currently grants access" (approved and window open).
+  * **Good, because** it is unambiguous, including while a renewal is pending.
+  * **Bad, because** the `status`-based counts, graph query and order-by must be refactored to use it.
+
+### Should this ADR include a decision backgorund-task vs computed fields as well?
+
+One of the main outcomes is that we remove a background task and introduce more computed fields in the backend.
```
stijn-janssens: I would add the computed field "is active" (or however we decide to do it) to this ADR because it is so important.

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Endpoint changes
+
+All paths stay on `/api/v2`; changes are additive.
+
+**Consumer side** (a data product or exploration managing what it consumes):
+
+* `GET /data_products/{id}/input_ports` and `GET /explorations/{id}/input_ports` — list a consumer's input ports. Response adds `is_expired`, `is_expiring_soon`, the effective end date (`valid_until`) and the request history; grant fields come from the active request, not the flat link.
+* `POST /data_products/{id}/input_ports` and `POST /explorations/{id}/input_ports` — request access; also the single entry point for renewal. If a link already exists for that (consumer, output port), it adds a new request instead of failing as "already exists" or mutating the old grant; on a renewal the previous justification is reused. Blocked only if a request is already PENDING on that link.
+* `DELETE /data_products/{id}/input_ports/{output_port_id}` (and exploration equivalent) — unlink; deletes the link, and its requests are removed by the FK cascade.
```
stijndehaes: I think we need a new api endpoint to fetch the list of requests underneath your input port no?

docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -0,0 +1,184 @@
...
+| 11 | Producer extend | Producer adds a new approved request (`decided_by` = producer) |
+| 12 | Second request while one pending | Blocked (one pending per link) → 400; DB partial-unique-index backstop |
+| 13 | Remove / unlink | Delete the link; requests removed by FK cascade |
+| 14 | Grant lapses while renewal still pending | No access until re-approval (a natural gap — replaces today's "renewing → PENDING") |
+
+## Open Questions
+
+### Link access indicator — reuse `status` vs a new `is_active`
+
+The link needs a queryable "does this consumer currently have access" signal for internal SQL — the approved-input-port counts, the lineage-graph edge, ordering and filtering. Today this is `input_ports.status`. Two ways to carry it forward; this needs a team decision.
```
stijn-janssens: I would suggest a combination of the 2.
I think we can only use a Status field, not a boolean flag, because I would want to reflect at least 3 states. Pending is (mostly on the first request) a valid state to have. But I would compute it somehow. And that is the tough implementation we have to do in this change.

The state needs to calculate at that point what the latest active non-pending state is and then optionally include whether another one is pending or not. The pending can then potentially be shown with a boolean?
I think it might make sense to still have the is_expired and is_pending on the input port link as a computed field for easy visualization in the frontend, but just keep the requests in an audit table as described.

stijndehaes (dismissed):

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Fair comment. Indeed maybe better to keep it, yes agreed, meaning in essence that there is never a delete from the input_ports table.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Well, I would keep it simple, the EXPIRED is just computed based on the valid_until in this design, but if you don't like that, we can add it as a state.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Well only if the new one is permanent I guess as well no? Becaues otherwise there is no way to make it timebound? or wdyt?

docs/adr/0021-input-ports-and-input-port-requests.md
stijn-janssens: from the requests, right? I think if the link is gone we do delete i?

docs/adr/0021-input-ports-and-input-port-requests.md
stijn-janssens: That follows directly out of the consuming product, no?

HannesDS: I like it

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: I moved that out, in the last version, dammit didn't update it. Fully agreed.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: My take was that if this us null as well as the valid_until is timebound. But that is a bit, indirect maybe?

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: I was doubting that, if this is something you want or not.

stijndehaes: We will build it when needed

docs/adr/0021-input-ports-and-input-port-requests.md
stijndehaes: Yeah I think that is too indirect and might be bug prone. You can also add a check constraint:
```
ALTER TABLE your_table_name
ADD CONSTRAINT chk_bla_not_null
CHECK (
    (text_column = 'bla' AND other_column IS NOT NULL) 
    OR 
    (text_column IS DISTINCT FROM 'bla')
);
```

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Currently decided to keep cascade after discussion since that is the current state as well, included an open point in the end about removing the option of deleting an input port.

stijn-janssens (approved):
Looks good to me, I think we will need a computed field for expiring_soon and is_renewing to clearly show it in the UX but that is implementation detail

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: As discussed the status of the input port is expire not the request, that is still approved or denied, however teh status of the input_port itself can become expired.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Agreed,permanents can't get renewed. Documented it.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: As discussed in the meeitng, we will keep certain computed fields fin the backend such as 
expiring_soon and is_renewing for easy UI rendering, but next to that all logic will be run by a background task.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Agreed, adding timebound or permanent AccessType as a field, making it direct.

docs/adr/0021-input-ports-and-input-port-requests.md
HannesDS: Decided not to go that route and keep ENUM

PR #3845, opened 2026-07-14
refactor: split input ports into a link and access requests (ADR-0021)

HannesDS:
refactor: split input ports into a link and access requests (ADR-0021)

  Split the single input_ports table into two: 
  - input_ports: the link (consumer <-> output port) carrying the current access status
  - input_port_requests: one row per access request/decision (append-only history)

  Existing behaviour and the API contract are kept intact: the link re-exposes the
  fields that moved to the request via proxy accessors, so response schemas, the
  OpenAPI spec and generated clients don't change.

  In scope
  - Alembic migration with backfill (one request per existing link) and the demo seed
  - Models, services, routers and test factories moved to the two-table shape

  Out of scope (follow-up branch)
  - Renewals, time-bound access logic and the expiry background job
  - Computed is_expiring_soon / is_renewing and the request-history endpoint
    The forward-looking columns exist but stay inert for now.
  
  ## Checklist
  
  - [x] ~~Update release notes~~ — N/A (no user-facing change; contract
  unchanged)
  - [x] ~~UI screenshots/video~~ — N/A (only an internal form-field rename)
  - [x] Updated sample data (`backend/sample_data.sql`) so the new structure can
  be tested


backend/app/data_products/output_ports/input_ports/service.py
HannesDS, line 243: @stijndehaes  Could you verify this specifically since there was a merge conflict on this method.

backend/app/abstract_data_product/input_ports/model.py
vlieven, line 1: A comment on the structure / separation of concerns in this file: I see that you have the new imports `from fastapi import HTTPException` and `from starlette import status`.

First remark: fastapi aliases most starlette imports to something in the fastapi namespace. It's preferable to use those - simply for readability. Then we just use stuff from fastapi, and the fact that fastapi uses starlette under the hood remains an implementation detail. For this specific import you have https://fastapi.tiangolo.com/reference/status/

Second remark: These are networking / request concerns, while this file handles the database. I see these are used in a function to approve a request. These belong in the service layer, not the database layer, so that function would need to move (and the imports with it).

vlieven (dismissed):

backend/app/abstract_data_product/input_ports/model.py
```diff
@@ -35,28 +39,17 @@ class InputPort(
...
-        Enum(DecisionStatus),
-        default=DecisionStatus.PENDING,
+    status: Mapped[InputPortStatus] = mapped_column(
+        Enum(InputPortStatus, native_enum=False),
+        default=InputPortStatus.PENDING,
     )
-    requested_on = Column(DateTime(timezone=False), server_default=utcnow())
-    approved_on = Column(DateTime(timezone=False))
-    denied_on = Column(DateTime(timezone=False))
+    expiry_event_sent: Mapped[bool] = mapped_column(Boolean, default=False)
```
stijndehaes: Is this currently used? Because if it isn't I would for now still remove it, we will see what it should look like later :) I would assume it maybe even is a value on the request that is about to expire

backend/app/abstract_data_product/input_ports/model.py
stijndehaes: For the status thing, this happens in more places in our codebase, I believe we can fix this with a ruff rule. Ok to do in follow up PR?

backend/app/data_products/output_ports/input_ports/service.py
stijndehaes: I think it's fine for now, but we probably want to change this to return our Requests instead of Inputports at some point :) Let's add it todo to refactor `get_user_requests` and `get_user_pending_actions`

backend/app/database/alembic/versions/2026_07_13_1522-aa9f983049d2_split_input_ports_and_requests.py
```diff
@@ -0,0 +1,221 @@
...
+               ip.requested_on,
+               CASE ip.status
+                   WHEN 'APPROVED' THEN ip.approved_by_id
+                   WHEN 'DENIED' THEN ip.denied_by_id
+                   END,
+               CASE ip.status
+                   WHEN 'APPROVED' THEN ip.approved_on
+                   WHEN 'DENIED' THEN ip.denied_on
+                   END,
+               NULL,
```
stijndehaes: I believe this is incorrect, didn't we set valid_from to the approved_on date in our current implementation?

stijndehaes (changes requested):

backend/app/abstract_data_product/input_ports/model.py
HannesDS: No not yet, removed it.

backend/app/abstract_data_product/input_ports/model.py
HannesDS: First thing noted down, second, fully agree, moved to service.

backend/app/database/alembic/versions/2026_07_13_1522-aa9f983049d2_split_input_ports_and_requests.py
HannesDS: yes indeed, changed that!

backend/app/data_products/output_ports/input_ports/service.py
HannesDS: Added a comment, TODO's are not possible. Find it a bit sad to be honest that we cantt put TODO's but maybe that is bad practice. And rather create a story out of it or issue?

stijndehaes: We don't allow for TODO's in a codebase because they are almost impossible to follow up. We have a ticket with a list of todo's in it, as a best practice we add it there :)

stijndehaes (approved):

PR #3867, opened 2026-07-16
Feature/3754 timebound access renewal backend

HannesDS:
## What
  Feature: Add time-bound access renewal for input ports and surface
  every access state  
#3754 
  - requested
  - active 
  - active, renewal pending
  - active renewal declined, 
  - active expiring soon, 
  - expired
  - expired renewal pending
  across the consumer and producer UIs.
 

  ## Changes

  **Backend**
  - status is now an `InputPortStatus`
    (pending/approved/denied/expired) with a computed renewal status.
  - Add an `id` to the user input-port response so a pending action can
    be matched back to its output-port row.
  - Regenerate OpenAPI spec, SDK, CLI, and frontend API client.

  **Frontend**
  - Add a Status column (badge + a 'renewal pending' / 'renewal
    declined' / 'expiring soon' tag) and an Expiry date column to both
    the consumer input-port tab and the producer consumers tab.
  - New shared `access-status` component (`RenewalTag`, `ExpiryDate`)
    with a configurable expiring-soon threshold (14 days).
  - Migrate the input-port display path from `DecisionStatus` to
    `InputPortStatus` (new `getInputPortStatus*` helpers) so expired
    grants render correctly.
  - Bump input-port table page size 5 → 10 and fix cell alignment.

  **Seed**
  - Seed all seven access states for John Scientist from both the
    consumer and producer perspective on 'DEI Insights Dashboard',
    including a dedicated producer output port with its owner role
    assignment.

  ## Screenshots
image
image


backend/app/abstract_data_product/input_ports/model.py
```diff
@@ -69,37 +67,99 @@ class InputPort(
...
+
+    # @property
+    # def approved_by(self) -> Optional["User"]:
+    #     grant = self.active_grant
+    #     return grant.decided_by if grant is not None else None
+    #
+    # @property
+    # def denied_by(self) -> Optional["User"]:
+    #     request = self.latest_request
+    #     return request.decided_by if request.decision == DecisionStatus.DENIED else None
```
stijndehaes:
```suggestion
```

backend/app/abstract_data_product/input_ports/model.py
```diff
@@ -69,37 +67,99 @@ class InputPort(
         lazy="raise",
     )
 
+    # This the current request
+    # When you have an initial output port request this will be the pending one
+    # When there is an active approved request it will be that
+    # No active approved request, and latest is declined it will be declined
+    # When the latest is approved but expired it will return that one
```
stijndehaes:
```suggestion
```

stijndehaes: Currently the method is better described by the implementation then this doc :D

backend/app/abstract_data_product/service.py
```diff
@@ -63,6 +65,39 @@ def get_input_ports(self, data_product_id: UUID) -> Sequence[InputPortModel]:
...
+            adp.abstract_data_product_type, access_duration_type
+        )
+        if access_duration is None or access_duration.days is None:
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail=(
+                    "No time-bound access duration is configured for "
+                    f"{adp.abstract_data_product_type.value} access to this output port"
+                ),
+            )
```
stijndehaes:
```suggestion
        access_duration = AccessDurationService(self.db).get_access_duration(
            adp.abstract_data_product_type, access_duration_type
        )
        if access_duration is None:
            raise HTTPException(
                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
                detail=(
                    "No time-bound access duration is configured for "
                    f"{adp.abstract_data_product_type.value} access to this output port"
                ),
            )
```

Can't we just do this?

backend/app/abstract_data_product/service.py
```diff
@@ -82,19 +117,29 @@ def _add_single_input_port(
...
+                )
+            if (
+                existing.active_grant is not None
+                and existing.active_grant.valid_until is None
+            ):
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="This input port already has permanent access; there is nothing to renew",
+                )
+            justification = existing.latest_request.justification
```
stijndehaes, line 145: I think this method should just fail when the input port exists no?

The method is called `_add_single_input_port`, but here we are updating the existing input port for some reason with a new justification?

backend/app/data_products/output_ports/input_ports/service.py
```diff
@@ -165,31 +157,18 @@ def deny_output_port_as_input_port(
...
-            select(InputPortRequestModel).where(
-                InputPortRequestModel.decision.in_(
-                    [DecisionStatus.PENDING, DecisionStatus.APPROVED]
-                ),
-                InputPortRequestModel.input_port_id == current_link.id,
-            )
-        )
-        if pending_request is None:
+        target = current_link.pending_request or current_link.active_grant
+        if target is None:
```
stijndehaes, line 161: I think we should only handle pending_requests here, never the active_grant no?

frontend/src/components/input-port/access-status.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export function RenewalTag({ status, renewalStatus, validUntil }: RenewalTagProps) {
+    const { t } = useTranslation();
+    if (status === InputPortStatus.Pending) {
+        return null;
+    }
+    if (renewalStatus === RenewalStatus.Pending) {
+        return <Tag color={'blue'}>{t('Renewal pending')}</Tag>;
+    }
+    if (renewalStatus === RenewalStatus.Denied && status !== InputPortStatus.Denied) {
```
stijndehaes:
```suggestion
    if (renewalStatus === RenewalStatus.Denied) {
```

Can't it just be this? The backend should never send renewal status denied when inputport is also denied, then there is no renewal status

stijndehaes (dismissed):

backend/app/abstract_data_product/service.py
HannesDS: Well it depends how you implement it, this is curreently the renewal flow, but maybe we should put that in a seperate method and split it. or at least rename it to _add for an new one and _update if it exists already? WDYT? This PR or next one?

backend/app/data_products/output_ports/input_ports/service.py
HannesDS: Currently you can revoke access as a producer, and that is also a denny request, so it depends on what you want, currently this seems to work and make sense imo. What would make more sense if the cancellation or the revoke is in a seperate method. WDYT? 
Also while testing, it seemss like this revoking or deniying also write a row withvalid from valid to, which doesn't make much sense. I will also adapt that.  (latest row in screenshot is a deny on an approve). 
image

frontend/src/components/input-port/access-status.tsx
HannesDS: Well not sure about that, what if the previous request is denied, you make a new renewal, the new one also gets denied?

backend/app/abstract_data_product/service.py
HannesDS: Well this `access_duration.days is None` can happen, not sure we want to remove that condition. Because it is an optional, since the same is used for Permanent and time bound.

backend/app/abstract_data_product/service.py
stijndehaes: Conclusion:
- split into 2 functions: add input port and renewal 
- Split into 2 routes: add input port, and renewal

Will do in next PR

backend/app/data_products/output_ports/input_ports/service.py
stijndehaes: Idea for after this PR: Make a revoked state for `input_port_request`, and a `revoked_at` field denoting the time it was revoked. This makes it easy to see that this request was once approved, but later revoked. 

This also means we need a revoked state on `input_port`

frontend/src/components/input-port/access-status.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return daysUntilExpiry >= 0 && daysUntilExpiry <= EXPIRING_SOON_THRESHOLD_DAYS;
+}
+
+type RenewalTagProps = {
+    status: string;
+    renewalStatus?: string | null;
+    validUntil: string | null;
+};
+
+export function RenewalTag({ status, renewalStatus, validUntil }: RenewalTagProps) {
```
stijndehaes: Make 2 tags:
- Renewal tag, only get parameter `renewalStatus`, returns null when no tag is needed
- IsExpiringSoon tag, returns expiring soo tag, returns null

frontend/src/components/input-port/access-status.tsx
stijndehaes: See https://github.com/conveyordata/data-product-portal/pull/3867/changes/BASE..1dda7a78999ae15c381801f05933c9e9668c6cf8#r3603136220

backend/app/abstract_data_product/input_ports/model.py
```diff
@@ -71,36 +71,66 @@ class InputPort(
...
+                and (request.valid_from is None or request.valid_from <= today)
+                and (request.valid_until is None or request.valid_until >= today)
+            ),
+            None,
+        )
 
     @property
-    def requested_by(self) -> "User":
-        return self.latest_request.requested_by
+    def renewal_status(self) -> Optional[RenewalStatus]:
```
stijndehaes, line 100: Make a test for this, naughty @stijndehaes did not do it

backend/app/abstract_data_product/service.py
HannesDS: Fixed by separating the get access type into another method and also returning the access_duration object itself directly.

backend/app/abstract_data_product/input_ports/model.py
HannesDS: Tested it, was not working so also fixed it, good that we tested it

backend/app/data_products/output_ports/model.py
```diff
@@ -159,6 +158,23 @@ def data_product_name(self) -> str:
         nullable=False,
     )
 
+    def get_access_duration_type(
```
stijn-janssens: is this something we want in the model file?

frontend/src/components/input-port/access-status.tsx
```diff
@@ -0,0 +1,59 @@
+import { Tag, Typography } from 'antd';
+import { differenceInCalendarDays, parseISO } from 'date-fns';
+import { useTranslation } from 'react-i18next';
+import { InputPortStatus, RenewalStatus } from '@/store/api/services/generated/dataProductsApi.ts';
+import { formatDateFromISOString } from '@/utils/date.helper.ts';
+
+export const EXPIRING_SOON_THRESHOLD_DAYS = 14;
```
stijn-janssens: I would expect this to be configurable

HannesDS: Where, in the settings?

backend/app/data_products/output_ports/model.py
HannesDS: Maybe not.

frontend/src/components/input-port/access-status.tsx
HannesDS: Updated to be exposed via dedicated route and made available through the settings.

backend/app/data_products/output_ports/model.py
HannesDS: Moved this away from the model to the service.

stijn-janssens (approved):

PR #3920, opened 2026-07-23
Feature/3754 timebound access revoke

HannesDS:

#3754 
Add revoke/cancel actions for input port access

  Producers could only end a consumer's access by hard-deleting the
  input port link, and denying a request doubled as a way to pull
  already-approved access, which didn't reflect what actually happened.
  Consumers had the same ambiguity: cancelling a pending request and
  giving up approved access were both the same hard delete.

  Introduce revoke (ends an approved grant, available to either side)
  and cancel (withdraws a still-pending request, consumer-only) as
  first-class, non-destructive actions. Revoke sets revoked_at /
  revoked_by_id on the approved request without touching its decision;
  cancel moves the request's decision to a new CANCELLED value on a
  dedicated InputPortRequestDecision enum, kept separate from the
  shared DecisionStatus used by other approval flows. Deny is now
  scoped to pending requests only.

  recompute_status and current_request were reworked so a pending
  request never hides a link's real prior outcome (Expired/Revoked/
  Denied) on its own - only a first-ever request, or a fresh re-ask
  after a plain denial, still reports Pending, since nothing else
  would otherwise signal that a decision is awaited.

  The UI only ever exposes revoke and cancel. The original hard-delete
  routes (producer-side /remove, consumer-side DELETE) stay alive in
  the backend as a deliberate two-way door, unreferenced by any button,
  in case fully removing an input port ever turns out to be genuinely
  needed.                                                                  
  
 Input Port table view (consumer side)
image

Consumers table view (producer side)
image


docs/adr/0021-input-ports-and-input-port-requests.md
```diff
@@ -162,25 +179,25 @@ All paths stay on `/api/v2`; changes are additive.
...
-* `DELETE /data_products/{id}/input_ports/{output_port_id}` (and exploration equivalent) — unlink.
+* `POST /data_products/{id}/input_ports` and `POST /explorations/{id}/input_ports` — request access; also the single entry point for renewal. If a link already exists for that (consumer, output port), it adds a new request instead of failing as "already exists" or mutating the old grant; on a renewal the previous justification is reused. Blocked if a request is already PENDING on the link, or if the current active grant is permanent (`valid_until = NULL`) — permanent access never lapses, so there is nothing to renew, unless it is first revoked (see Ending access).
```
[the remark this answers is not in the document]
HannesDS: applied

frontend/src/components/abstract-data-products/input-port-tab/components/input-port-table/input-port-action-button.component.tsx
```diff
@@ -0,0 +1,148 @@
...
+            try {
+                setLoading(true);
+                await handleRevoke(outputPortId);
+                dispatchMessage({
+                    content: t('Access to Output Port {{name}} has been revoked', { name }),
+                    type: 'success',
+                });
+            } catch (error) {
+                console.error('Failed to revoke access to Output port', error);
+            } finally {
```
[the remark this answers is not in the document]
HannesDS: Agreed, added

frontend/src/utils/status.helper.ts
```diff
@@ -39,27 +40,44 @@ export function getBadgeStatus(
...
+        return t('Revoked');
+    }
     switch (status) {
         case DecisionStatus.Pending:
             return t('Requested');
         case DecisionStatus.Approved:
             return t('Available');
         case DecisionStatus.Denied:
             return t('Rejected');
+        case 'cancelled':
```
stijn-janssens: is it possible to use the enum here?

frontend/src/components/abstract-data-products/input-port-tab/components/input-port-table/input-port-action-button.component.tsx
```diff
@@ -35,27 +37,54 @@ export function InputPortActionButton({
...
-                            ? t('Request to link Output Port {{name}} has been cancelled', { name })
-                            : t('Output Port {{name}} has been removed', { name }),
+                    content: t('Request to link Output Port {{name}} has been cancelled', { name }),
                     type: 'success',
                 });
             } catch (error) {
-                status === InputPortStatus.Pending
-                    ? console.error('Failed to cancel Output port link request', error)
-                    : console.error('Failed to remove Output port', error);
+                console.error('Failed to cancel Output port link request', error);
```
stijn-janssens: I would remove the console error, we typically don't log to console

backend/app/events/enums.py
stijn-janssens, line 1: big fan of adding new proper events. But you have to adjust the frontend for this as well.
Check `frontend/src/types/events/event-types.ts`, add the new types there and verify where they are used (e.g. in the history-helper) in order to make sure they are properly fetching the required information

stijn-janssens (changes requested):
Overall really nice PR, i think the eventing is the only major remark I have. The rest are smaller comments

frontend/src/utils/status.helper.ts
HannesDS: Yes indeed, this is not nice, I also made it consistent to use InputPortStatus instead of DecisionStatus, since that is not used in the backend for input port. So now frontend is more inline with backend.

frontend/src/components/abstract-data-products/input-port-tab/components/input-port-table/input-port-action-button.component.tsx
HannesDS: Good to know! Fixed

backend/app/events/enums.py
HannesDS: Aaah cool! Okay added it

stijn-janssens (approved):

PR #3927, opened 2026-07-24
feat: Expire lapsed input port grants via a daily background task

HannesDS:

#3754 

Add a daily (+ on-startup) task that recomputes link status for
approved/expired links and emits the access-ended event once a grant's
window has passed. An expiry_event_sent flag guarantees delivery: it's
only set on a successful webhook call, so a failed delivery is retried
on the next run, and it's cleared whenever a new grant is approved.


backend/app/abstract_data_product/input_ports/background_tasks.py
```diff
@@ -0,0 +1,66 @@
...
+from sqlalchemy.orm import Session, selectinload
+
+from app.abstract_data_product.input_ports.enums import InputPortStatus
+from app.abstract_data_product.input_ports.model import InputPort
+from app.core.context import close_event_context, open_event_context, pop_events
+from app.core.logging import logger
+from app.core.webhooks.v2 import call_v2_webhook
+from app.database.database import SessionLocal
+
+CHECK_INTERVAL_SECONDS = 86400
```
stijn-janssens: I prefer to have this as a calculation so it's easy to see at once. `24*60*60` is for me immediately 24h, where 86400 seconds is not :P 
```suggestion
CHECK_INTERVAL_SECONDS = 24*60*60
```

stijn-janssens (changes requested):
Why did you decide to work with the expiry_event_sent? As far as I know currently we don't have any checks on whether events got send out / received correctly for any of the other events? Is this a special case?

HannesDS:
> Why did you decide to work with the expiry_event_sent? As far as I know currently we don't have any checks on whether events got send out / received correctly for any of the other events? Is this a special case?

Well that is what I discussed with @stijndehaes to have as a column. Not sure what he thinks, otherwise I can remove it but it makes is quite easy to know to sent an event or not.

HannesDS:
> > Why did you decide to work with the expiry_event_sent? As far as I know currently we don't have any checks on whether events got send out / received correctly for any of the other events? Is this a special case?
> 
> Well that is what I discussed with @stijndehaes to have as a column. Not sure what he thinks, otherwise I can remove it but it makes is quite easy to know to sent an event or not.

So you are right, no need for this. I learned there is a middleware that keeps track of the context and the class EventTrackedMixin that by default adds events to the context and emits them when needed. Because of that I don't need any state, and the backgorund task uses that same emit_events method as the middleware does.

backend/app/main.py
```diff
@@ -76,16 +80,21 @@ async def lifespan(_: FastAPI):
             AuthorizationService(db).reload_enforcer()
         db.commit()
 
+    with database.SessionLocal() as db:
```
stijn-janssens: why does this have to be executed upfront explicitly?

stijn-janssens: the task is immediately called and executed I think?

backend/app/core/webhooks/v2.py
```diff
@@ -1,17 +1,26 @@
...
 
-async def call_v2_webhook(event_type: str, data: dict) -> None:
+async def emit_all_events(events: Sequence[V2Event]) -> None:
+    # Good for keeping sequential order, bad for speed, maybe also good for not DoSing the receiver.
+    # To be checked if we ever run into issues with this.
+    for event in events:
+        await call_v2_webhook(type(event).event_type(), event.model_dump(mode="json"))
+
+
+async def call_v2_webhook(event_type: str, data: dict) -> bool:
```
stijn-janssens: why the boolean return?

HannesDS: No need indeed, AI slop.

backend/app/main.py
HannesDS: You are right, agin AI slop.

stijn-janssens (approved):
LGTM!

PR #3928, opened 2026-07-24
Feature/3574 timebound access docs

HannesDS:
- [x] Update the release notes document if needed in the [release notes](../docs/docs/release-notes.md).
- [x] When updating the UI please provide screenshots or videos to the reviewers.
- [x] Updated sample data so the new feature can be easily tested.

Update the documentation for feature #3754 timebound access as well as the release notes.

PR #3940, opened 2026-07-27
Feature/3574 timebound access docs

HannesDS:
## What
Feature: document the timebound access flow for input/output ports.

## Why
Supersedes #3928, which was accidentally branched off `feature/3574-timebound-access-background-task` instead of `main`, so its diff dragged in that whole branch's commits. This branch is cherry-picked cleanly onto `main` with just the docs commit.

## Changes
- Document renewal/expiry behavior for input ports
- Document access duration behavior for output ports
- Add release note entry
- Document creating products with timebound access


docs/docs/release-notes.md
stijn-janssens, line 1: I would refer to the detailed docs you wrote in this release note

stijn-janssens (approved):

PR #3975, opened 2026-07-30
fix: Compare admin expiry in UTC instead of server clock

HannesDS:
check_expired_admins compared admin_expiry against func.now(),
  which resolves to the database server's session time zone rather
  than a guaranteed UTC value. Comparing against a Python UTC
  datetime instead removes that ambiguity.

  Split the loop from its single-pass logic (revoke_expired_admins)
  so the expiry check can run in isolation, matching the pattern
  already used by expire_input_ports/expire_input_ports_task. Added
  unit tests covering lapsed, future, and unset expiry.

### DIFF
--- a/backend/app/core/authz/background_tasks.py
+++ b/backend/app/core/authz/background_tasks.py
@@ -1,6 +1,8 @@
 import asyncio
+from datetime import datetime, timezone
 
-from sqlalchemy import func, select
+from sqlalchemy import select
+from sqlalchemy.orm import Session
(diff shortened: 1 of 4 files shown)


stijn-janssens (approved):

stijn-janssens:
maybe add this to the bugfixes in the release notes

HannesDS:
> maybe add this to the bugfixes in the release notes

Agreed!

Issue #3981, opened 2026-08-03
Access modes work


stijndehaes:
- [x] First implementation
- [x] Should access modes on output port be a query instead of a property?
- [x] Should we allow adding/removing access modes after Technical Asset Creation? What updates are currently allowed => OK to not do for now, let's see if we get feedback
- [x] Add documentation and release notes
- [x] Check if we can handle error better, I improved how we generate them backend, but frontend rtk query does not generate errors for it that we can handle. Also we now skip error handling per endpoint in the rtk base query, see if we can improve that. => Currently best we can do is add override when adding the tags
- [x] Configure access modes per Technical Asset Type instead
- [x] What should happen when someone removes a technical asset with access modes from an output port, with a request with access mode? => Leave as is for now, we will fix when issues are raised
- [x] Add access mode to review modal

PR #4000, opened 2026-08-04
Added access mode to technical asset

stijndehaes:
Currently only implement the backend:
- Access modes can be created by admin, and the description can be updated
- Access mode can be added during creation of technical asset
- Access modes can be seen on output port
- Only compatible technical assets can be added to a single output port
- User can request a single access mode when requesting access

In support of: #3981


backend/app/data_products/output_ports/model.py
```diff
@@ -117,6 +118,12 @@ def quality_status(self) -> Optional[str]:
     def data_product_name(self) -> str:
         return self.data_product.name
 
+    @property
+    def access_modes(self) -> list["AccessMode"]:
```
stijndehaes, line 122: I have a todo to make this a query

backend/tests/app/data_products/output_ports/test_router.py
```diff
@@ -378,31 +378,11 @@ def test_remove_output_port(self, client):
         response = self.get_output_port(client, ds.id, ds.data_product.id)
         assert response.status_code == 404
 
-    def test_get_dataset_with_invalid_dataset_id(self, client):
+    def test_get_output_port_with_invalid_id(self, client):
         dp = DataProductFactory()
         dataset = self.get_output_port(client, dp.id, self.invalid_id)
         assert dataset.status_code == 404
 
-    def test_get_dataset(
```
stijndehaes, line 386: a test_get_output_port already exists, this was from before the rename

backend/app/abstract_data_product/service.py
```diff
@@ -49,6 +50,8 @@ def _ensure_not_deleting(self, adp: AbstractDataProduct) -> None:
                 detail=f"{adp.abstract_data_product_type.value} '{adp.name}' is pending deletion and cannot be modified",
             )
 
+    import uuid
```
vlieven: This should not be floating here, you already `from uuid import UUID`, so the types below can just be `Optional[UUID]`

vlieven (approved):

PR #4002, opened 2026-08-04
Added GLOBAL__MANAGE_FINALIZERS action

stijndehaes:
This action is used to manage finalizers
and can only be used by admin role.


stijndehaes:
@vlieven I am correct in the assumption that admin role has access to everything automatically right?

vlieven:
> @vlieven I am correct in the assumption that admin role has access to everything automatically right?

The admin indeed has `*` permissions

vlieven (changes requested):
I am not sure about your goal. If you want this to really be only for admins, the check should be on if the user is an admin, not if the user has permission on an action that is not assignable.

If you intend that `GLOBAL__MANAGE_FINALIZERS` is never actually assigned, I would not specify it.

stijndehaes:
> I am not sure about your goal. If you want this to really be only for admins, the check should be on if the user is an admin, not if the user has permission on an action that is not assignable.
> 
> If you intend that `GLOBAL__MANAGE_FINALIZERS` is never actually assigned, I would not specify it.

I could also just make is assignable?

In the end we now have a system account that is allowed to manage things. Having an admin check feels counter intuitive, so let's just add it to the list? Or we can tie it to allowing delete of data product or exploration as an action? I think that might be a clean solution as well wdyt?

vlieven:
Tying it into only the delete does not work, as you also need it for the create to add it I think. Maybe it is fine to have it and add it for the system account.

vlieven (approved):

stijn-janssens (approved):

PR #4003, opened 2026-08-04
FormModal removed

stijndehaes:
It just renamed some values and set a single default. It did not bring any value so removed it to simplify code


frontend/src/components/modal/user-popup/user-popup.tsx
```diff
@@ -40,7 +39,13 @@ export function UserPopup({ onClose, isOpen, roles, item, isLoading, userIdsToHi
     }, [userIdsToHide, users, searchTerm]);
 
     return (
-        <FormModal title={t('Add User')} onClose={onClose} isOpen={isOpen} footer={(_, { CancelBtn }) => <CancelBtn />}>
+        <Modal
+            title={t('Add User')}
+            onCancel={onClose}
+            open={isOpen}
+            footer={(_, { CancelBtn }) => <CancelBtn />}
```
vlieven, line 46: Is this needed? What does it do?

vlieven (approved):

frontend/src/components/modal/user-popup/user-popup.tsx
stijndehaes: Don't know, it was there before :p

PR #4016, opened 2026-08-06
Improve background task handling

stijndehaes:
We now log failures.
When canceling we also wait until
the registered background tasks are done.

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -1,9 +1,10 @@
 import asyncio
 import re
 import urllib.parse
-from collections.abc import Awaitable, Callable
-from contextlib import asynccontextmanager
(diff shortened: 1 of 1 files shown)


vlieven (approved):
I wanted to do a spike around this code part.
Seems like we want some sort of scheduling component to handle these (seems to be quite brittle now).

I suppose this is already an improvement over what we had

stijndehaes:
> I wanted to do a spike around this code part. Seems like we want some sort of scheduling component to handle these (seems to be quite brittle now).
> 
> I suppose this is already an improvement over what we had

I agree this is not the correct way to handle this, at least we now wait until they are stopped

PR #4059, opened 2026-08-11
Use the name to filter plugins

stijndehaes:
Before we used __name__.
Also now the name of the DB
model is alligned with the
plugin name.

### DIFF
--- a/backend/app/technical_asset_configuration/service.py
+++ b/backend/app/technical_asset_configuration/service.py
@@ -37,7 +37,7 @@ class PluginService:
         configured_metadata = [
             name
             for name in data_output_configurations
-            if name.__name__ in configured_plugins
+            if name.name in configured_plugins
         ]
         return [
             metadata_response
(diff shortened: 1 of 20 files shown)


backend/app/technical_asset_configuration/service.py
```diff
@@ -66,6 +66,7 @@ def _build_metadata_response(
         self, plugin_class: type[AssetProviderPlugin]
     ) -> Optional[UIElementMetadataResponse]:
         """Build a complete metadata response for a plugin"""
+        platform_meta = plugin_class.get_platform_metadata()
         try:
             platform_meta = plugin_class.get_platform_metadata()
```
vlieven: You're now defining this twice, once inside and once outside the try. That does not seem correct

vlieven (commented):
I dislike the globals for `NAME` a bit, but I suppose if you define it on the class directly, you get circular reference errors. So I will live with it

backend/app/technical_asset_configuration/service.py
stijndehaes: Fixed, mistake after refactor and revert

stijndehaes:
> I dislike the globals for `NAME` a bit, but I suppose if you define it on the class directly, you get circular reference errors. So I will live with it

I so agree with you, I first had it defined on the plugin itself, and use in the model. However the issue was that the plugin definces ORM model, I don't think that that ORM model is actually used, but if you remove that it's not imported and sqlalchemy can't make it's model. So yeah.
There are other ways to fix this I suppose, but I wanted to focus on ensuring the names was right

vlieven (approved):

PR #4066, opened 2026-08-12
Track provisioning in daily metrics

stijndehaes:
Reports provisioned vs total data products
and explorations.

### DIFF
--- a/backend/app/core/logging/posthog_analytics.py
+++ b/backend/app/core/logging/posthog_analytics.py
@@ -14,39 +14,14 @@ from app.database.database import SessionLocal
 from app.settings import settings
 
 
-class PortalPosthog:
-    """Thin wrapper around the Posthog client that injects portal-level
-    context (e.g. Host) into every capture call automatically."""
-
-    def __init__(self, client: Posthog, host: str) -> None:
-        self._client = client
-        self._host = host
-
-    def capture(
-        self,
-        distinct_id: Any,
-        event: str,
-        properties: Optional[dict[str, Any]] = None,
-        **kwargs: Any,
-    ) -> None:
-        merged = {"host": self._host, **(properties or {})}
-        self._client.capture(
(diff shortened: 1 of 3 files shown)


backend/app/core/logging/posthog_analytics.py
```diff
@@ -1,6 +1,6 @@
 import asyncio
 from datetime import datetime, timedelta
-from typing import Any, Optional
+from typing import Any, Dict, Optional
```
vlieven:
```suggestion
from typing import Any, Optional
```

You don't need to import `Dict` as a type, you can just use `dict[X, Y]` as an annotation

backend/app/core/logging/posthog_analytics.py
```diff
@@ -57,55 +32,91 @@ def _seconds_until_next_midnight_utc() -> float:
     return (next_midnight - now).total_seconds()
 
 
-async def report_consumption_metrics_task() -> None:
-    """
-    Daily background task that reports the total number of approved input ports
-    to PostHog as a consumption metric, broken down by ADP type. Fires once at
-    midnight UTC regardless of how many times the app restarts during the day.
+def consumption_metrics() -> Dict[str, Any]:
```
vlieven:
```suggestion
def consumption_metrics() -> dict[str, Any]:
```

backend/app/core/logging/posthog_analytics.py
```diff
@@ -57,55 +32,91 @@ def _seconds_until_next_midnight_utc() -> float:
...
+    return {
+        "total_approved_input_ports": total,
+        **{
+            f"approved_input_ports_{adp_type}": count
+            for adp_type, count in counts_by_type.items()
+        },
+    }
+
+
+def provisioning_metrics() -> Dict[str, Any]:
```
vlieven:
```suggestion
def provisioning_metrics() -> dict[str, Any]:
```

vlieven (approved):

PR #4068, opened 2026-08-12
Added documentation on acces modes

stijndehaes:
In support of: #3981

### DIFF
--- a/docs/docs/concepts/technical-assets.md
+++ b/docs/docs/concepts/technical-assets.md
@@ -32,7 +32,7 @@ Output Ports should follow the advice that data that belongs together should als
 When creating a Technical Asset, you can choose between two provisioning strategies:
 
 ### Default Mapping
-This is the **recommended option for most users**. Default mapping uses the platform's standard provisioning conventions, automatically generating technical identifiers (paths, table names, schemas, etc.) based on your Data Product's namespace and configuration.
+This is the **recommended option for most users**. Default mapping uses the platform's standard provisioning conventions,automatically generating technical identifiers (paths, table names, schemas, etc.) based on your Data Product's namespace and configuration.
 
 **Benefits:**
 - Follows organizational naming conventions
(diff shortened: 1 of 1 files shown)


docs/docs/concepts/technical-assets.md
```diff
@@ -50,6 +50,25 @@ This option allows you to override the default provisioning and specify your own
...
+An example of a scenario where access modes are useful is a technical asset that is an API.
+Each different access mode could represent a different level of access to the API. For example:
+
+- **Basic**: Allows consumers to retrieve non-sensitive data through the API.
+- **Admin**: Allows consumers to retrieve sensitive data through the API.
+
+This could be represented with multiple Technical Assets as well, but that means you need multiple output ports. Which might
+result in a lot of Output Ports to choose from. Access Modes allows you to have a single Output Port in these cases, making
+things less overwhelming for consumers.
+
```
katarinamil, line 69: I would add: Output port can contain multiple technical assets only if all of technical assets belonging to the output port have the same access modes. If not, producer will have to create different output ports, which will lead to a lot of output ports to choose from, or create technical assets with identical access modes, which will allow to add them to the same output port.

docs/docs/concepts/technical-assets.md
```diff
@@ -50,6 +50,25 @@ This option allows you to override the default provisioning and specify your own
...
 **Recommendation:** Start with default mapping unless you have a specific requirement for custom configuration.
 
+## Access modes
+
+Technical Assets might have access modes defined or not.
+When no access modes are defined, granting access to a technical asset via the output port should grant read access
+through [provisioning](../developer-guide/provisioner.md) to the underlying technical asset.
+
+However when a Technical Asset exposes access modes, a consumer of an Output Port containing the Technical Asset must
+choose the access mode they want to use.
```
katarinamil: I would add: When creating technical asset, producer will see if admin has allowed multiple access modes for the asset and will be able to choose from the list of possible access modes for that asset type. Only then I would say that the consumer will be able to choose from the list of access modes.

docs/docs/concepts/technical-assets.md
stijndehaes: I added something like it but did not spell it out directly. Found it a bit too much text

docs/docs/concepts/technical-assets.md
```diff
@@ -32,7 +32,7 @@ Output Ports should follow the advice that data that belongs together should als
 When creating a Technical Asset, you can choose between two provisioning strategies:
 
 ### Default Mapping
-This is the **recommended option for most users**. Default mapping uses the platform's standard provisioning conventions, automatically generating technical identifiers (paths, table names, schemas, etc.) based on your Data Product's namespace and configuration.
+This is the **recommended option for most users**. Default mapping uses the platform's standard provisioning conventions,automatically generating technical identifiers (paths, table names, schemas, etc.) based on your Data Product's namespace and configuration.
```
stijn-janssens:
```suggestion
This is the **recommended option for most users**. Default mapping uses the platform's standard provisioning conventions, automatically generating technical identifiers (paths, table names, schemas, etc.) based on your Data Product's namespace and configuration.
```

stijn-janssens (approved):

PR #4069, opened 2026-08-12
Performance improvements for Portal

### LINKED ISSUE #3910: Heavy load causes health check to fail
## Describe the bug
When scraping portal the health check becomes slower, causing timeouts.

### Steps to reproduce
Put portal under load, check health check at the same time

### Expected behavior
Health check should not be impacted by load on Portal

stijndehaes:
I benchmarked our 2 most used endpoints:
- Search Output Ports
- Fetch Data Products

This resulted first in an abysmal score, 2 bottlenecks where detected:
- Database connections
- Loading of embedding model

So now we allow for users to tweak database connections and by default we use more connections.
We also properly cache the embedding model.

This result in a tremendous improvement under load testing when running about 50 users with 5r/s.
We went from a median in the 100's of ms, to 10's of ms

Fixes: #3910


stijndehaes:
Question for reviewe I currently left the performance script in, but we can remove it. It's AI generated and took about a minute. WDYT?

backend/app/data_products/output_ports/schema_response.py
```diff
@@ -39,6 +39,8 @@ class GetOutputPortAccessDurationsResponse(ORMModel):
 
 
 class BaseOutputPortGet(ORMModel):
+    model_config = ConfigDict(from_attributes=True, populate_by_name=True)
```
vlieven: According to the [docs](https://pydantic.dev/docs/validation/dev/api/pydantic/config/#pydantic.config.ConfigDict.populate_by_name), populate_by_name should not be used anymore.
I'm not sure what you want to achieve by it, but the modern replacement is this:

```suggestion
    model_config = ConfigDict(from_attributes=True, validate_by_name=True, validate_by_alias=True)
```

backend/app/core/embed/model.py
```diff
@@ -1 +1,14 @@
+from functools import lru_cache
```
vlieven:
```suggestion
from functools import cache
```

backend/app/core/embed/model.py
```diff
@@ -1 +1,14 @@
...
+
+from fastembed import TextEmbedding
+
 EMBEDDING_MODEL = "BAAI/bge-small-en-v1.5"
+
+
+@lru_cache(maxsize=1)
```
vlieven: This can just be `cache`, should be a bit more performant (eviction logic is skipped, and the function is parameterless, so there never can be anything to evict)

```suggestion
@cache
```

vlieven:
> Question for reviewe I currently left the performance script in, but we can remove it. It's AI generated and took about a minute. WDYT?

Are we planning to reuse it in the future? If so then we can keep it around. If not, maybe better to remove it, and document that load testing can be done by using locust and letting AI generate a script?

stijndehaes:
> > Question for reviewe I currently left the performance script in, but we can remove it. It's AI generated and took about a minute. WDYT?
> 
> Are we planning to reuse it in the future? If so then we can keep it around. If not, maybe better to remove it, and document that load testing can be done by using locust and letting AI generate a script?

I wouldn't mind running this nightly in CI and ensuring the performance is all right. I'll check this out in the following weeks if that  is ok

vlieven (approved):
Pico bello

PR #4114, opened 2026-08-18
Feature/coder access integration

HannesDS:

image

Feature: add a CoderPlugin workspace launcher tile, and drive the technical-asset and dataset action tiles from the existing plugin system instead of hardcoded Collibra/Datahub tiles.

The technical-asset and dataset action panels had Collibra/Datahub baked in directly, even though the backend already has a generic plugin/tile system (AssetProviderPlugin, /v2/plugins/platform-tiles). Coder is a new integration we want as a tile (a data-product-scoped launcher into a Coder VS Code workspace, following the same pattern as the existing Conveyor and Agno launcher plugins) — adding it was the forcing function to finally remove the hardcoded tiles and make both action panels fully plugin-driven.
  
  - Backend: add CoderPlugin (technical_asset_configuration/coder/schema.py) — builds a Coder workspace URL from the data product's GitHub repo (CODER_BASE_URL + CODER_GITHUB_ORG settings, both new and empty by default). Not added
  to ENABLED_PLUGINS, so it's inert until an environment configures it.
  - Frontend: TechnicalAssetActions and DatasetActions now fetch tiles from useGetPlatformTilesQuery()/useLazyGetPluginUrlQuery() instead of a hardcoded Collibra/Datahub tile list; added coder-logo.svg; removed the now-unused collibra-logo.svg and datahub-logo.svg assets and their translation strings.
  - Docs: fixed stale app/data_output_configuration/... path references in adding-integrations.md (the package was renamed to technical_asset_configuration).
  - Added test coverage for the new plugin (test_schema.py, test_router.py) and both refactored action components.


backend/tests/app/technical_asset_configuration/coder/test_schema.py
```diff
@@ -0,0 +1,28 @@
...
+        assert meta.display_name == "Coder"
+        assert meta.platform_key == "coder"
+        assert meta.show_in_form is False
+        assert meta.has_environments is False
+
+    @patch("app.technical_asset_configuration.coder.schema.settings")
+    def test_get_url_builds_workspace_link_from_data_product_namespace(
+        self, mock_settings, session
+    ):
+        mock_settings.CODER_BASE_URL = "https://ide.test.dp.uhasselt.be"
```
vlieven: Do we want uhasselt as example in our codebase?

frontend/src/pages/output-port/components/dataset-actions/dataset-actions.component.tsx
```diff
@@ -1,67 +1,86 @@
...
-// part of the backend platform_tiles endpoint
-const getDataPlatforms = (t: TFunction): CustomDropdownItemProps<string>[] => [
-    {
-        label: t('Collibra'),
-        value: 'collibra',
-        icon: collibraLogo,
-    },
-    { label: t('Datahub'), value: 'datahub', icon: datahubLogo, disabled: true },
-];
-
```
vlieven, line 26: I agree that the todo should be resolved, but now you just throw everything away? I think we want/need to support these. Pretty sure at least datahub is used in the wild

HannesDS: Well, they are not yet added as a platform and it was only hardcoded in the frontend to actually support Collibra and Datahub again, I need new backend plugins: technical_asset_configuration/collibra/schema.py and .../datahub/schema.py, each subclassing AssetProviderPlugin with get_url.

Basically create it, I think it is a bit out of scope for this PR, but if you want I can add them again?

backend/tests/app/technical_asset_configuration/coder/test_schema.py
HannesDS: Indeed, maybe better to have something general.

frontend/src/pages/output-port/components/dataset-actions/dataset-actions.component.tsx
vlieven: If you think it's out of scope to add them to the PR, I'd argue it's also out of scope to remove them :)

HannesDS: Discuess IRL, summary: The fields never worked and were just hardcoded in the front-end nothing matches it on backend side making them just visual mocks.

backend/app/technical_asset_configuration/rustfs/schema.py
```diff
@@ -0,0 +1,103 @@
...
+from app.technical_asset_configuration.rustfs.model import (
+    NAME,
+)
+from app.technical_asset_configuration.rustfs.model import (
+    RustFSTechnicalAssetConfiguration as RustFSTechnicalAssetConfigurationModel,
+)
+from app.users.schema import User
+
+
+class RustFSTechnicalAssetConfiguration(AssetProviderPlugin):
```
stijn-janssens, line 31: Isn't it cleaner to let this inherit from S3TechnicalAssetConfiguration?

HannesDS: Well I could, but then I ingerit quite some fields that I don't need, and if there are ever changes, specific for AWS or S3 I also inherit them which is not sure that I want it. What we could have in the future is an `ObjectStorageTechincalAssetConfiguration` maybe, that is general, however for now that seems a bit overkill. WDYT?

stijn-janssens (approved):

PR #4133, opened 2026-08-24
Allow uploading data model through UI

stijndehaes:

image

As owner on empty page:
Screenshot 2026-08-24 at 14 26 08

As non owner on empty:
image


backend/app/open_api_export.py
```diff
@@ -97,5 +107,38 @@ def custom_openapi(app: FastAPI):
     return app.openapi_schema
 
 
+def fix_file_upload_methods(
```
stijn-janssens, line 110: can you briefly explain what has to happen here or why the initial generation breaks?

docs/docs/developer-guide/schema-information.md
```diff
@@ -66,10 +66,11 @@ This gives consumers a quick understanding of the structure of the data an outpu
 Obtain or create the Bitol ODCS data contract that describes your output port.
 A complete example contract is available [here](https://github.com/conveyordata/data-product-portal/blob/main/integrations/bitol/data-contract-example.yml)
 
-Portal accepts the full ODCS document, so you can post the contract as-is; only its `schema` section is used.
+Portal accepts the full ODCS document, so you can post the contract as-is; only its `schema` section is used.w
```
stijn-janssens:
```suggestion
Portal accepts the full ODCS document, so you can post the contract as-is; only its `schema` section is used.
```

stijn-janssens (approved):

backend/app/open_api_export.py
stijndehaes: Added an explenation, thanks for pointing this out :)

stijn-janssens: alright, clear reply. Anything we can change on rtk side to be better compatible with 3.1? Or just something they have to fix at some point?

PR #4134, opened 2026-08-24
Casbin broadcasts changes so processes can reload

stijn-janssens:
Noticed some inconsistencies because one of our portal deploys is running on multiple tasks. This caused inconsistencies between actual access and access reported by the API.
I believe this is the issue and the solution


stijndehaes (changes requested):
Why not use the official casbin postgres watcher?
https://casbin.apache.org/docs/watchers

stijn-janssens:
According to my understanding, the official package is built for synchronous frameworks such as Django and Flask.
It spawns a child process and communicates via a pipe. To bridge this into a async lifespan for fastapi we would end up using more code.
Next to that (although minor) the package also seems not very well maintained, last push is 6 months ago and it has 1 star. Maybe maintaining the 50 lines of code we add is easier?

The goal was to implement a similar pattern as the watcher has but with native async capabilities.
I can for sure add this to the ADR if you agree with this reply.

stijndehaes:
> According to my understanding, the official package is built for synchronous frameworks such as Django and Flask. It spawns a child process and communicates via a pipe. To bridge this into a async lifespan for fastapi we would end up using more code. Next to that (although minor) the package also seems not very well maintained, last push is 6 months ago and it has 1 star. Maybe maintaining the 50 lines of code we add is easier?
> 
> The goal was to implement a similar pattern as the watcher has but with native async capabilities. I can for sure add this to the ADR if you agree with this reply.

This package works very similar to what you do :) Just native functionallity of casbin.
The pipe is just a communication channel between the background listener and the watcher itself to let it know about any changes.


In general the packages of casbin don't receive many updates, extensions typically just don't change too often to I don't think it's an issue :) I do notice that all golang packages are in general a lot better used and maintained than the python ones.

backend/app/core/authz/watcher.py
```diff
@@ -0,0 +1,53 @@
...
+                await conn.close()
+            raise
+        except Exception as exc:
+            if conn:
+                with contextlib.suppress(Exception):
+                    await conn.close()
+            logger.warning(
+                "[authz] Casbin watcher disconnected (%s), retrying in 5 s", exc
+            )
+            await asyncio.sleep(5)
```
stijndehaes: Here there may be issues, should we not trigger a reload right after we start listening again? Just to be sure all policies are loaded?

In golang we just reload every minute, so I think reloading should be safe

stijndehaes (changes requested):

PR #4151, opened 2026-08-25
Switch Casbin to SyncedEnforcer and add auto reload every 30 sec

stijn-janssens:

### DIFF
--- a/backend/app/core/authz/authorization.py
+++ b/backend/app/core/authz/authorization.py
@@ -3,8 +3,8 @@ from typing import Awaitable, Callable, Sequence, TypeAlias, Union
 from uuid import UUID
 
 import casbin_sqlalchemy_adapter as sqlalchemy_adapter
-from cachetools import Cache, LRUCache, cachedmethod
-from casbin import Enforcer
+from cachetools import Cache, TTLCache, cachedmethod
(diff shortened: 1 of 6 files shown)


backend/app/core/authz/authorization.py
```diff
@@ -26,21 +26,26 @@
 
 class Authorization(metaclass=Singleton):
     def __init__(self) -> None:
-        self._enforcer: Enforcer = self._initialize()
-        self._cache: Cache = LRUCache(maxsize=settings.AUTHORIZER_CACHE_SIZE)
+        self._enforcer: SyncedEnforcer = self._initialize()
+        self._cache: Cache = TTLCache(
```
stijndehaes, line 30: I think we can just use `SyncedCachedEnforcer` they have it implemented, it's just not in the docs, then we can throw away our own cache

stijndehaes (changes requested):

backend/app/core/authz/authorization.py
stijndehaes: It's not released yet, so we probably should make a ticket to upgrade and keep things as is for now

stijndehaes (approved):

PR #4155, opened 2026-08-25
Hidden data products ADR

stijndehaes:

### DIFF
--- a/docs/adr/0022-data-product-visibility.md
+++ b/docs/adr/0022-data-product-visibility.md
@@ -1,0 +1,105 @@
+# Data product visibility
+
+## Context and Problem Statement
+
+We will add a `visibility` field to data products with two states: `discoverable` and `hidden`. Hidden data products should
+only be visible to users with an approved role on the product, while discoverable data products should be visible to any
+authenticated user.
+
+Currently there is already such an implementation for Output Ports, however there are some holes in the implementation.
+For example sub routers of output port like: contract, data quality etc do not check for visibility and thus can leak information on hidden output ports.
+
+We do not want to repeat these mistakes.
+
+We currently already use Casbin for authorization and resource-scoped role assignments. There is an opportunity to keep visibility
+decisions in the same system instead of adding a separate application-side check that can drift from the RBAC model or be missed on nested routes.
+
+This requirement applies not just to the data product detail route, but also to child resources under a product,
+such as technical assets and services. We need the visibility decision to be centralized, auditable, and consistent across all product-scoped routes.
+
+We also want to reuse this mechanism for Input Ports, and the list of consumers for Output Ports. Hidden Data Products to
+which you do not have access should be redacted from the list. Only public information should be visible: ID, requester (so you know whom to talk to when reviewing), business reason etc.
+Not the name of the Data Product, description ...
+So we should be able to reuse the same mechanism, for these list calls. This will not be possible through endpoint
+dependency checks, the service itself will have to do filtering and redaction based on the visibility of the product.
+
+
+## Decision Drivers
+
+* Align visibility enforcement with the existing Casbin authorization model
+* Reuse the current data-product assignment flow instead of introducing a second access mechanism
+* Keep discoverable products available to any authenticated user without per-user allowlisting
+* Avoid route-by-route checks that are easy to miss on nested or derived routes
+* Preserve the existing assignment approval and revocation flow for hidden products
+* Keep the implementation predictable as product visibility changes over time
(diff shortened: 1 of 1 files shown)


docs/adr/0022-data-product-visibility.md
```diff
@@ -0,0 +1,95 @@
+# Data product visibility
+
+## Context and Problem Statement
+
+We added a `visibility` field to data products with two states: `discoverable` and `hidden`. Hidden data products should
```
stijn-janssens: this is not yet there right? It's worded like it is :)

stijn-janssens (commented):
I think this is a solid approach for access inside the data product itself. Being a member of a hidden data product allows you to see it, clear.

However. How does this work for transitive access? What would happen to members of consuming products of hidden data products?
How can hidden data products be consumed? What with the explorer graph?

Does this also all work with casbin? I think this can be further clarified

I might be introducing scope creep with this question if we have signed of on hidden products being completely hidden and only accessible by it's own members. However then we also need to take into account further steps such as blocking shopping for input ports for these products etc etc

So long story short, ADR makes sense and is clear for me on memberships. But the connectedness or non-allowing of connectedness is not clear for me

docs/adr/0022-data-product-visibility.md
stijndehaes: Yeah it's not yet there

docs/adr/0022-data-product-visibility.md
```diff
@@ -0,0 +1,105 @@
...
+
+We currently already use Casbin for authorization and resource-scoped role assignments. There is an opportunity to keep visibility
+decisions in the same system instead of adding a separate application-side check that can drift from the RBAC model or be missed on nested routes.
+
+This requirement applies not just to the data product detail route, but also to child resources under a product,
+such as technical assets and services. We need the visibility decision to be centralized, auditable, and consistent across all product-scoped routes.
+
+We also want to reuse this mechanism for Input Ports, and the list of consumers for Output Ports. Hidden Data Products to
+which you do not have access should be redacted from the list. Only public information should be visible: ID, requester (so you know whom to talk to when reviewing), business reason etc.
+No the name of the Data Product, description ...
```
stijn-janssens:
```suggestion
Not the name of the Data Product, description ...
```

stijn-janssens (approved):

docs/adr/0022-data-product-visibility.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Option 2: Casbin resource-role visibility
+
+* **Good, because** it keeps visibility decisions in the same authorization system as all other data-product access.
+* **Good, because** it reuses the existing approved assignment flow for hidden products and avoids a second permission model.
+* **Good, because** discoverable products are represented as a public resource role, which fits the existing Casbin grouping model.
+* **Good, because** we can easily check if all routes have casbin enforcement through a test, instead of now where some routes have casbin enforcement and some don't.
+* **Good, because** doesn't repeat the same security loopholes that where present in Output Port visibility.
+* **Good, because** can be reused to redact information in list calls, such as Input Ports list or Output Port consumers.
+* **Neutral, because** the application must keep Casbin state synchronized with product visibility and assignment changes.
```
vlieven: I would classify this as bad, there is no good way to guarantee that these cannot drift out of sync, but the error rate will also be vanishingly small. However, certain admins that like to manipulate visibility directly in the DB will be out of luck, as the change will not carry over to casbin.

docs/adr/0022-data-product-visibility.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Option 2: Casbin resource-role visibility
+
+* **Good, because** it keeps visibility decisions in the same authorization system as all other data-product access.
+* **Good, because** it reuses the existing approved assignment flow for hidden products and avoids a second permission model.
+* **Good, because** discoverable products are represented as a public resource role, which fits the existing Casbin grouping model.
+* **Good, because** we can easily check if all routes have casbin enforcement through a test, instead of now where some routes have casbin enforcement and some don't.
+* **Good, because** doesn't repeat the same security loopholes that where present in Output Port visibility.
+* **Neutral, because** the application must keep Casbin state synchronized with product visibility and assignment changes.
+* **Bad, because** it introduces policy synchronization complexity and requires careful handling when a product changes from hidden to discoverable or vice versa.
```
HannesDS: How is this one different than the one above? I don't understand the difference between the two?

HannesDS (approved):
Agreed

docs/adr/0022-data-product-visibility.md
stijndehaes: Cleared up explanation

Issue #4162, opened 2026-08-27
Hidden data products implementation


stijndehaes:
Hidden data products adr is here: https://github.com/conveyordata/data-product-portal/blob/main/docs/adr/0022-data-product-visibility.md

Tasks:
- [x] Update our backend to support hidden data products in enforcer
- [x] Ensure a hidden data product can only contain restricted output ports
- [x] Add redacted version of hidden data products to show in: consumers list of output ports, input ports list of data product
- [ ] ~Ensure producers can add consumers, so people can add consumers to hidden data products~ Out of scope
- [x] Remove hidden data products from explorer, or add redacted version
- [ ] Remove hidden data products from mini explorer or add redacted version, if you have hidden consumer show redacted version
- [ ] Ensure people can create hidden data products in the UI
- [ ] Fix Output ports to ensure that they use the same mechanism as data products for authorization and filtering
- [x] Fix mcp, now it exposes hidden data products and output ports
- [ ] Should we allow migration from hidden to discoverable and vice versa -> currently we do not allow that

PR #4163, opened 2026-08-27
First implementation of hidden data products

stijndehaes:
This implements hidden data products in the backend but no API endpoints currently allow to create one. This is on purpose, development is cut into multiple smaller PR's.

In support of: #4162

Follow up PR's will:
- redact hidden data products in input ports, and consumers lists
- Allow creation of hidden data products
- Ensure hidden data products can only contain restricted output ports

### DIFF
--- a/backend/app/authorization/roles/model.py
+++ b/backend/app/authorization/roles/model.py
@@ -1,10 +1,12 @@
 import uuid
 
 from sqlalchemy import Column, Integer, SmallInteger, String
+from sqlalchemy import Enum as SAEnum
 from sqlalchemy.dialects.postgresql import ARRAY, UUID
-from sqlalchemy.orm import mapped_column
+from sqlalchemy.orm import mapped_column, validates
 
-from app.authorization.roles.schema import Prototype
+from app.authorization.roles.schema import Prototype, Scope
+from app.core.authz.actions import AuthorizationAction
 from app.database.database import Base
 from app.shared.model import BaseORM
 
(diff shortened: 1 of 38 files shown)


backend/app/authorization/roles/model.py
```diff
@@ -14,7 +16,34 @@ class Role(Base, BaseORM):
...
+            validate_strings=True,
+        ),
+        nullable=False,
+    )
     prototype = Column(SmallInteger, default=Prototype.CUSTOM)
     description = Column(String)
     permissions = Column(ARRAY(Integer))
+
+    @validates("permissions")
+    def validate_permissions(self, key: str, permissions: list[int]) -> list[int]:
```
stijn-janssens, line 34: Can't we solve this by including the permission in the seed and excluding it from the settings page in the UI? That way if someone intentionally removes it from the db permissions it is their own problem? Or would that make it too cumbersome? In any case I would exclude it from the UX if we readd it anyways

backend/app/authorization/service.py
```diff
@@ -34,8 +32,7 @@ def __init__(self, db: Session) -> None:
         self.authorizer = Authorization()
 
     def reload_enforcer(self) -> None:
-        removed = self._clear_casbin_table()
```
stijn-janssens, line 37: I am not following why the changes in this file happen.
Why is the clean and reload no longer needed? Why is the data product reader role the only being synced?

stijn-janssens: nm became clearer when checking out other files

backend/tests/factories/role.py
```diff
@@ -20,16 +20,26 @@ class Meta:
...
-        "random_element", elements=("global", "data_product", "dataset")
-    )
+    scope = factory.Faker("random_element", elements=("global", "dataset"))
     description = factory.Faker("text")
     permissions = factory.Faker(
         "random_elements", elements=list(map(int, AuthorizationAction)), unique=True
     )
+    #
+    # @staticmethod
+    # def _ensure_data_product_read_permission(permissions, scope):
```
stijn-janssens: comments still needed?

stijn-janssens (approved):
looks good !

backend/app/authorization/roles/model.py
stijndehaes: Well the seed doesn't solve anything right? If a user makes a new data product role, it won't include the new right so the new data product role won't work.

Btw do you mean or sample_data.sql as seed? Our customers don't use that right? But maybe I miss something.

backend/tests/factories/role.py
stijndehaes: Nope thanks for seeing this

backend/app/authorization/roles/model.py
stijn-janssens: no I meant seed in the migration. But fair point, I forgot about new roles.
But to be clear, it is not needed to add this as a checkbox in the UX of the roles then, right?

stijndehaes: Yes thanks for reminding me, I didn't actually check if it's visible somewhere.

stijndehaes: As expected they are not shown anywhere

PR #4165, opened 2026-08-27
Redaction of output port consumers

stijndehaes:
We now redact output port consumers list
and access reviews.
The code is made so the exporting schema
`AbstractDataProductInfo` redacts by default
unless a check is done. This is to ensure
we don't make mistakes.

In the front end we show blurred text with
a tooltip when we show redacted things to
the user.

In support of: #4162 

It currently is still impossible to requests them in the front end but slowly working there!

Request with hidden data product:
image (8)

image (9)

Hidden data product as consumer:
image (11)


stijn-janssens (approved):
Can we write a test that checks if all usages of DataProductInfo make correct use of the redaction?
So we can check if things are unredacted when necessary?

backend/app/data_products/output_ports/input_ports/service.py
```diff
@@ -208,6 +218,52 @@ def remove_output_port_as_input_port(
...
+        )
+
+    def get_consuming_data_products(
+        self, current_user: User, output_port_id: UUID, data_product_id: UUID
+    ) -> Sequence[OutputPortInputPort]:
+
+        output_port = self.db.scalar(
+            select(OutputPortModel)
+            .where(OutputPortModel.id == output_port_id)
+            .where(OutputPortModel.data_product_id == data_product_id)
```
HannesDS, line 243: Why is this needed?

stijndehaes: Good question, I should probably write this reasoning down.

Since we have all output port endpoints under the data product endpoints, a dataproduct ID is included in the URL. We need to add this to ensure that people have submitted the right data product id. This is to ensure we don't leave any security backdoors open later.
When we for example ensure that Data Product owners can change anything on Output Ports under that data product, we will check in authorisation the data product ID, if we don't include it in the query here this might result in people being able to edit random output port id's (or in this case request them)

PR #4186, opened 2026-08-31
Filter get data products

stijndehaes:
Ensure get data products is filtered
on visibility

In support of: #4162

### DIFF
--- a/backend/app/data_products/router.py
+++ b/backend/app/data_products/router.py
@@ -1,15 +1,13 @@
-from typing import Annotated, Sequence
+from typing import Sequence
 from uuid import UUID
 
 from fastapi import (
     APIRouter,
     BackgroundTasks,
     Depends,
-    Query,
     status,
 )
 from fastapi.responses import Response
-from pydantic.json_schema import SkipJsonSchema
 from sqlalchemy.orm import Session
 
 from app.abstract_data_product.input_ports.enums import InputPortStatus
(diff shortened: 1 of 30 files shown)


backend/app/data_products/router.py
```diff
@@ -573,15 +571,15 @@ def revoke_input_port_for_data_product(
 @router.get("")
 def get_data_products(
     db: Session = Depends(get_db_session),
-    filter_to_user_with_assigment: Annotated[
-        UUID | SkipJsonSchema[None], Query()
-    ] = None,
+    current_user_assigned: bool = True,
+    current_user: User = Depends(get_authenticated_user),
```
vlieven: Can these two parameters not be collapsed into `current_user: Optional[User]`, with the value being None if there is no current user assigned? Or do I misunderstand what `current_user_assigned` means?

vlieven: It seems my suggestion is incorrect, can we change `current_user_assigned` to `only_assigned`? Or even better add a `filter` or `scope` enum with the values `ALL` and `ASSIGNED`?

vlieven (changes requested):

backend/app/data_products/router.py
stijndehaes: I am thinking of:
```
assignment_filter: AssignmentFilter = AssignmentFilter.All
```
Second option: AssignmentFilter.ONLY_ASSIGNED

vlieven: Sounds good :)

backend/app/mcp/search.py
```diff
@@ -126,10 +133,15 @@ def search_data_products(
     query: Optional[str] = None,
     domain_id: Optional[str] = None,
     status: Optional[str] = None,
+    current_user_assigned: bool = False,
```
vlieven: This parameter is not used I think?

vlieven (approved):

backend/app/mcp/search.py
stijndehaes: Yes, removed now

Issue #4201, opened 2026-09-01
Set of environments for domain


HannesDS:
## Summary
[Brief explanation of the feature.](https://app.notion.com/p/datamindedbe/Set-of-environments-for-domain-3a093b0d1c7180cfae91cdff4e96cae6?v=2be93b0d1c718053913b000c0e7a6f32&source=copy_link)

PR #4203, opened 2026-09-01
feature: add domain environments capability

HannesDS:
Add a new table in the backend that links domains and environemnts. Add a field in the environments table that shows if it is global or not. Update the domain modal one edit and create to show two options, use global envs, or custom list.
Show the environemnts the domain is using in the table in the admin settings

- [x] Updated sample data so the new feature can be easily tested. -> Added a Demo Domain Environments domain and product to showcase what you have if you have a custom selection of environments on a domain. 

image

image
image

image

image

#4201


backend/app/configuration/environments/model.py
```diff
@@ -15,3 +15,4 @@ class Environment(Base, BaseORM):
     acronym = Column(String, unique=True)
     context = Column(String)
     is_default = Column(Boolean, default=False)
+    is_global = Column(Boolean, default=True, nullable=False)
```
vlieven, line 18: What is the difference between a global environment and one that is not? Global environments are automatically added to all domains?

HannesDS: Yes indeed, I added some information to the docs: https://github.com/conveyordata/data-product-portal/pull/4203/changes#diff-579df285d75e44a184ee28ea29dbfe3be00f5269766a907cbcb1e10c796063d7

backend/app/configuration/environments/service.py
```diff
@@ -17,3 +18,30 @@ def get_environments(self) -> Sequence[Environment]:
...
+                select(EnvironmentModel).filter(
+                    EnvironmentModel.id != environment_id,
+                    EnvironmentModel.is_global.is_(True),
+                )
+            )
+            if not other_global_exists:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="At least one environment must remain global",
+                )
```
vlieven: Why is it mandatory to have one global environment? I can imagine there are companies that want to have a split between domains, with no shared environment between them?

backend/app/data_products/technical_assets/service.py
```diff
@@ -98,11 +100,36 @@ def get_status_for(technical_mapping: TechnicalMapping):
...
+            for environment in self.db.scalars(
+                select(EnvironmentModel).where(EnvironmentModel.is_global.is_(True))
+            ).all()
+        }
+        for technical_asset in technical_assets:
+            domain_environments = technical_asset.owner.domain.environments
+            allowed_environment_ids = (
+                {environment.id for environment in domain_environments}
+                if domain_environments
+                else global_environment_ids
```
vlieven, line 117: Is this correct? I would expect that this is the union of global + specific domains instead of hiding the global environments if domain-specific ones are defined

frontend/src/components/data-access/data-access-tile-grid/data-access-tile-grid.tsx
```diff
@@ -55,8 +62,10 @@ export function DataAccessTileGrid({
                             key={dataPlatform.value}
                             dataPlatform={dataPlatform}
                             environments={getEnvironment(dataPlatform) ?? []}
-                            isDisabled={isDisabled || isLoading || isLoadingEnvironments || !canAccessData}
-                            isLoading={isLoading || isLoadingEnvironments}
+                            isDisabled={
```
vlieven: I think the loading state also prevents interaction (~ disabled), so I think you can exclude the loading states from the `isDisabled` parameter

backend/app/data_products/technical_assets/service.py
HannesDS: No that is actually what was in the story. Either global OR custom for now, no merging the two. Descibedi n the ticket.

backend/app/configuration/environments/service.py
HannesDS: I thought it would be weird to not have a single env, but in hindsight I don't think this requirement fixes anything or has any merit to it. So removed.

frontend/src/components/data-access/data-access-tile-grid/data-access-tile-grid.tsx
HannesDS: Fixed

backend/app/configuration/domains/service.py
```diff
@@ -50,15 +53,33 @@ def get_domain(self, id: UUID) -> GetDomainResponse:
...
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Invalid environment ids provided",
+            )
+        return list(environments)
+
     def create_domain(self, domain_create: DomainCreate) -> CreateDomainResponse:
-        domain = DomainModel(**domain_create.parse_pydantic_schema())
+        domain_schema = domain_create.model_dump(exclude={"environment_ids"})
+        environments = self._get_environments(domain_create.environment_ids)
```
stijn-janssens, line 69: Can you explain why you need to fetch and set the environments explicitly? 
Normally the SQLAlchemy mapping ensures that UUIDs are mapped to the actual entities and this code is obsolete

backend/app/configuration/environments/model.py
stijn-janssens, line 1: maybe the reason you need to map the environments explicitly is because there is no relationship with Domains defined in this file?

backend/app/configuration/domains/service.py
HannesDS: Well as I understand it environment_ids is a raw list[UUID] on the request schema, not a mapped attribute so it is not possible to "parse" that automatically in SQLAlchemy, or I don't know mechanisms that can't? Also the relationship is currently one-directional, and there is no link between Environment -> Domain yet. Is that bad convention? Should you directly implement it with backpopulates immediatly?

stijn-janssens: what is the advantage of doing it this way? It feels easier for me to just set up a proper mapping both ways and allow the entities to be fetched automatically?

backend/app/data_products/technical_assets/service.py
vlieven: Ok, I suppose we can still change this later if other users feel the same way

backend/app/configuration/domains/service.py
HannesDS: Well nothing that I know of. This is AI's respones:

There's no ORM mechanism that turns a list of bare UUIDs into list[EnvironmentModel]
  instances for a secondary= relationship; you have to fetch them. The same exact shape exists twice more in this codebase:

  - _get_access_modes in app/data_products/technical_assets/service.py:78-92 — identical select(...).where(id.in_(...)) + count-mismatch → HTTPException(400)
  - _get_tags in app/data_products/service.py:182-183 and technical_assets/service.py:71-76 — same idea, per-id via ensure_tag_exists
  
  So _get_environments isn't obsolete, it's the house style for "resolve incoming FK id list → entities" on a many-to-many, and it's also your only validation that a submitted environment id actually exists (400 vs. a silent no-op
  or an FK error).

  On bidirectionality, the codebase is split: Tag relationships (Tag.datasets/data_products/data_outputs) are bidirectional with back_populates, but the other two secondary= many-to-manys — TechnicalAsset.access_modes and
  OutputPort.access_modes — are one-directional, same as your Domain.environments. So one-directional isn't a convention violation here; it matches the access_modes precedent, not an outlier.

stijn-janssens: Apparently I was dreaming of things that don't exist 😅 I stand corrected

vlieven (approved):

stijn-janssens (approved):

PR #4204, opened 2026-09-01
Only private OPs can be created for hidden DPs

stijndehaes:
In support of: #4162

### DIFF
--- a/backend/app/data_products/output_ports/service.py
+++ b/backend/app/data_products/output_ports/service.py
@@ -34,6 +34,10 @@ from app.core.namespace.validation import (
     NamespaceValidator,
 )
 from app.data_products.model import (
+    DataProduct as DataProductModel,
+)
+from app.data_products.model import (
+    DataProductVisibility,
     ensure_data_product_exists,
 )
 from app.data_products.output_port_technical_assets_link.model import (
@@ -88,13 +92,16 @@ class OutputPortService:
         self.namespace_validator = NamespaceValidator(OutputPortModel)
         self.embedding_model = get_text_embedding_model()
 
-    def _ensure_data_product_not_deleting(self, data_product_id: UUID) -> None:
+    def _ensure_data_product_not_deleting(
+        self, data_product_id: UUID
+    ) -> DataProductModel:
         dp = ensure_data_product_exists(data_product_id, self.db)
         if dp.status == AbstractDataProductStatus.DELETING:
             raise HTTPException(
                 status_code=status.HTTP_409_CONFLICT,
                 detail=f"Data product '{dp.name}' is pending deletion and cannot be modified",
             )
+        return dp
 
     def get_access_durations(
         self, id: UUID, user: UserModel, data_product_id: Optional[UUID] = None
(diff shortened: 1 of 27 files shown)


stijndehaes:
~TODO: dataset update can change access type, we should also ensure we can't change it for hidden data products~ done

backend/app/data_products/output_ports/service.py
```diff
@@ -312,24 +319,42 @@ def _fetch_tags(self, tag_ids: Iterable[UUID] = ()) -> list[TagModel]:
         return tags
 
     def create_output_port(
-        self, data_product_id: UUID, dataset: CreateOutputPortRequest
+        self, data_product_id: UUID, create_output_port_request: CreateOutputPortRequest
     ) -> OutputPortModel:
-        self._ensure_data_product_not_deleting(data_product_id)
+        # Convert public to unrestricted.
+        if create_output_port_request.access_type == OutputPortAccessType.PUBLIC:
```
stijn-janssens: why do we even have PUBLIC still? I thought we changed into UNRESTRICTED?

stijndehaes: It's still in our code base. We can probably remove it, but didn't want to think about this in this PR. I just moved this old code to the right location

backend/app/data_products/output_ports/service.py
```diff
@@ -312,24 +319,42 @@ def _fetch_tags(self, tag_ids: Iterable[UUID] = ()) -> list[TagModel]:
         return tags
 
     def create_output_port(
-        self, data_product_id: UUID, dataset: CreateOutputPortRequest
+        self, data_product_id: UUID, create_output_port_request: CreateOutputPortRequest
     ) -> OutputPortModel:
-        self._ensure_data_product_not_deleting(data_product_id)
+        # Convert public to unrestricted.
```
HannesDS: Don't see the value of the comment

backend/tests/app/data_products/output_ports/test_router.py
```diff
@@ -130,7 +127,42 @@ def test_create_output_port_type_public(
...
+            role_id=role.id,
+        )
+        data_product_id = DataProductFactory(visibility=DataProductVisibility.HIDDEN).id
+
+        # Public not allowed
+        output_port_payload["access_type"] = OutputPortAccessType.PUBLIC.value
+        created_dataset = self.create_output_port(
+            client, data_product_id, output_port_payload
+        )
+        assert created_dataset.status_code == 400
```
HannesDS, line 148: Find it a bit weird there are 3 test (scenarios) in the same test. Also don't think the comments are needed.

backend/tests/app/data_products/output_ports/test_router.py
```diff
@@ -745,7 +779,7 @@ def test_update_dataset_duplicate_namespace(self, client):
 
         assert response.status_code == 400
 
-    def test_history_event_created_on_create_dataset(self, dataset_payload, client):
+    def test_history_event_created_on_create_dataset(self, output_port_payload, client):
```
HannesDS, line 770: Maybe also change the method name in this case. Maybe even worth doing on all the test names ctrl F.

HannesDS (changes requested):
Some questions

backend/app/data_products/output_ports/service.py
stijndehaes: Removed it and made a migration in pydantic

backend/tests/app/data_products/output_ports/test_router.py
stijndehaes: Removed comments, what would you rather see? Just 3 tests?

backend/tests/app/data_products/output_ports/test_router.py
stijndehaes: Well I just renamed the payload :) But not all the tests. I wouldn't blow up the scope of my PR with that cleanup, but maybe we can have AI refactor this in a seperate PR?

backend/app/data_products/output_ports/service.py
stijndehaes: Removed the public conversion

backend/app/database/alembic/versions/4c40a4ff5a7f_initialize_database.py
```diff
@@ -137,7 +137,7 @@ def upgrade() -> None:
         sa.Column(
             "access_type",
             sa.Enum(OutputPortAccessType),
-            default=OutputPortAccessType.PUBLIC,
+            default="public",
```
stijn-janssens, line 140: why put this on public to change into unrestricted by a later migration? Why not just start of with unrestricted?

stijn-janssens (approved):

backend/app/database/alembic/versions/4c40a4ff5a7f_initialize_database.py
stijndehaes: Maybe it doesn't matter, but I did not want to think about the implications:) The original migration set it to public, so I just keep it like it was.

backend/tests/app/data_products/output_ports/test_router.py
HannesDS: I think so, small brain me likes it when a single test fails, it only has a limited amount of scenario's and asserts to quicker understand and fix it. But not a breaking point, so fine to leave it as well.

backend/tests/app/data_products/output_ports/test_router.py
HannesDS: Agreed

HannesDS (approved):

PR #4205, opened 2026-09-01
Try making db clearing faster

stijndehaes:
Using truncate where possible


stijndehaes:
It's sometimes a little bit faster, it should mostly be faster in deleting a whole lot of data. I think it makes sense to keep

stijn-janssens (approved):

PR #4212, opened 2026-09-02
Use aws boto3 refresh session instead of custom one

stijndehaes:

### DIFF
--- a/backend/app/core/aws/refreshable_session.py
+++ b/backend/app/core/aws/refreshable_session.py
@@ -1,116 +1,65 @@
-from datetime import datetime
-from time import time
-from typing import Optional
-from uuid import uuid4
+from datetime import datetime, timedelta, timezone
+from typing import TypedDict
 
-import pytz
 from boto3 import Session
 from botocore.credentials import RefreshableCredentials
 from botocore.session import get_session
 
 
-class RefreshableBotoSession:
-    """
-    Boto Helper class which lets us create a
-    refreshable session so that we can cache the client or resource.
+class RefreshableSessionMetadata(TypedDict):
+    access_key: str
+    secret_key: str
+    token: str | None
+    expiry_time: str
 
-    Usage
-    -----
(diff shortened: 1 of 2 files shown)


backend/app/core/aws/refreshable_session.py
```diff
@@ -1,116 +1,58 @@
...
-                "token": response.get("SessionToken"),
-                "expiry_time": response.get("Expiration").isoformat(),
+def get_refreshable_session(
+    region_name=None,
+    profile_name=None,
+    sts_arn=None,
+    session_name="default-session",
+    session_ttl=3000,
+):
+    from boto3 import Session
```
vlieven: Why is it better to do the import here? I'd just put it on top with the rest.

backend/app/core/aws/refreshable_session.py
```diff
@@ -1,116 +1,58 @@
...
+
+    def refresh_credentials():
```
vlieven: Should this return a specialized return type (namedtuple or dataclass) instead of a dict?

HannesDS:
Just for my understanding, what is the reason of this PR, was it broken or?

backend/app/core/aws/refreshable_session.py
stijndehaes: I originally did this because importing boto3 is super slow, and I noticed that the app boots up slowly. 

But I probably should fix that in another PR. I'll remove this again. Since there are other boto3 imports that are not fixed

backend/app/core/aws/refreshable_session.py
stijndehaes: Probably I just did the minimal refactor necessary and this was how it was before. I will fix it

stijndehaes:
> Just for my understanding, what is the reason of this PR, was it broken or?

I noticed this, and saw an opportunity to remove some unneeded code. So I asked AI to fix things for me :)

stijn-janssens (approved):

Issue #4213, opened 2026-09-02
Add links to review access request modal


HannesDS:
## Summary
Add links to the requesting data product in the review modal. 

### Motivation
Reviewer of the access request cannot currently navigate to the requesting product from the review modal because the product (or exploration) is not linked. So now they need to approve/deny without looking at the product or going out of modal and searching for it. When this task is done, they can navigate from the modal directly to the requesting product.

PR #4214, opened 2026-09-02
improvement: Add navigation links to review-access-request modal cards

HannesDS:
image

#4213

### DIFF
--- a/frontend/src/components/pending-access-requests-modal/review-request-modal.tsx
+++ b/frontend/src/components/pending-access-requests-modal/review-request-modal.tsx
@@ -17,9 +17,9 @@ import {
 } from 'antd';
 import type { GlobalToken } from 'antd/es/theme/interface';
 import { addDays, isPast } from 'date-fns';
-import { type ReactNode, useState } from 'react';
+import { type CSSProperties, type ReactNode, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useNavigate } from 'react-router';
+import { Link } from 'react-router';
 import AccessMode from '@/components/access-modes/access-mode.component.tsx';
 import BlurredText from '@/components/blurred/blurred-text';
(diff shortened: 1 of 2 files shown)


frontend/src/components/pending-access-requests-modal/review-request-modal.tsx
```diff
@@ -420,7 +445,13 @@ export function ReviewRequestModal({ action, open, onClose, onAccept, onReject,
...
                 <Flex vertical gap="middle">
                     <Flex gap="small" align="stretch">
-                        <Card size="small" variant="outlined" style={{ flex: 3 }}>
+                        <Card
+                            size="small"
+                            variant="outlined"
+                            hoverable={Boolean(sourceLinkTo)}
+                            style={{ flex: 3, cursor: sourceLinkTo ? 'pointer' : 'default' }}
+                            onClick={sourceLinkTo ? () => navigate(sourceLinkTo) : undefined}
+                        >
```
stijndehaes: A nitpick of mine, but onClick is not the right way to do this.
The downside of onClick is that it's not an actual link, so you can't rightclick on it to copy it, or open it in a new tab.
In general I prefer to use an actual link `import { Link } from 'react-router-dom';`, around the card, can you try if that is easy?

Another thing, I assume that when you click on this your modal closes? SHould we maybe open a new tab by default?

stijndehaes (approved):
Approved, just check my comment, but it doesn't stop this improvement for me :)

stijn-janssens (approved):

frontend/src/components/pending-access-requests-modal/review-request-modal.tsx
HannesDS: Very nice comments! Thank you! Agreed on both.

stijndehaes (approved):
Approved, just check my comment, but it doesn't stop this improvement for me :)

PR #4215, opened 2026-09-02
Output port refactor

stijndehaes:
Currently a private output port can not
add consumers in the UI. So currently
there is no fixe needed to show a
private output port to it's consumers.
In the future we will need to decide
on what a private output port consumers can
see.

Also fixed missing update query stats permission

In support of: #4162

Limitations:
- ~Data product members can not see private output ports unless they have access to the output port. We should eventually fix this by ensuring data product roles can have automatic rights to all output ports? Maybe we should fix viewing them for now, what does the reviewer think?~ Fixed data product member with output port actions can see them
- When you consume a private output port you can not see it. But you currently also can't add consumers...


backend/tests/app/data_products/test_router.py
```diff
@@ -842,13 +842,6 @@ def test_retain_deleted_data_product_name_in_history(self, client: TestClient):
             == data_product.name
         )
 
-    def test_get_output_ports(self, client: TestClient):
```
stijndehaes, line 845: already tested in output ports router

stijn-janssens (approved):
Cool PR, I believe something is wrong with the merging of some of previous Alembic migrations. But apart from that, no comments

Issue #4185, opened 2026-08-31
Default lifecycle on data products


stijndehaes:
Currently the default lifecycle is filled in when we fetch the data products,
instead we should just have filled this in when we create a data product.
Now when default lifecycle changes a lot of data products might change their
default lifecycle, which might be unexpected.

Also this would make the get data products call faster.

THis is the current code in data products service:
```
    def get_data_products(
        self,
        filter_to_user_with_assigment: Optional[UUID] = None,
    ) -> Sequence[DataProductModel]:
        default_lifecycle = self.db.scalar(
            select(DataProductLifeCycleModel).filter(
                DataProductLifeCycleModel.is_default
            )
        )
        query = select(DataProductModel).options(
            selectinload(DataProductModel.tags).raiseload("*"),
            undefer(DataProductModel.input_port_count),
        )
        if filter_to_user_with_assigment:
            query = query.filter(
                DataProductModel.assignments.any(
                    user_id=filter_to_user_with_assigment,
                    decision=DecisionStatus.APPROVED,
                )
            )
        query = query.order_by(asc(DataProductModel.name))

        dps = self.db.scalars(query).unique().all()

        for dp in dps:
            if not dp.lifecycle:
                dp.lifecycle = default_lifecycle

        return dps
```

Or their might be something wrong, it's not optional in the schema, so not default lifecycle results in it being none.
So check that as well

PR #4219, opened 2026-09-03
fix: default lifecycle backend and frontend

### LINKED ISSUE #4202: Not possible to change default lifecycle
## Describe the bug
Currently it it not possible to change the default lifecycle thought the UI 

### Steps to reproduce

1. Make sure you are admin.
2. Go to the admin settings page and chekcout the Metadata tab. 
3. When scrolling to the bottom you can see the Lifcycle settings.
4. When creating a new lifecycle, you can't opt in for the `is_default` because you can only have a single default.
5. Even after creating and you want to uncheck the current default, it is not possible, and thus you can never switch. 

### Expected behavior
Either:
- Allowing to check the checkbox when creating a new lifecycle, automatically in the backend resolving that the previous is no longer the default. Give a small heads up that this is the case when checked.
- Not showing the is_default on creation, and only after creation allowing to switch the settings 

### Environment
/

### Additional context
Tested while in admin mode.

HannesDS:
backend now enforces a single default lifecycle on create/update (unsetting the previous one atomically), the admin checkbox is enabled except when locked on the currently-saved default (so it's impossible to end up with zero defaults from the modal), output ports get the current default assigned at creation if none is given, and the GET endpoints for data products/output ports no longer silently mutate-and-persist a default lifecycle onto records they merely read.

Also update frontend to disable the checkboxes

image

Closes #4202 and #4185


backend/app/configuration/data_product_lifecycles/service.py
```diff
@@ -27,12 +27,22 @@ def get_data_product_lifecycles(self) -> Sequence[DataProductLifeCyclesGetItem]:
             select(DataProductLifeCycleModel).order_by(DataProductLifeCycleModel.name)
         ).all()
 
+    def _unset_other_defaults(self, exclude_id: Optional[UUID] = None) -> None:
+        stmt = update(DataProductLifeCycleModel).where(
+            DataProductLifeCycleModel.is_default
+        )
+        if exclude_id is not None:
+            stmt = stmt.where(DataProductLifeCycleModel.id != exclude_id)
+        self.db.execute(stmt.values(is_default=False))
```
vlieven: The naming of the function and parameter can be a bit clearer IMO. Suggestion:

```suggestion
    def _ensure_single_default(self, default_lifecycle_id: Optional[UUID] = None) -> None:
        stmt = update(DataProductLifeCycleModel).where(
            DataProductLifeCycleModel.is_default
        )
        if default_lifecycle_id is not None:
            stmt = stmt.where(DataProductLifeCycleModel.id != default_lifecycle_id)
        self.db.execute(stmt.values(is_default=False))
```

stijndehaes (commented):
We can improve this via the DB.
We can add a unique index:

```
CREATE UNIQUE INDEX single_true_row_idx 
ON my_table (is_active) 
WHERE is_active IS TRUE;
```
This is an example that enforces that there can only be a single true value.

This does not ensure however that we have a true value.


Don't we also need a migration? To ensure that all output ports/data products have lifecycle filled in? And make the column non nullable?

backend/app/configuration/data_product_lifecycles/service.py
```diff
@@ -27,12 +27,24 @@ def get_data_product_lifecycles(self) -> Sequence[DataProductLifeCyclesGetItem]:
...
     def create_data_product_lifecycle(
         self, data_product_lifecycle: DataProductLifeCycleCreate
     ) -> CreateDataProductLifeCycleResponse:
         data_product_lifecycle = DataProductLifeCycleModel(
             **data_product_lifecycle.parse_pydantic_schema()
         )
+        if data_product_lifecycle.is_default:
+            self._ensure_single_default()
         self.db.add(data_product_lifecycle)
         self.db.commit()
```
stijndehaes, line 49: While you are at it, remove the commit here

backend/app/configuration/data_product_lifecycles/service.py
stijndehaes, line 62: Remove commit for things we are fixint :)

backend/tests/app/data_products/test_router.py
```diff
@@ -226,6 +226,32 @@ def test_get_data_product(self, client):
...
+        LifecycleFactory(is_default=True)
+        data_product = DataProductFactory(lifecycle=None)
+
+        response = self.get_data_product(client, data_product.id)
+        assert response.status_code == 200
+        assert response.json()["lifecycle"] is None
+
+        session.expire_all()
+        assert session.get(type(data_product), data_product.id).lifecycle_id is None
+
```
stijndehaes, line 241: I am unsure if we want to allow this case, don't we want to make lifecycle filled in with default on data product creation?

stijndehaes (changes requested):

backend/tests/app/data_products/test_router.py
HannesDS: Well it is a confusing test. What happened before was that if a lifecycle got dropped, it kept on being None. And when you did a get request, weirdly it got filled in with the default on a get. Which was weird. So this test I added to make sure that doesn't happen again. Which might not be clear from the test.

That said, the current is lifecycle_id: UUID is a required field in both DataProductCreate/DataProductUpdate schemas, and the frontend form is a required <Select> with no default value, so the user has to manually pick one every time,  nothing pre-selects the default lifecycle. This in contrast to output ports which silently default anyway. So is not exactly the same so maybe we would like to do the same here? Wdyt?

backend/app/configuration/data_product_lifecycles/service.py
HannesDS: Agreed.

backend/app/configuration/data_product_lifecycles/service.py
HannesDS: Agreed

HannesDS:
Closed, will discuss default lifecycle as whole.

PR #4220, opened 2026-09-03
Added ORM based filter for Data Products

stijndehaes:
In a follow up PR we need to fix both output ports and abstract data products.

In support of: #4162

### DIFF
--- a/backend/app/configuration/domains/model.py
+++ b/backend/app/configuration/domains/model.py
@@ -1,9 +1,9 @@
 import uuid
 from typing import TYPE_CHECKING
 
-from sqlalchemy import Column, String, func, literal_column, select, text
+from sqlalchemy import Column, String
 from sqlalchemy.dialects.postgresql import UUID
-from sqlalchemy.orm import Mapped, Session, deferred, relationship
+from sqlalchemy.orm import Mapped, Session, relationship
 
 from app.database.database import Base, ensure_exists
 from app.shared.model import BaseORM
@@ -21,13 +21,6 @@ class Domain(Base, BaseORM):
 
     abstract_data_products: Mapped[list["AbstractDataProduct"]] = relationship(
         "AbstractDataProduct", lazy="raise"
-    )
-
-    abstract_data_product_count = deferred(
-        select(func.count(literal_column("id")))
-        .select_from(text("abstract_data_products"))
-        .scalar_subquery(),
-        raiseload=True,
     )
 
 
--- a/backend/app/configuration/theme_settings/router.py
+++ b/backend/app/configuration/theme_settings/router.py
@@ -3,7 +3,7 @@ from sqlalchemy.orm import Session
 
 from app.core.authz import Action, Authorization
 from app.core.authz.resolvers import EmptyResolver
-from app.database.database import get_db_session
+from app.database.deps import get_db_session
 
 from .schema_response import ThemeSettings
 from .service import ThemeSettingsService
--- a/backend/app/core/authz/authorization.py
+++ b/backend/app/core/authz/authorization.py
@@ -12,7 +12,7 @@ from sqlalchemy.orm import Session
 from app.core.auth.auth import get_authenticated_user
 from app.data_products.model import DataProduct, DataProductVisibility
 from app.database import database
-from app.database.database import get_db_session
+from app.database.deps import get_db_session
 from app.settings import settings
 from app.users.schema import User
 from app.utils.singleton import Singleton
--- a/backend/app/data_products/model.py
+++ b/backend/app/data_products/model.py
@@ -1,10 +1,19 @@
 import enum
+import uuid
 from typing import TYPE_CHECKING
 
-from sqlalchemy import Column, ForeignKey, String, func, select
+from sqlalchemy import Column, ForeignKey, String, event, func, or_, select
 from sqlalchemy import Enum as SAEnum
 from sqlalchemy.dialects.postgresql import UUID
-from sqlalchemy.orm import Mapped, Session, column_property, mapped_column, relationship
+from sqlalchemy.orm import (
+    Mapped,
+    Session,
+    column_property,
+    mapped_column,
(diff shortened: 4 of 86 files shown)


backend/app/configuration/domains/model.py
```diff
@@ -23,13 +23,6 @@ class Domain(Base, BaseORM):
         "AbstractDataProduct", lazy="raise"
     )
 
-    abstract_data_product_count = deferred(
```
stijndehaes, line 26: Remove and replaced with a query specific in the service, I wanted to remove the raw text query as it can skip the orm based filter. We have a ruff check that does not allow text now

backend/app/configuration/theme_settings/router.py
```diff
@@ -3,8 +3,8 @@
 
 from app.core.authz import Action, Authorization
 from app.core.authz.resolvers import EmptyResolver
-from app.database.database import get_db_session
 
+from ...database.deps import get_db_session
```
vlieven: Why is this a relative import?

backend/app/core/authz/authorization.py
```diff
@@ -11,12 +11,12 @@
 
 from app.core.auth.auth import get_authenticated_user
 from app.database import database
-from app.database.database import get_db_session
 from app.settings import settings
 from app.users.schema import User
 from app.utils.singleton import Singleton
 
 from ...data_products.model import DataProduct, DataProductVisibility
+from ...database.deps import get_db_session
```
vlieven: Why are these relative imports?

backend/app/data_products/model.py
```diff
@@ -34,6 +43,65 @@ class DataProductVisibility(enum.Enum):
...
+    )
+
+
+def _is_user_admin(user_id: uuid.UUID):
+    from app.users.model import User
+
+    return (
+        select(User.id)
+        .where(User.id == user_id)
+        .where(User.admin_expiry.isnot(None))
```
vlieven: is the `.isnot(None)` check not redundant here?

backend/app/data_products/model.py
```diff
@@ -34,6 +43,65 @@ class DataProductVisibility(enum.Enum):
...
+    )
+
+
+def _is_system_account(user_id: uuid.UUID):
+    from app.users.model import User
+
+    return (
+        select(User.id)
+        .where(User.id == user_id)
+        .where(User.external_id == "systemaccount_bot")
```
vlieven: We should extract the magic string somewhere probably, so we centralize the ID.
Edit: unless this is the only place where it is defined, and in that case it is perfect.

backend/app/data_products/model.py
```diff
@@ -122,3 +190,55 @@ def ensure_data_product_exists(
...
+        desc.get("entity") is DataProduct
+        for desc in execute_state.statement.column_descriptions
+    ) or _statement_references_data_product(execute_state.statement)
+
+    if not is_data_product_query:
+        return
+
+    if user_id is None:
+        raise Exception(
+            "User id must be set or skip_data_product_visibility_filter must be True"
```
vlieven:
```suggestion
            "User id must be set when skip_data_product_visibility_filter is False"
```

backend/app/data_products/model.py
stijndehaes: Makes sense, I changed it. But not in the migration, as I don't want a rename to have an impact on that migration since we need a new migration then

backend/app/data_products/model.py
stijndehaes: It is redundant it's removed

backend/app/data_products/model.py
stijndehaes: Extracted

backend/app/data_products/model.py
stijndehaes: Change it to `is False or not set`

backend/app/database/database.py
```diff
@@ -112,4 +113,11 @@ def get_db_session():
         db.close()
 
 
-get_db_context = contextmanager(get_db_session)
+def get_unauthenticated_db_session() -> Generator[Session]:
```
vlieven: I read the function name as "get a DB session that is not authenticated", instead of "get a DB session without checking if there is an authenticated user", which is what I think it is used for? Maybe `get_system_db_session` or `get_raw_db_session` is better?

backend/app/mcp/deps.py
```diff
@@ -140,10 +127,16 @@ def get_mcp_authenticated_user(db: Session = Depends(get_db_session)) -> UserMod
     return user_model
 
 
+def get_user_db_session(user: User = Depends(get_mcp_authenticated_user)):
+    with get_db_session() as db:
+        db.info["current_user_id"] = user.id
+        yield db
```
vlieven: No equivalent "pop the user off the session" here?

backend/app/database/database.py
```diff
@@ -100,7 +100,8 @@ def get_one(self, primary_key_type: Column, primary_key_value: Any) -> BaseORM:
 Base = declarative_base()
 
 
-def get_db_session():
+@contextmanager
+def get_db_session() -> Generator[Session]:
```
vlieven: Maybe this should be just `db_session`? This might fit better with being used as a context manager and to distinguish it from `get_db_session` in `database/deps.py` which most callers should use?

vlieven (approved):
Left some comments regarding function naming. Functionally this seems sound.

backend/app/database/database.py
stijndehaes: `db_session_without_user`?

backend/app/mcp/deps.py
stijndehaes: Thanks for catching that, that is rather important

backend/app/database/database.py
stijndehaes: Decided on `get_system_db_session`

PR #4222, opened 2026-09-03
adr: Include external identities Groups and Machine Users

dixie-flatline999:
The ADR proposes how to include external identities that come from an identity provider; adds Groups and Machine Users, a SCIM interface and API endpoints. Uses also those identities for role assignment in Data Products and globally.


docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,248 @@
...
+Portal needs also a way to introduce these identities from an external identity provider and a way to expose them through the API.
+
+## Decision Drivers
+
+* Allow users, groups and machine users to receive Global and Data Product roles through the same assignment model.
+* Keep users as the only identity type that can authenticate and operate Portal.
+* Grant users permissions and Data Product visibility inherited from their groups.
+* Use a standard protocol for synchronizing users, groups and memberships.
+* Support machine users even when they are not available through SCIM.
+* Keep Portal as the only component with direct write access to its database.
```
stijndehaes:
```suggestion
```

This is a given, you can remove it

docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,248 @@
...
+* Provide provider-agnostic APIs for identities, memberships and Data Product role assignments.
+* Preserve existing user-linked workflow and audit relationships.
+* Avoid nested-group resolution, as the supported identity provider supplies flat memberships.
+* Keep the current user-specific Exploration model unchanged.
+
+## Considered Options
```
stijndehaes, line 31: I miss how we will map this to our authorization framework.

We need to ensure that a user via a group will be able to authenticate. I don't think this is hard. In casbin we can have role inheritance.
So a group can be a role that get's access to data product roles etc. And the user can get access to the group role when it's in the group

docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,248 @@
...
+
+For synchronization:
+
+* **Option A: Custom Portal synchronization API** — receive every identity type and membership through Portal-specific API operations.
+* **Option B: SCIM interface with a Machine User API** — receive users, groups and memberships through SCIM and receive machine users through a Portal-specific
+  API.
+
+## Decision Outcome
+
+**Chosen options:** *Option 1: Common Identity with separate subtype tables* and *Option B: SCIM interface with a Machine User API*.
```
stijndehaes: I would actually expect that we have our API's in portal. And that we have a SCIM extension that can use this. SCIM could even be a seperate service that people run no?

We need to support other use cases then SCIM to sync users, but they can be build on top of our normal API

docs/adr/0023-external-identities-group-based-auth-model.png
stijndehaes, line 1: Can we maybe include this as a mermaid diagram instead. This takes up less space and can be edited after the fact :) I let AI help me, and I think this is correct no?:
```mermaid
erDiagram
    identities {
        string id PK
        string type "USER | GROUP | SP"
    }

    group_membership {
        string group_id PK, FK
        string member_id PK, FK
    }

    users {
        string id PK, FK
        string email
        string external_id
        string first_name
        string second_name
    }

    service_principals {
        string id PK, FK
        string display_name
        string external_id
    }

    groups {
        string id PK, FK
        string display_name
        string external_id
    }

    identities ||--o| users : "1:0..1"
    identities ||--o| service_principals : "1:0..1"
    identities ||--o| groups : "1:0..1"
    identities ||--o{ group_membership : "member_id"
    groups ||--o{ group_membership : "group_id"
```

docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,248 @@
...
+requested_by_id  → users.id
+decided_by_id    → users.id
+```
+
+Existing Global and Data Product role assignments must be migrated by creating an Identity row for every existing user and replacing their assignment `user_id`
+references with the corresponding `identity_id`.
+
+Explorations remain associated with users through their existing ownership model. Their relationships are not migrated to `identities.id` by this decision.
+
+### Effective authorization and visibility
```
stijndehaes: Ok so here comes the authorization, for me it's ok to just remove this and have a bullet point that we will use casbin group inheritance in the Option 1 pros and cons

docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,248 @@
...
+  },
+  "decision": "approved"
+}
+```
+
+Embedding the identity avoids an additional lookup while preserving the original assignment.
+
+Global role assignments are not added to the external read API because they only affect authorization inside Portal.
+
+## Consequences
```
stijndehaes: Let's remove this, it's not part of our standard template makes things longer. All important things should have been said already

stijndehaes (changes requested):

docs/adr/0023-external-identities-group-based-auth.md
dixie-flatline999: Absolutely, used AI to assist on writing and this one slipped my review...

docs/adr/0023-external-identities-group-based-auth.md
dixie-flatline999: Regarding the API's, I rephrased it, as the idea is to implement CRUD operations for every identity type.

About SCIM, the proposal is that there are SCIM endpoints and API endpoints, but internally they will behave the same -SCIM complies with the interface but in the end it is just Create/Update/Delete operations-. This way there is no need to add another running service.

Portal would be exposing:
```
/api/v2/users
/scim/v2/users
```

Added `v2` on the SCIM path as technically is in v2, can be removed tho...

docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,235 @@
...
+Identity → Global role
+Identity → Data Product role
+```
+
+Fields that represent assignment targets reference identities, while fields that represent Portal actors remain attached to users:
+
+```text
+identity_id      → identities.id
+requested_by_id  → users.id
+decided_by_id    → users.id
```
HannesDS, line 164: I assume that this will never be done trough a service principal?

docs/adr/0023-external-identities-group-based-auth.md
```diff
@@ -0,0 +1,235 @@
...
+A common Identity provides one target for Global and Data Product role assignments while keeping the attributes and behaviour of users, groups and machine users
+separated. Only users can authenticate, request or approve changes and act inside Portal, while every identity type can receive roles and both users and machine
+users can belong to groups.
+
+Portal will implement a SCIM interface for users, groups and group memberships. Machine users and their memberships will be created and updated through Portal 
+API operations until they can be synchronized through SCIM. Both entry points will use the same internal identity services and persistence model.
+
+### Confirmation
+
+* The database contains `identities`, `users`, `groups`, `machine_users` and `group_memberships`.
```
HannesDS, line 56: Is there the need of the distinction between a user and an agent that does an action on behalf a of a user and the rights that user/agent has that is acting on behalf of a user. Is that something that needs to be considered or out of scope of this adr?

docs/adr/0023-external-identities-group-based-auth.md
dixie-flatline999: That's the point of this clarification. Any identity type can have a role in a data product, but only users can be assigned for auditing i.e., as a user cannot login in behalf of a group and a SP won't use the Portal itself

docs/adr/0023-external-identities-group-based-auth.md
dixie-flatline999: We don't do that distinction now and I don't think there is any plan to do it at the moment

stijndehaes (approved):

stijn-janssens (approved):

stijn-janssens:
Can you quickly check the failing pre-commit hooks on your ADR? It's mainly about file styling. If you have issues with it reach out and I'd be happy to help

PR #4237, opened 2026-09-04
A data product role has output ports rights

stijndehaes:
Ensure that a data product role can give
rights to it's output ports.

In support of: #4162

Adds a manage output ports to the data product role.
image

### DIFF
--- a/backend/app/data_products/output_port_technical_assets_link/service.py
+++ b/backend/app/data_products/output_port_technical_assets_link/service.py
@@ -7,11 +7,15 @@ from fastapi import HTTPException, status
 from sqlalchemy import asc, or_, select
 from sqlalchemy.orm import Session
 
+from app.authorization.role_assignments.data_product.model import (
+    DataProductRoleAssignment as DataProductRoleAssignmentModel,
+)
 from app.authorization.role_assignments.enums import DecisionStatus
 from app.authorization.role_assignments.output_port.model import (
     DatasetRoleAssignment as DatasetRoleAssignmentModel,
 )
 from app.core.authz import Action, Authorization
+from app.data_products.model import DataProduct as DataProductModel
 from app.data_products.output_port_technical_assets_link.model import (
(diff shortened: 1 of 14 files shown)


backend/app/data_products/output_port_technical_assets_link/service.py
```diff
@@ -156,10 +160,19 @@ def get_user_pending_actions(
...
+                            )
+                        ),
+                        DataOutputDatasetAssociationModel.output_port.has(
+                            OutputPortModel.data_product.has(
+                                DataProductModel.assignments.any(
+                                    DataProductRoleAssignmentModel.user_id == user.id
+                                )
+                            )
+                        ),
                     )
```
stijndehaes, line 176: Current gap is there that we do not check the action, should we?

stijndehaes:
This conflicts with:
https://github.com/conveyordata/data-product-portal/pull/4215

COnverted that one to draft as I think this is the first step to improving things

backend/app/data_products/output_port_technical_assets_link/service.py
stijn-janssens: Think it is fine for now? It's only used for linking TA's to OPs?

stijn-janssens (approved):
looks good, the one remark I think is currently also the case. So it does not make it worse. But maybe keep it documented somewhere in the tech debt

backend/app/data_products/output_port_technical_assets_link/service.py
stijndehaes: true

vlieven: Ideally we do, otherwise the actions are sort-of pointless

stijndehaes: I'll make a seperate ticket to fix this :)

stijndehaes: https://github.com/conveyordata/data-product-portal/issues/4239

PR #4240, opened 2026-09-04
Remove commits from tests

stijndehaes:
We should never do commits ourselves,
in the end I want tests in a transaction that
can be rolled back.


backend/tests/factories/technical_asset.py
```diff
@@ -24,7 +24,7 @@ class Meta:
     @factory.post_generation
     def platform_id(self, create, extracted, **kwargs):
         self.platform_id = self.service.platform_id
-        test_session.commit()
+        # test_session.flush()
```
vlieven: Commented?

stijn-janssens (approved):

PR #4262, opened 2026-09-08
ORM filter for output port

stijndehaes:
In support of: #4162

### DIFF
--- a/backend/app/abstract_data_product/background_tasks.py
+++ b/backend/app/abstract_data_product/background_tasks.py
@@ -3,7 +3,6 @@ from datetime import datetime, timezone
 
 from sqlalchemy import select
 
-from app.abstract_data_product.model import AbstractDataProduct
 from app.core.logging import logger
 from app.data_products.status import AbstractDataProductStatus
 from app.database.database import SessionLocal
@@ -13,6 +12,8 @@ STUCK_THRESHOLD_SECONDS = 3600  # warn after 1 hour in DELETING
 
 
 async def check_stuck_deletions() -> None:
+    from app.abstract_data_product.model import AbstractDataProduct
+
     while True:
         try:
             with SessionLocal() as db:
--- a/backend/app/settings.py
+++ b/backend/app/settings.py
@@ -68,6 +68,7 @@ class Settings(BaseSettings):
     # Authorizer
     AUTHORIZER_CACHE_SIZE: int = 128
     AUTHORIZER_STARTUP_SYNC: bool = True
+    AUTHORIZER_AUTOLOAD_ENABLED: bool = True
     AUTHORIZER_AUTOLOAD_INTERVAL: int = 30  # seconds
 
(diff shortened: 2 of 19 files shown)


backend/app/abstract_data_product/background_tasks.py
```diff
@@ -13,6 +12,8 @@
 
 
 async def check_stuck_deletions() -> None:
+    from app.abstract_data_product.model import AbstractDataProduct
```
stijn-janssens, line 15: why the import move?

backend/app/settings.py
```diff
@@ -68,6 +68,7 @@ class Settings(BaseSettings):
     # Authorizer
     AUTHORIZER_CACHE_SIZE: int = 128
     AUTHORIZER_STARTUP_SYNC: bool = True
+    AUTHORIZER_AUTOLOAD_ENABLED: bool = True
```
stijn-janssens, line 71: Why introduce this setting? Is there ever a case where we want this false?

backend/tests/conftest.py
```diff
@@ -36,6 +37,7 @@
 def setup_and_teardown_database():
     from app.db_tool import init  # noqa: E402
 
+    settings.AUTHORIZER_AUTOLOAD_ENABLED = False
```
stijn-janssens, line 40: which tests break if this is enabled?

stijn-janssens (commented):
just some clarifying questions

backend/app/settings.py
stijndehaes: Tests, basically in tests we only run a single enforcer anyway, I was debugging some DB connection things, and casbin queries always were in the way :P

backend/tests/conftest.py
stijndehaes: They don't but a lot of extra queries are being done, which is super annoying when trying to debug the queries I was debugging :)

backend/app/abstract_data_product/background_tasks.py
stijndehaes: I think it was because of a cyclical import, but that might not be the case anymore, will check

stijndehaes: It's because of a circular import, I can probably fix it by moving authz/db_utils.py in another modele. The init file of that model breaks it, perhaps authz/db/utils.py?

stijn-janssens: I'll let you make the call on whether this import is uglier or the new file positioning is ;)

stijndehaes: So moving the files doesn't work, since in authorization in the init file certain imports happen, and they always happen. This small change is the easist fix.

I also tried making dataproduct in some part a nice type hint, and move the visibility enum, but in the end the issue is that init file in authorization

stijn-janssens (approved):

PR #4291, opened 2026-09-11
Add CI compile script

stijn-janssens:

### DIFF
--- a/.github/workflows/docs-workflow.yml
+++ b/.github/workflows/docs-workflow.yml
@@ -3,7 +3,6 @@ on:
   pull_request:
     paths:
       - 'docs/**'
-      - '.github/workflows/docs-workflow.yml'
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-workflow.yml
```diff
@@ -0,0 +1,28 @@
...
+permissions:
+  contents: read
+
+jobs:
+  test-docs:
+    name: Test docs build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
HannesDS, line 17: It seems like you would need to manually update the version to get this workflow to work no? Or does it get updated automatically in another flow somewhere?

HannesDS (changes requested):

.github/workflows/docs-workflow.yml
stijn-janssens: not following completely. if you are talking about the version of the checkout action this is managed by dependabot and indeed gets automatically updated

HannesDS (approved):

PR #4292, opened 2026-09-11
Adr/0024 dynamic plugin system

HannesDS:

if you are curious about how AI would implement it, there is a spike at `spike/0024-dynamic-plugin-system ` which implements a spike version of the feature.

### DIFF
--- a/docs/adr/0024-dynamic-plugin-system.md
+++ b/docs/adr/0024-dynamic-plugin-system.md
@@ -1,77 +1,77 @@
 # Dynamically Loadable Plugin System
 
-**Status:** Proposed — draft for iteration, not yet accepted.
+**Status:** Proposed - draft for iteration, not yet accepted.
 
 ## Context and Problem Statement
 
-The portal already lets data products connect to external platforms (S3, Snowflake, Postgres, Databricks, and others) through a plugin mechanism. Today, adding a brand-new one of these means: writing new code, creating a brand-new database table just for that one integration, writing a database migration to create that table, and editing a couple of internal registration lists — all inside the core project. Because all of that lives in the core project, adding an integration always means submitting a change to it (or forking it).
+The portal already lets data products connect to external platforms (S3, Snowflake, Postgres, Databricks, and others) through a plugin mechanism. Today, adding a brand-new one of these means: writing new code, creating a brand-new database table just for that one integration, writing a database migration to create that table, and editing a couple of internal registration lists - all inside the core project. Because all of that lives in the core project, adding an integration always means submitting a change to it (or forking it).
 
-A second, less visible problem sits underneath that. Each integration's basic identity — its name, its icon, which vendor it belongs to, and what fields it needs — is currently written down **twice**: once in the code that implements it, and once as rows that have to be separately created in the database. Nothing keeps these two copies in sync automatically; they only agree because whoever set both of them up happened to spell everything identically. When they don't match exactly, the integration's tile still shows up in the interface, but trying to actually use it fails, without a clear reason why.
+A second, less visible problem sits underneath that. Each integration's basic identity - its name, its icon, which vendor it belongs to, and what fields it needs - is currently written down **twice**: once in the code that implements it, and once as rows that have to be separately created in the database. Nothing keeps these two copies in sync automatically; they only agree because whoever set both of them up happened to spell everything identically. When they don't match exactly, the integration's tile still shows up in the interface, but trying to actually use it fails, without a clear reason why.
 
 A third, related problem is how the portal stores per-environment infrastructure details (which storage buckets exist in "dev" vs. "production", which credentials apply where). This information is currently spread across several separate database tables that have grown into a genuinely complicated arrangement, with some facts (like a shared account identifier) effectively repeated across places that are hard to keep in sync with each other.
 
 One concrete request is driving this redesign: a customer needs to add their own custom integration without submitting a change to the core project and without forking it.
 
-**Core question: how should a plugin's code run, and how should a plugin's information be modeled and stored, so that adding one never requires a database migration and never requires a change to the core project — including for an integration that needs to remember real, structured settings?**
+**Core question: how should a plugin's code run, and how should a plugin's information be modeled and stored, so that adding one never requires a database migration and never requires a change to the core project - including for an integration that needs to remember real, structured settings?**
 
 ## Decision Drivers
 
-* Adding or changing an integration should never require a database migration — including one that needs real, persisted settings.
+* Adding or changing an integration should never require a database migration - including one that needs real, persisted settings.
 * Adding an integration should never require submitting a change to the core project, or maintaining a fork of it.
-* Whatever a plugin author has to do to add an integration should be as light as reasonably possible — dropping something into a folder is preferable to having to build and run a separate always-on service.
+* Whatever a plugin author has to do to add an integration should be as light as reasonably possible - dropping something into a folder is preferable to having to build and run a separate always-on service.
 * Keep what already works well: form fields are already generated generically from whatever fields an integration declares, and the human-readable summary text shown to users is already filled in generically using those same fields, by name. Any redesign should build on this, not throw it away.
-* Where today's way of storing an integration's identity and its environment-specific configuration is itself the cause of duplication or confusion, it is fair to change it — but a full, unrelated rename or redesign of every related concept is not the goal here.
+* Where today's way of storing an integration's identity and its environment-specific configuration is itself the cause of duplication or confusion, it is fair to change it - but a full, unrelated rename or redesign of every related concept is not the goal here.
 * Whatever is decided needs a realistic, low-disruption way to move the handful of integrations that already exist onto it, rather than leaving them permanently on a separate, older system.
 * Simplicity beats defensiveness: given every customer runs their own separate, self-hosted copy of the portal, prefer the simplest option that works over one that defends against risks that don't apply to this deployment model.
-* What gets loaded into the running portal should always be something an image author named on purpose — never something that got in automatically because it happened to be installed.
+* What gets loaded into the running portal should always be something an image author named on purpose - never something that got in automatically because it happened to be installed.
 
 ## Considered Options
 
 ### Where the plugin's code runs
 
 #### Option 1: Build it into the core project (today's situation)
-Rejected — this is the exact problem being solved.
+Rejected - this is the exact problem being solved.
 
 #### Option 2: Ship it as a separately installable package
 A real, versioned package (built the same way as any other Python package here), installed into the customer's own image. No fork needed, but it still means rebuilding and redeploying the customer's own copy every time. Two ways the portal could learn it's there:
 
-* **2a — discovered automatically**, the way Python packages can advertise themselves through a standard registration mechanism. Costs nothing to set up, but it's a real predictability problem: anything already installed — including some dependency of a dependency, never chosen on purpose — could advertise itself the same way and get loaded into the running portal, with nothing in the Dockerfile naming it.
-* **2b — discovered by explicit name**: the image author writes down, in one setting, exactly which packages/modules to load. Nothing loads that wasn't named on purpose. Costs one extra line to maintain, and it's possible to forget to add it after installing the package, which fails quietly (the package is present, but never loaded) rather than loudly.
+* **2a - discovered automatically**, the way Python packages can advertise themselves through a standard registration mechanism. Costs nothing to set up, but it's a real predictability problem: anything already installed - including some dependency of a dependency, never chosen on purpose - could advertise itself the same way and get loaded into the running portal, with nothing in the Dockerfile naming it.
+* **2b - discovered by explicit name**: the image author writes down, in one setting, exactly which packages/modules to load. Nothing loads that wasn't named on purpose. Costs one extra line to maintain, and it's possible to forget to add it after installing the package, which fails quietly (the package is present, but never loaded) rather than loudly.
 
 #### Option 3: Drop it into a scanned folder, loaded in-process
-The portal scans a folder on startup and loads a plugin's file directly into its own running program. Simple, but it gives up two things a real package would have: nowhere to declare the plugin's own dependencies (a customer has to install those by hand, disconnected from the plugin that needs them), and no version to pin or roll back — rolling one back means restoring the right files, not changing a number.
+The portal scans a folder on startup and loads a plugin's file directly into its own running program. Simple, but it gives up two things a real package would have: nowhere to declare the plugin's own dependencies (a customer has to install those by hand, disconnected from the plugin that needs them), and no version to pin or roll back - rolling one back means restoring the right files, not changing a number.
 
 #### Option 4: Drop it into a scanned folder, run as a separate program per call
 Instead of loading the dropped-in file into its own program, the portal treats it as its own small, separate program, and runs a fresh copy of it each time it needs something, exchanging only plain, simple information back and forth. This allows a plugin to be written in any language at all, and a mistake in one plugin can't directly corrupt the portal's own running program the way it could with Option 3.
 The trade-off: starting a fresh small program for every single request is somewhat slower and more work to build, and it still offers no protection against code that is deliberately malicious, only against accidental interference. This idea fits a folder of loose files; it doesn't naturally extend to an installed package that gets imported by name (Option 2), since there the whole point is that Python already loads it directly.
 
 #### Option 5: Run it inside the portal, but in a locked-down sandbox
-Cannot touch the filesystem, network, or database unless explicitly allowed to. This means a plugin author never has to run their own server, while genuinely bad code still cannot cause real damage. The real cost: a plugin has to be written in, or compiled into, a specific portable, sandboxable format, and the portal needs new infrastructure able to run that format — a real, non-trivial addition.
+Cannot touch the filesystem, network, or database unless explicitly allowed to. This means a plugin author never has to run their own server, while genuinely bad code still cannot cause real damage. The real cost: a plugin has to be written in, or compiled into, a specific portable, sandboxable format, and the portal needs new infrastructure able to run that format - a real, non-trivial addition.
 
 #### Option 6: Run it entirely on a separate server the plugin author operates
 The portal only ever sends it requests over the network. Two flavors: **6a**, where that server also decides exactly what the screen looks like (a fully custom mini-screen embedded into the page); or **6b**, where that server only answers questions, and the screen itself stays a plain description of fields and labels that the portal renders the same generic way it already does today.
 
 #### Option 7: Build a public marketplace
-Let people browse and install plugins written by others, with some form of review before something gets listed. This is not really a way of running code by itself — it's a way of finding it, and it could sit on top of any of the options above.
+Let people browse and install plugins written by others, with some form of review before something gets listed. It's a way of finding plugin code, and it could sit on top of any of the options above.
 
 ### Where is the basic metadata of the plugin stored
 
 This is the basic identity of an integration: its name, its icon, and the list of fields it needs.
-Today, this is defined twice — once by the plugin's own code, and once by rows that must be separately, manually created in the database, kept in sync only by both sides spelling everything identically.
+Today, this is defined twice - once by the plugin's own code, and once by rows that must be separately, manually created in the database, kept in sync only by both sides spelling everything identically.
 
 #### Option 1:  Portal code as well as database row (as today)
-Rejected — this is the exact duplication problem being fixed.
+Rejected - this is the exact duplication problem being fixed.
 
 #### Option 2: Only keep the database row.
 These filled in by an admin through a form or an API call, with no matching copy of that information anywhere in a plugin's own code.
 
 #### Option 3: Store externally and import at runtime
 It carries a small, plain description of its own name, icon, and fields, sitting right alongside its own code.
-The portal reads this description at the same moment it discovers the plugin's code, and that description is the *only* place this information exists — nobody has to separately re-type it anywhere. An extension or plugin ships its own description together with its own code, instead of a separate person entering the same facts into a form by hand.
+The portal reads this description at the same moment it discovers the plugin's code, and that description is the *only* place this information exists - nobody has to separately re-type it anywhere. An extension or plugin ships its own description together with its own code, instead of a separate person entering the same facts into a form by hand.
 
 ### Where a technical asset's configured values are stored
 
-Not "this integration exists," but "this particular data product's use of it has these exact values filled in."
+This is about the values one data product fills in for its use of an integration, separate from whether the integration itself exists.
 Today, every single integration type gets its own dedicated table in the database purely to hold this, as a result adding one requires a database migration.
 
 #### Option 1: Keep a dedicated table per integration type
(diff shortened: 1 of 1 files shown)


docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+plugins are used in a manner to translate portal concepts (Technical Assets) to an external system or link them to something that holds up in the "real world". It is not the job to actually create these resources (this is part of the provisioners' job). So in some way or form, the plugins translate the Portal core resources (technical asset/dataproducts) into real world concepts.
+However, since the provisioner works separately from that, which is the resource that does both things again, it translates events, and actually creates or provisions those resources. In other words there is duplicate responsibility shared among concepts which needs to be cleared out.
+The same holds for the concepts we use in Portal core such as `Platforms` and `PlatfomServices`, which currently are used in the plugin concept, but are not in the provisioner, which is a bit contradictory.
+
+Important that these are NOT part of this ADR. These are follow-up work.
+
+
+For reference, I use Portal Core to reference the source code in the main repo and plugin as the dynamically loadable plugin.
```
vlieven, line 18: A bit confusing to read, I don't know who "I" is supposed to refer to here. Maybe clearer to write ADRs from a 3rd person view

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Decision Drivers
+
+* Adding or changing an integration should never require a database migration to portal core to any of the core project's own, non-plugin-specific tables.
+* Adding an integration should never require submitting a change to the core project, or maintaining a fork of it.
+* The portal should assist with migrations when a plugin needs to change its own configuration, rather than leaving that entirely to the plugin author.
+* Whatever a plugin author has to do to add an integration should be as light as reasonably possible, without giving up the flexibility to add new functionality.
+* Whatever is decided needs a realistic, low-disruption way to move the handful of integrations that already exist onto it, rather than leaving them permanently on a separate, older system.
+* Simplicity beats defensiveness: given every customer runs their own separate, self-hosted copy of the portal, prefer the simplest option that works over one that defends against risks that don't apply to this deployment model.
+* A plugin should be usable as soon as its package is installed - the portal discovers it automatically, without a separate step to name it.
```
vlieven, line 28: This sounds like it implies hot reloading, we want to enable packaging plugins into docker containers, not installing them on the fly

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+* For a first version, this should be built on top of the platform, platform services, and environments data model as it exists today - or simply not use it at all. Redefining that model completely is out of scope here, but this decision shouldn't make it harder to change later.
```
vlieven, line 29: I'd argue redefining the model is the first step towards converging to a simple interface

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+As this needs some work and is currently not exactly clear what the data model should be and is not.
+* Still out of scope: the concrete data model a dynamically loaded plugin's technical asset settles into long-term (tags/access-mode compatibility rules specific to a plugin, search indexing, and similar). Two things *are* now decided and implemented, narrowing this: (1) see "How the API describes a plugin's configuration" - a dynamically loaded plugin's technical asset is created and read through the same `/v2/data_products/{id}/technical_assets` endpoints as every other type, not a separate plugin-specific API; and (2) what a plugin is allowed to say about its own fields for UI rendering purposes is deliberately small - see "The plugin's field vocabulary, and who owns turning it into a form" below - not the richer `UIElementMetadata` model the built-in types can use.
+* An admin interface for managing integrations. Integrations (including the ones with no code) remain reachable only through the underlying data/API for now, matching how they're exposed to the rest of the interface today.
```
vlieven, line 35: What does this mean?

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+* Attaching integrations beyond data products. Integrations continue to attach to a data product's technical assets only.
```
vlieven, line 36: This is a weird way to say "only support custom technical assets"

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Decision Outcome
+- Where the plugin's code runs: an installable Python package.
+- Loading the python packages inside portal core: Option 1 - automatic discovery via Python entry points.
+- How the portal recognizes a plugin's shape: Option 2 - an ABC (`TechnicalAssetPlugin`), living in the SDK's `sdk.plugins` package, that a plugin class inherits from.
+- Where and how is the configuration of the plugin stored: Option 2 - a dedicated table per plugin, migrated by the plugin's own Alembic revisions.
+- How the API describes a plugin's configuration: Option 1, with the published contract describing configuration generically for every type, served through the existing technical-asset endpoints rather than a dedicated plugin-only API.
+- The plugin's field vocabulary, and who owns turning it into a form: Option 1 - a minimal, string-only vocabulary (`name`, `label`, `required`, an optional regex `pattern`), adapted by the portal into the existing `UIElementMetadataResponse` shape.
```
vlieven, line 145: Does that mean that our existing plugins will be migrated to the minimal vocabulary too? Or does this mean that we have a functional split between first- and third-party plugins?

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+
+**Core question:** how should a plugin's code run, and how should a plugin's information be modeled and stored, so that adding one never requires a change to the core project?
+
+Next to that I also want to highlight some context of what is out of scope and to think about those next steps. Currently,
+plugins are used in a manner to translate portal concepts (Technical Assets) to an external system or link them to something that holds up in the "real world". It is not the job to actually create these resources (this is part of the provisioners' job). So in some way or form, the plugins translate the Portal core resources (technical asset/dataproducts) into real world concepts.
+However, since the provisioner works separately from that, which is the resource that does both things again, it translates events, and actually creates or provisions those resources. In other words there is duplicate responsibility shared among concepts which needs to be cleared out.
```
stijndehaes: Love this analysis, there is indeed a double translation going on

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
+# Dynamically Loadable Plugin System
+
+**Status:** Proposed - draft for iteration, not yet accepted.
+
+## Context and Problem Statement
+The portal already lets data products connect to external platforms (S3, Snowflake, Postgres, Databricks, and others) through a plugin mechanism. Today, adding a brand-new one of these means: writing new code, creating a brand-new database table just for that one integration, writing a database migration to create that table, and editing a couple of internal registration lists - all inside the core project. Because all of that lives in the core project, adding an integration always means submitting a change to it (or forking it).
```
stijndehaes:
```suggestion
Data Product Portal already lets data products connect to external platforms (S3, Snowflake, Postgres, Databricks, and others) through a plugin mechanism. Today, adding a brand-new one of these means: writing new code, creating a brand-new database table just for that one integration, writing a database migration to create that table, and editing a couple of internal registration lists - all inside the core project. Because all of that lives in the core project, adding an integration always means submitting a change to it (or forking it).
```

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+Important that these are NOT part of this ADR. These are follow-up work.
+
+
+For reference, I use Portal Core to reference the source code in the main repo and plugin as the dynamically loadable plugin.
+
+## Decision Drivers
+
+* Adding or changing an integration should never require a database migration to portal core to any of the core project's own, non-plugin-specific tables.
+* Adding an integration should never require submitting a change to the core project, or maintaining a fork of it.
+* The portal should assist with migrations when a plugin needs to change its own configuration, rather than leaving that entirely to the plugin author.
```
stijndehaes:
```suggestion
* Data Product Portal should assist with migrations when a plugin needs to change its own configuration, rather than leaving that entirely to the plugin author.
```

docs/adr/0024-dynamic-plugin-system.md
stijndehaes:
```suggestion
* A plugin should be usable as soon as its package is installed - the portal discovers it automatically, without a separate step to name it. It won't scan at runtime though, things need to be available a boot time
```

docs/adr/0024-dynamic-plugin-system.md
stijndehaes: Maybe the impact isn't that big? But I would like to see some examples, and the question for me remains what do we recommend people to do. Giving options is all good and well but we should have an opinion on what is our first choice

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+* Simplicity beats defensiveness: given every customer runs their own separate, self-hosted copy of the portal, prefer the simplest option that works over one that defends against risks that don't apply to this deployment model.
+* A plugin should be usable as soon as its package is installed - the portal discovers it automatically, without a separate step to name it.
+* For a first version, this should be built on top of the platform, platform services, and environments data model as it exists today - or simply not use it at all. Redefining that model completely is out of scope here, but this decision shouldn't make it harder to change later.
+
+## Out of scope
+* Revising the core data model, including the per-environment infrastructure tables (`env_platform_configs`, `env_platform_service_configs`, `platform_service_configs`) - a separate ADR. These keep existing as they are; a plugin built under this decision can read them if it needs to. This also include the defenition of the relation shp with techincal asset and techincal asset type, is a plugin exactly that or not.
```
stijndehaes:
```suggestion
* Revising the core data model, including the per-environment infrastructure tables (`env_platform_configs`, `env_platform_service_configs`, `platform_service_configs`) - a separate ADR. These keep existing as they are; a plugin built under this decision can read them if it needs to. This also include the definition of the relationship with technical asset and technical asset type
```

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+As this needs some work and is currently not exactly clear what the data model should be and is not.
```
stijndehaes:
```suggestion
This will be a next ADR
```

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+* An admin interface for managing integrations. Integrations (including the ones with no code) remain reachable only through the underlying data/API for now, matching how they're exposed to the rest of the interface today.
+* Attaching integrations beyond data products. Integrations continue to attach to a data product's technical assets only.
+* Logging, tracing, and observability conventions for plugin authors, beyond the portal catching exceptions at the plugin call boundary. Structured logging format, distributed tracing, and metrics for plugin calls are not addressed here.
+* Concurrency aspects on replicas when running on a cluster setup.
+* Tooling to test out alembic or database migrations of the plugin's own tables.
```
stijndehaes: I do think this should be in scope, people should somehow be able to test out their own plugin before installing it no?

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+* An admin interface for managing integrations. Integrations (including the ones with no code) remain reachable only through the underlying data/API for now, matching how they're exposed to the rest of the interface today.
+* Attaching integrations beyond data products. Integrations continue to attach to a data product's technical assets only.
+* Logging, tracing, and observability conventions for plugin authors, beyond the portal catching exceptions at the plugin call boundary. Structured logging format, distributed tracing, and metrics for plugin calls are not addressed here.
+* Concurrency aspects on replicas when running on a cluster setup.
```
stijndehaes: Unsure why this is here? This btw should be in scope, but you can also just remove it

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+* Tooling to test out alembic or database migrations of the plugin's own tables.
+* Portal Core has no responsibility of keeping and storing permanent credentials for a plugin's own API calls. If a plugin needs to make authenticated calls, it can implement calls based on the user authenticated and making the plugin call.
+* Setting up the proper `UIElementMetadata` for the plugin's own fields, beyond the minimal adapter that translates a plugin's declared list of string-typed fields into custom form fields.
```
stijndehaes: I don't understand this, can you try in other words :)

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+The alternatives below are recorded for completeness
+
+1. **Build it into the core project (today's situation)** - rejected, this is the exact problem being solved.
+2. **Ship it as a separately installable package** - built and versioned like any other Python package, installed into the customer's own image.
+   * **2a - discovered automatically**, via Python's standard package-registration mechanism.
+   * **2b - discovered by explicit name** - the image author names exactly which modules to load, in one setting.
+3. **Drop it into a scanned folder, loaded in-process** - the portal imports a plugin file directly at startup, no packaging needed.
+4. **Drop it into a scanned folder, run as a separate program per call** - any language, isolates a broken plugin, but slower per call and more work to build.
+5. **Run it inside the portal, in a locked-down sandbox** - safe by construction, but needs a new sandboxable-format runtime.
+6. **Run it entirely on a separate server the plugin author operates** - 6a renders its own screen, 6b just answers questions for the portal's generic renderer.
```
stijndehaes:
```suggestion
6. **Run it entirely on a separate server the plugin author operates**:
  * **6a** renders its own screen
  - **6b** just answers questions for the portal's generic renderer.
```

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+7. **Build a public marketplace** - a way to find plugin code; could sit on top of any option above.
+
+### Loading the python packages inside portal core
```
stijndehaes, line 46: Is there a relation with `### Where the plugin's code runs`?

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Option 1: Discover plugins automatically via Python entry points
+- Python's standard entry-points mechanism - the same one Airflow's own classic plugin registration uses for operators, hooks, executors, and UI views. (Airflow separately uses `pluggy`'s hookspec/hookimpl system for its listener API, where many plugins can each optionally implement several lifecycle hooks - a different shape of problem than one class per plugin, so not a reason to pull `pluggy` in here.)
+- Installing the package is enough - nothing extra to configure.
+- Robust and well-established: this is what "advertise yourself as a plugin" means in the Python packaging ecosystem.
+- Harder to see at a glance what's loaded, since nothing in the Dockerfile names it - it's implicit in whatever got installed.
```
stijndehaes: When we run portal we can easily show what is loaded, I don't think that this is an issue

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+- More manual bookkeeping: installing a plugin isn't enough by itself, it also has to be named.
+- Doesn't fit "just install it and it works."
+
+### How the portal recognizes a plugin's shape
+
+Once a plugin class is loaded (see "Loading the python packages inside portal core" above), the portal needs to know it actually has the right shape - the right class attributes and methods - before treating it as a real plugin.
+
+#### Option 1: `typing.Protocol`, checked structurally
+- The portal declares the shape as a `Protocol` (`@runtime_checkable`); a plugin class only needs to match it structurally - the same attribute and method names - not inherit from anything the portal ships.
+- A plugin package never needs the portal, or even its SDK, as a dependency to be recognized.
```
stijndehaes: I don't think this will ever be the case no? Don't you need the platforms etc? or at least environments

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+- Installing the package is enough - nothing extra to configure.
+- Robust and well-established: this is what "advertise yourself as a plugin" means in the Python packaging ecosystem.
+- Harder to see at a glance what's loaded, since nothing in the Dockerfile names it - it's implicit in whatever got installed.
+
+#### Option 2: No automatic discovery, hardcode it through settings
+- Easy to understand - one setting lists exactly what's loaded.
+- More manual bookkeeping: installing a plugin isn't enough by itself, it also has to be named.
+- Doesn't fit "just install it and it works."
+
+### How the portal recognizes a plugin's shape
```
stijndehaes: Maybe just remove it, since ABC is pretty obvious?

docs/adr/0024-dynamic-plugin-system.md
stijndehaes: Maybe add to appendix when it's not important :) Appendix tells people only read this when interested but not required reading

docs/adr/0024-dynamic-plugin-system.md
stijndehaes: As discussed with @HannesDS we will do this in a follow up ADR

docs/adr/0024-dynamic-plugin-system.md
stijndehaes: As discussed with @HannesDS can be removed, since alembic migrations have no issues running concurrently, they take a lock to ensure no issues can happen

docs/adr/0024-dynamic-plugin-system.md
stijndehaes: @HannesDS discussed, we make this in scope, this shouldn't be complex, just some docs on how to run the alembic migrations. Plugin authors should be a able  to test out their migrations

docs/adr/0024-dynamic-plugin-system.md
stijndehaes: `def get_ui_metadata(cls, db: Session) -> list[UIElementMetadata]:`

Is a method in the backend that returns the things to be rendered in the front end, so this is untrue

docs/adr/0024-dynamic-plugin-system.md
stijndehaes, line 1: What is missing:


Plugin developers need portal as a python package to be able to:
  - Use environments, platforms, ... in general to interact with Portal
  - Inherit the abstract plugin class

In general they only need to a subset or portal, but probably describing that boundary is too difficult for now, OK to check this in 3 months after we migrated our plugins

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+
+Alongside that, one shared table tracks each plugin's table, its version, and its schema. Only the latest version needs to be kept there - older versions live in the plugin's own migration history, not in this tracking table.
+
+- Comes with a built-in versioning system via Alembic - and this isn't custom machinery: Alembic already supports running multiple, independent migration histories side by side ("Run Multiple Alembic Environments" in its own docs), each with its own version-tracking table.
+- Easier to put indexes and joins in place than against a JSON column.
+- Lots of new tables, growing with however many plugins are installed.
+
+ The core project's own migrations already run as a deploy-time step, before the app starts serving traffic . That same step extends to also walk every named plugin: each plugin's own class hardcodes the revision it wants (set by the plugin author, versioned right alongside its code, discovered the same way its identity is  the portal core compares that to the plugin's tracked current revision and reconciles the table to match, upgrading or downgrading as needed.
+
+### How the API describes a plugin's configuration
```
stijndehaes: Current API endpoints are already generic, we can reuse that. You can remove the section below

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### Option 2: Dedicated endpoints per plugin
+- Would mean a new endpoint per plugin, built-in or custom - but a build-time-generated endpoint can't exist yet for a plugin that isn't installed at build time, so this contradicts the "generic value for every type" answer above. Not pursued further.
+
+### The plugin's field vocabulary, and who owns turning it into a form
```
stijndehaes: As discussed already in place currently, you can remove the option and refer to current impl

stijndehaes (changes requested):

docs/adr/0024-dynamic-plugin-system.md
HannesDS: Well maybe that is obvious but I don't want to include anything related to UI screesn for Admins where they can a "manage" the plugins, to disable, enable, make it default, etc.

docs/adr/0024-dynamic-plugin-system.md
HannesDS: This is indeed not very clear. The idea is indeed to also move them out of tree, and make sure that we eat our own dog food. This will prob. move in seperate PR's and not a bing bang.

docs/adr/0024-dynamic-plugin-system.md
HannesDS: Cleared out, agree that it shoul be 3rd person.

docs/adr/0024-dynamic-plugin-system.md
HannesDS: Added a section that describes this, fully agree.

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,159 @@
...
+- Robust and well-established: this is what "advertise yourself as a plugin" means in the Python packaging ecosystem.
+
+#### Option 2: No automatic discovery, hardcode it through settings
+- Easy to understand - one setting lists exactly what's loaded.
+- More manual bookkeeping: installing a plugin isn't enough by itself, it also has to be named.
+- Doesn't fit "just install it and it works."
+
+### Where and how is the configuration of the plugin stored
+
+A plugin's identity - its name, icon, and fields - lives directly on its own Python class, inherited from the portal's `TechnicalAssetPlugin` ABC (see Decision Outcome), alongside the configuration behavior described below; it isn't a separate thing to store. The icon is referenced as a bundled file path next to the plugin's own code, not inlined as a raw string.
```
stijndehaes:
```suggestion
A plugin's identity - its name, icon, and fields - lives directly on its own Python class, inherited from the portal's `TechnicalAssetPlugin` ABC (see Decision Outcome), alongside the configuration behavior described below; it isn't a separate thing to store.
```

I would remove the icon reference, an ADR shouldn't concern such details :)

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,159 @@
...
+
+### The plugin's field vocabulary, and who owns turning it into a form
+
+A plugin declares its fields by returning `UIElementMetadata` instances directly - the same class already used to render every technical asset type's form, imported from the new package (see "Distributing portal core as an installable Python package" below) rather than reimplemented or expressed as JSON. There is no separate, plugin-only vocabulary and no adapter step translating one shape into another: a plugin author gets exactly the same building block core types use, which is far less work than hand-writing a JSON schema. The two are meant to fully converge - core's own technical asset types move onto this same interface too (see Decision Outcome), so there ends up being one single, working way to declare a technical asset's form, not two.
+
+#### Option 1: Single table for all plugins with a JSON column
```
stijndehaes, line 66: We mis the title on this option I believe? Or is it supposed to be under `The plugin's field vocabulary, and who owns turning it into a form` I am a bit confused :)

docs/adr/0024-dynamic-plugin-system.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Option 2: Separate table for each plugin, plus one table tracking versions
+Creating a table for each plugin - the schema can be partially defined by portal core (an id, a name, a version, ...) and partially self-defined by the plugin. This stays a single table per plugin, not a denormalized model.
+Included is also some basic tooling to test out Alembic database migrations of the plugin's own tables, since plugin authors should be able to try out their own migrations before shipping them, the same as they would for portal core's own.
+Alongside that, one shared table tracks each plugin's table, its version, and its schema. Only the latest version needs to be kept there - older versions live in the plugin's own migration history, not in this tracking table.
+
+- Comes with a built-in versioning system via Alembic - and this isn't custom machinery: Alembic already supports running multiple, independent migration histories side by side ("Run Multiple Alembic Environments" in its own docs), each with its own version-tracking table.
+- Easier to put indexes and joins in place than against a JSON column.
```
stijndehaes:
```suggestion
- Easier to put indexes, joins and constraints in place than against a JSON column.
```

stijndehaes (approved):

docs/adr/0024-dynamic-plugin-system.md
HannesDS: Was a bit too hasty with removing stuff, always fun. Added the title back.

stijn-janssens (approved):
I like the idea and the ADR, however I am not sure we can disregard looking at the platforms, platforms services and environments completely when trying to migrate existing plugins to this system.
I believe we would need to tackle that topic first before we can do the migration. New plugins could ofcourse be created in this way without refering to the platforms or platform services

PR #4293, opened 2026-09-11
Fix issues with commit not happening

stijn-janssens:
Commits were removed from the add output link call.
This results in a 200 being send, immediate approve being called but because commit has not executed before 200 being send the approve got a 404. After a refresh you could approve.
So commit is necessary here.
Second bug: is admin checks against utc now and can malfunction on different timezones.


backend/app/core/authz/db_utils.py
```diff
@@ -12,7 +13,7 @@ def is_user_admin(user_id: uuid.UUID):
     return (
         select(User.id)
         .where(User.id == user_id)
-        .where(User.admin_expiry > func.now())
+        .where(User.admin_expiry > utcnow())
```
HannesDS, line 16: Well this is what it was before no? The problem with UTS is that if the machine running it is not running in UTC your admin credneitals expire directly since they were in the past, how did you encounter this issue?

HannesDS (changes requested):

stijndehaes (changes requested):
It should actually not be necessary to introduce a commit here, this means our setup is problematic no?

Because the commit should happen before we send out a 200, the commit might still fail and result in a an error we have to present the user. This might also break a lot of other calls in Portal, shouldn't we look into the root cause?

HannesDS:
> It should actually not be necessary to introduce a commit here, this means our setup is problematic no?
> 
> Because the commit should happen before we send out a 200, the commit might still fail and result in a an error we have to present the user. This might also break a lot of other calls in Portal, shouldn't we look into the root cause?

It tuns out, on FastAPI 0.141.1 the commit in a yield dependency runs after the response bytes are on the wire.

stijndehaes:
> > It should actually not be necessary to introduce a commit here, this means our setup is problematic no?
> > Because the commit should happen before we send out a 200, the commit might still fail and result in a an error we have to present the user. This might also break a lot of other calls in Portal, shouldn't we look into the root cause?
> 
> It tuns out, on FastAPI 0.141.1 the commit in a yield dependency runs after the response bytes are on the wire.

After look at it, I don't think so the only commit in that version that changes router is: https://github.com/fastapi/fastapi/commit/d62354434b2e508fe89024213b220ca8e67dea5e I don't think this breaks that

stijn-janssens:
closing this pr, admin fix to be seen if needed still

PR #4295, opened 2026-09-11
If a node is removed because it is hidden, make sure the connections are also removed

### LINKED ISSUE #4294: Explorer graph breaks with hidden data products
## Describe the bug
Hidden data products are filtered out from returned responses, but are somewhere referenced in expected explorer renderings. This results in some sort of not found error in the console and an empty explorer.

### Steps to reproduce
Clear, sample database -> Go to explorer
Explorer is empty

### Expected behavior
Explorer not empty

stijn-janssens:
Closes #4294 

### DIFF
--- a/backend/app/graph/service.py
+++ b/backend/app/graph/service.py
@@ -233,4 +233,11 @@ class GraphService:
                 ]
             )
 
+        node_ids = {node.id for node in nodes}
+        edges = [
+            edge
+            for edge in edges
+            if edge.source in node_ids and edge.target in node_ids
+        ]
+
         return Graph(nodes=set(nodes), edges=set(edges))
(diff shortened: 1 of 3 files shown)


backend/app/graph/service.py
```diff
@@ -233,4 +233,11 @@ def get_graph_data(
                 ]
             )
 
+        node_ids = {node.id for node in nodes}
+        edges = [
+            edge
+            for edge in edges
+            if edge.source in node_ids and edge.target in node_ids
+        ]
```
HannesDS, line 241: Looks good to me, only wondering if this shouldn't be a seperate method, since it is already quite a method.

HannesDS (approved):

backend/app/graph/service.py
stijn-janssens: Fair remark, but I think for now not super needed to fix?

PR #4301, opened 2026-09-14
Fix: DB commit happened after response was sent

stijndehaes:
Changing the db dependency to scope=function
ensures that the commit happens before
the response is being send out.

Replaces: #4293


vlieven:
You can define an annotation like so, to avoid having to define the scope everywhere and probably forgetting it in some spots.

```
DbSession = Annotated[Session, Depends(get_db_session, scope="function")]
```

Although DbSession is a stupid name

stijndehaes:
> You can define an annotation like so, to avoid having to define the scope everywhere and probably forgetting it in some spots.
> 
> ```
> DbSession = Annotated[Session, Depends(get_db_session, scope="function")]
> ```
> 
> Although DbSession is a stupid name

I just added a test so we can't forget :) Unsure if the annotation you propose is better, they could work together I think. What do you think?

stijn-janssens (approved):

PR #4324, opened 2026-09-16
Stage generated SDK files only in CI

HannesDS:
A pre-commit hook must not run git add. During a commit, pre-commit saves unstaged changes to a patch, runs the hooks, then restores that patch. A hook that stages files behind its back leaves the index in a state the restore cannot reconcile, which showed up as sdk/sdk/* appearing staged at sdk/* and a commit that could never succeed.

The git add only ever served the CI branch below it, where it makes newly generated files show up in the diff that explains the failure. Move it there.


stijn-janssens (approved):
I assume you had issues with this?

HannesDS:
Yes indeed! Had it on a worktree

Issue #4325, opened 2026-09-16
Dynamically Loadable plugins


HannesDS:
## Summary
https://app.notion.com/p/datamindedbe/Dynamically-loadable-asset-types-prep-work-handover-to-imec-39193b0d1c71804da986edf572000c17?v=2be93b0d1c718053913b000c0e7a6f32&source=copy_link

### Open Points

- [ ] Remove the hardcoded list ENABLED_PLUGINS, this is a bit confusing and you don't want end-users to have to enable there own new plugins. A future point in the far future is to make it onboarding for users and let them select those. In addtion for new ones added by engineers themsevles it should be auto added, only for the ones we provide you should hardcode it. 
- [ ] TechnicalAssetPlugin has a configuration type string, but get_confituraiton retuerns a generaic not saying much 
```configuration_type: str — which plugin this technical asset is. The value is "S3TechnicalAssetConfiguration". It's the discriminator stored in
  data_output_configurations.configuration_type and the key the registry looks up.

  ConfigType — a union of per-environment platform infrastructure schemas:

  ConfigType = AWSS3Config | AWSGlueConfig | DatabricksConfig | SnowflakeConfig | ...

  Those describe what a platform service offers in an environment — which buckets exist in AWS in production, which Databricks workspaces, and so on. They come from
  EnvironmentConfigsGetItem.config, nothing to do with the plugin's own identity.
```

- [ ] Clean up UIElementMetadataResponse, there is platform, parent_platform. All of this is very confusing. 
- [ ] INvestigate the python installation in the dockerfile at the root to make sure we don't do unneccary things
- [ ] Currently there is a nasty script that copies the schema's for the plugins of portal to the sdk so the current aspects are typed. 
however this is quite nasty and we should think is there no easier way of resolving that. Or at least more clenaly, copy everything or find another way to expose those classes. 
- [ ] write proper documentation
- [ ] Make sure that when we copy the asset configurations to the sdk it is less complex. It is not really copy but file construction and rather complex. Investigate other patterns to make it easier
- [ ] add tests for:  [ackend/app/shared/schema.py](https://github.com/conveyordata/data-product-portal/pull/4327/files/cd4fc2681578ee3f800b2534bcee206e2b1f9a6b#diff-5e270a01b4041b87c9ce7203e983b9df42c77eff6dc31456fc08afd4cce82ecc)
```@staticmethod
    def _orm_kwargs(schema: "ORMModel") -> dict:
```

PR #4326, opened 2026-09-16
Dynamically Loadable Plugins - plugin registry

HannesDS:
## Summary 

Flow of the plugin discovery and enablement.

This is not yet used and is scaffolding for the future PR's and dynamicaally loadable plugins. Part of feature #4325 

```mermaid
sequenceDiagram
    participant Image as Docker image build
    participant Registry as PluginRegistry
    participant EP as importlib.metadata
    participant Pkg as portal_plugins
    participant Builtin as schema_union

    Image->>Image: pip install data-product-portal-plugins
    Note over Image: the wheel declares<br/>[project.entry-points."data_product_portal.plugins"]

    Registry->>Builtin: import (registers in-tree plugin classes)
    Registry->>EP: entry_points(group="data_product_portal.plugins")
    EP-->>Registry: github, s3

    loop each entry point
        Registry->>Pkg: entry_point.load()
        alt import fails or not a TechnicalAssetPlugin
            Registry->>Registry: log and skip
            Note over Registry: one broken plugin never<br/>stops the others
        else ok
            Pkg-->>Registry: plugin class
        end
    end

    Registry->>Registry: log every plugin discovered
```

Discovery and exposure are separate questions. `enabled()` filters what
`discovered()` found through `ENABLED_PLUGINS`, because our own plugins ship in
one package installed by default, so installed and should-be-visible are not the
same thing.

```mermaid
flowchart LR
    A[installed packages] -->|entry points| B[discovered]
    C[in-tree classes] --> B
    B -->|ENABLED_PLUGINS| D[enabled]
    D --> E[forms, tiles, create, read]
    B --> F[migrations, access-tile URLs]
```

Migrations and URL lookups use `discovered()`, not `enabled()`: a table still
needs migrating, and an existing asset still needs its URL, even while its plugin
is switched off in the UI.


.github/workflows/release-workflow.yml
```diff
@@ -209,6 +209,36 @@ jobs:
...
+          cache: 'poetry'
+          cache-dependency-path: backend/poetry.lock
+      - name: Build portal core
+        working-directory: backend
+        run: |
+          poetry build
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
+        with:
+          packages-dir: backend/dist/
```
stijndehaes, line 285: Have we set up pypi?

backend/app/plugins/registry.py
```diff
@@ -0,0 +1,73 @@
...
+                    f"Entry point '{entry_point.name}' does not subclass "
+                    "TechnicalAssetPlugin, skipping"
+                )
+                continue
+            plugins[plugin.name] = plugin
+
+        for plugin in TechnicalAssetPlugin.__subclasses__():
+            name = getattr(plugin, "name", None)
+            if name:
+                plugins.setdefault(name, plugin)
```
stijndehaes: I think we can drop this if we register our own plugins through entry points no, then this part is not required?

backend/app/plugins/registry.py
```diff
@@ -0,0 +1,73 @@
...
+        return list(self._plugins.values())
+
+    def enabled(
+        self,
+    ) -> list[type[TechnicalAssetPlugin]]:
+        return [
+            plugin
+            for plugin in self.discovered()
+            if plugin.name in settings.ENABLED_PLUGINS
+        ]
```
stijndehaes, line 45: I dislike that `settings.ENABLED_PLUGINS` means that you need to install your plugin and then you need to enable is. Can we brainstorm on a solution?

backend/app/plugins/runtime.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any, Callable
+
+from fastapi import HTTPException, status
+
+from app.core.logging import logger
+
+
+def call_plugin(
```
stijndehaes: I don't think this method is needed? All plugins implement the same abstract class, this looks like a smell, since all methods should be defined by that abstract class no?

backend/app/technical_asset_configuration/__init__.py
```diff
@@ -1,29 +1,4 @@
...
-    "RedshiftTechnicalAssetConfiguration",
-    "RustFSTechnicalAssetConfiguration",
-    "SnowflakeTechnicalAssetConfiguration",
-    "S3TechnicalAssetConfiguration",
-    "AzureBlobTechnicalAssetConfiguration",
-]
+# Plugin classes are registered by importing them, which
+# app/technical_asset_configuration/schema_union.py does. Keeping this module
+# empty means importing base_schema (which every plugin does) stays cheap and
+# free of import cycles.
```
stijndehaes:
```suggestion
```

backend/app/technical_asset_configuration/schema_union.py
```diff
@@ -2,13 +2,21 @@
 
 from pydantic import Field
 
+from app.technical_asset_configuration.agno.schema import AgnoPlugin  # noqa: F401
```
stijndehaes: Why are these needed if they are not used? I assume because you removed everything from `backend/app/technical_asset_configuration/__init__.py` and all of a sudden the sublcasses are not detected right? I believe registering all of them through  entry points fixes this

stijndehaes (changes requested):

backend/tests/app/plugins/test_registry.py
```diff
@@ -0,0 +1,99 @@
...
+                "glue",
+                "app.technical_asset_configuration.glue.schema:"
+                "GlueTechnicalAssetConfiguration",
+            )
+        ],
+    )
+
+    names = {plugin.name for plugin in registry.discovered()}
+
+    assert "GlueTechnicalAssetConfiguration" in names
```
stijndehaes: Isn't this test a bit stupid? Because the `GlueTechnicalAssetConfiguration` will be detected through subclasses anyway? TO make this work we need to ensure that those subclasses do not return `GlueTechnicalAssetConfiguration`

.github/workflows/release-workflow.yml
HannesDS: Not yet :)

stijn-janssens:
have not yet looked at the code but I don't agree that access tiles (and URLs) should always use discovered and not enabled. It is perfectly possible that we don't want to enable certain access tiles that we ship (such as Conveyor), who are shown on the product page.
Could be that we have to change this logic in the future, but to stay backwards compatible for now, I think the access tiles also need to look at enabled setting as well, at least for in tree plugins

backend/app/mcp/loader.py
```diff
@@ -1,26 +1,25 @@
...
-from app.technical_asset_configuration.schema_union import DataOutputs
+from app.plugins.registry import plugin_registry
 
 
 def load_plugins(mcp: FastMCP) -> None:
-    """Register MCP tools from all AssetProviderPlugin subclasses."""
-    for cls in get_args(DataOutputs):
-        cls.register_mcp_tools(mcp)
+    """Register MCP tools from every enabled plugin."""
+    for plugin in plugin_registry.enabled():
```
stijn-janssens, line 16: This also contradicts the discovery vs enabled discussion. If you have assets of disabled plugins you would want to be able to use them in the MCP right?

backend/app/plugins/registry.py
HannesDS: Yes indeed, I was thinking to later move them all to the new structure, but there is no point in also not doing it same step by step approach.

backend/app/technical_asset_configuration/schema_union.py
HannesDS: Yes indeed, again, agreed.

backend/app/plugins/runtime.py
HannesDS: Well not really, it is called at 3 different points in the code, by having this, I only need to catch it once, and make sure that the error in the Plugin is properly caught. If you disagree happy to make it happen 3 times inline.

backend/tests/app/plugins/test_registry.py
HannesDS: Yes it is, overlooked it, thanks!

backend/app/plugins/registry.py
HannesDS: Discussed IRL. Put this on the TODO list. An idea for future refernces is to introduce an onboarding with steps, where at startup you would confirm which plugins you want/need.

HannesDS:
> have not yet looked at the code but I don't agree that access tiles (and URLs) should always use discovered and not enabled. It is perfectly possible that we don't want to enable certain access tiles that we ship (such as Conveyor), who are shown on the product page. Could be that we have to change this logic in the future, but to stay backwards compatible for now, I think the access tiles also need to look at enabled setting as well, at least for in tree plugins

Tiles already respect enabled. get_platform_tiles → get_all_technical_assets_ui_metadata → plugin_registry.enabled(), so if Conveyor isn't in ENABLED_PLUGINS its tile never renders on the product page. So that case you're describing 
laready works.
 On get_url specifically, in main scans AssetProviderPlugin.__subclasses__() with no enablement check, so it already ignores ENABLED_PLUGINS as is. Making it respect enabled would be a behaviour change, not backwards compatibility, I kept it the same for now but I can change it if you want.

backend/app/mcp/loader.py
HannesDS: Well would you? not sure? I think for now I would keep the logic as it is on main, and maybe we can revisit this. I already discussed with Stijn to maybe remove enabled pluings in the future and add an onboarding track that adds asks the user on onboarding which plugins they want/need.

.github/workflows/release-workflow.yml
HannesDS: Done, or requested at least 
image

stijn-janssens:
perfect, I think indeed keeping functionality as is on main makes the most sense for now. Might have just been confused with the diagram then as access tiles is worded under discovery there :)

backend/app/mcp/loader.py
stijndehaes: And there if a user disables a plugin, we can just remove all technical assets and warn them about this

backend/app/plugins/registry.py
```diff
@@ -0,0 +1,65 @@
...
+        plugins: dict[str, type[TechnicalAssetPlugin]] = {}
+
+        for entry_point in entry_points(group=ENTRY_POINT_GROUP):
+            try:
+                plugin = entry_point.load()
+            except Exception:
+                logger.exception(
+                    f"Failed to load plugin entry point '{entry_point.name}', skipping"
+                )
+                continue
```
stijndehaes:
```suggestion
            plugin = entry_point.load()
```

I would just let it crash and this serves as a warning to the platform dev or portal dev that something is broken

backend/app/plugins/registry.py
```diff
@@ -0,0 +1,65 @@
...
+            if not (
+                isinstance(plugin, type) and issubclass(plugin, TechnicalAssetPlugin)
+            ):
+                logger.error(
+                    f"Entry point '{entry_point.name}' does not subclass "
+                    "TechnicalAssetPlugin, skipping"
+                )
+                continue
```
stijndehaes:
```suggestion
                raise Exception("The registered....
```

We want to warn the dev this is not working correctly no?

backend/app/technical_asset_configuration/service.py
```diff
@@ -94,6 +88,11 @@ def _build_metadata_response(
                 detailed_name=platform_meta.detailed_name,
                 has_environments=platform_meta.has_environments,
             )
+        except Exception:
+            logger.exception(
+                f"Plugin '{plugin_class.name}' failed to describe its form, skipping"
+            )
+            return None
```
stijndehaes: Why don't we just let this fail like before?

backend/tests/app/technical_asset_configuration/test_service.py
```diff
@@ -0,0 +1,60 @@
...
+        platform_key="raises",
+        detailed_name="Raises",
+    )
+
+    @classmethod
+    def get_ui_metadata(cls, db: Session):
+        raise RuntimeError("this plugin is broken")
+
+
+def test_get_all_technical_assets_ui_metadata__drops_only_the_plugin_that_raises(
```
stijndehaes: SO I don't think we want this behavior

Dockerfile
stijndehaes, line 75: I think we can simplify this, because Portal was installed and copied over from python-build, perhaps we don't need to copy it again

stijndehaes (changes requested):

backend/app/technical_asset_configuration/service.py
HannesDS: Agreed, removed this.

backend/app/plugins/registry.py
HannesDS: Agreed.

Dockerfile
HannesDS: We have it duplicated indeed, so I will simplify this and also install it from a wheel. Since this is also what an end-user might due from Pypi.

backend/app/technical_asset_configuration/base_schema.py
```diff
@@ -108,7 +108,7 @@ class PlatformMetadata(ORMModel):
     show_in_form: bool = True
 
 
-class AssetProviderPlugin(ORMModel, ABC):
+class TechnicalAssetPlugin(ORMModel, ABC):
     """Base class for all data output provider plugins"""
```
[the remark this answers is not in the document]
HannesDS: Will be fixed later on.

backend/app/plugins/registry.py
```diff
@@ -0,0 +1,59 @@
...
+        return list(self._plugins.values())
+
+    def enabled(
+        self,
+    ) -> list[type[TechnicalAssetPlugin]]:
+        return [
+            plugin
+            for plugin in self.discovered()
+            if plugin.name in settings.ENABLED_PLUGINS
+        ]
```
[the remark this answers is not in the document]
HannesDS: This will be fixed later.

Dockerfile
stijndehaes: Checked with @HannesDS, is added as todo: https://github.com/conveyordata/data-product-portal/issues/4340

stijndehaes (approved):

PR #4327, opened 2026-09-16
Dynamically Loadable Plugins: Describe every plugin's configuration the same way

HannesDS:
## Describe every plugin's configuration the same way- #4327

Second part of the Dynamically loadable plugin which prepared the configuration of the assets in such a way that it can be discovered at runtime instead of at importtime. Note that both are still possible since every plugin we currently supports still sits in tree. 

Describe every plugin's configuration the same way

A technical asset's configuration was a Pydantic discriminated union over
a hardcoded list of types, so the published OpenAPI schema named exactly
the plugins compiled in. A plugin installed from a package could never
appear there, and a deployment with different plugins generated a
different SDK, CLI and frontend client from the same release.

The union becomes an annotated alias. It resolves configuration_type
through the registry, validates a dict or an ORM row into that plugin,
and serialises it back to a plain object. The published schema is a
free-form object for every plugin, identical in every deployment. What
goes over the wire does not change: configuration_type stays a key inside
the configuration object, so the frontend still reads it as before. The
spec loses 519 lines of per-plugin schemas, and test_open_api_export
asserts no plugin name can leak back into it.

A plugin can also ship its own icon. The portal reads it from the
plugin's package and puts it on the existing /v2/plugins response as a
data URI, rendered through an img tag so a third-party SVG cannot run
script. In-tree types leave it unset and keep resolving icon_name against
the frontend's own assets.

## Creating a technical asset

```mermaid
sequenceDiagram
    participant Client
    participant Schema as CreateTechnicalAssetRequest
    participant Alias as DataOutputConfiguration
    participant Registry as PluginRegistry
    participant Plugin as S3TechnicalAssetConfiguration
    participant DB as Postgres

    Client->>Schema: {configuration: {configuration_type: "S3...", bucket, path}}
    Schema->>Alias: PlainValidator(configuration)
    Alias->>Registry: get("S3TechnicalAssetConfiguration")
    alt unknown
        Registry-->>Client: 400
    else installed
        Registry-->>Alias: plugin class
        Alias->>Plugin: model_validate(dict)
    end
    Plugin->>DB: insert, class table inheritance unchanged
```

The wire format does not change: `configuration_type` stays a key inside the
`configuration` object, so the frontend still reads it exactly as before. What
changes is the published OpenAPI type, so the SDK, CLI and frontend client are
regenerated. The spec loses 519 lines of per-plugin schemas.

### Context

its base is the PR below it.

| # | PR | |
|---|----|-|
| 0 | `fix/precommit-sdk-hook-staging` | #4324 |
| 1 | `feat/4325-plugin-registry` | #4326 |
| 2 | `feat/4325-generic-configuration` | |
| 3 | `feat/4325-plugin-migrations` | |
| 4 | `feat/4325-optional-platform-service` | |
| 5 | `feat/4325-plugins-out-of-tree` | |

Each is functional on its own: none leaves the portal broken if the ones above
it never merge.


.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,48 @@
+---
+name: code-review
+description: 'Reviews a pull request diff for the problems reviewers in this repository keep finding by hand; use when reviewing a PR, before approving or requesting changes.'
+---
```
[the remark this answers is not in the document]
HannesDS: INdeed removed

backend/app/technical_asset_configuration/azure_blob/schema.py
```diff
@@ -22,22 +22,21 @@
...
-    name: ClassVar[str] = NAME
+    name: ClassVar[str] = CONFIGURATION_TYPE
     version: ClassVar[str] = "1.0"
 
     domain: str = ""
     path: str = ""
     container_name: str
 
-    configuration_type: Literal[DataOutputTypes.AzureBlobTechnicalAssetConfiguration]
+    configuration_type: str = CONFIGURATION_TYPE
```
stijndehaes: I think we can remove this since it's the same as name?

backend/app/technical_asset_configuration/base_schema.py
```diff
@@ -177,6 +176,26 @@ def get_platform_metadata(cls) -> PlatformMetadata:
...
+                resources.files(platform_meta.icon_package)
+                .joinpath(platform_meta.icon_name)
+                .read_bytes()
+            )
+        except Exception:
+            logger.exception(
+                f"Plugin '{cls.name}' has no readable icon at "
+                f"{platform_meta.icon_package}/{platform_meta.icon_name}"
+            )
+            return None
```
stijndehaes, line 200: I assume our front end has a fallback icon?

backend/app/technical_asset_configuration/schema_response.py
```diff
@@ -23,12 +24,13 @@ class UIElementMetadataResponse(ORMModel):
...
     result_tooltip: str = "The path you can access through this technical asset"
-    platform: str  # e.g., "s3", "redshift", "snowflake"
-    display_name: str  # Display name for the platform tile
-    icon_name: str  # Icon filename (e.g., "s3-logo.svg")
-    parent_platform: Optional[str] = None  # e.g., "aws" for s3, redshift, glue
-    platform_tile: Optional[PlatformTile] = None  # Complete tile structure
-    show_in_form: bool = True  # Whether to show this platform in the configuration form, can be set to False for platforms that are only meant to be shown as tiles without detailed configuration options
+    platform: str
+    display_name: str
+    icon_name: str
```
stijndehaes, line 29: Is incon name still needed since we now have `icon_data_uri`? Or do our current plugins still use this?
Maybe in a follow up PR we refactor them to supply their icon backend?

backend/app/technical_asset_configuration/schema_response.py
```diff
@@ -23,12 +24,13 @@ class UIElementMetadataResponse(ORMModel):
...
+    icon_data_uri: Optional[str] = None
+    parent_platform: Optional[str] = None
+    platform_tile: Optional[PlatformTile] = None
+    show_in_form: bool = True
```
stijndehaes, line 35: The comment was usefull here no?

backend/app/technical_asset_configuration/schema_union.py
```diff
@@ -1,59 +1,49 @@
...
+    if isinstance(value, TechnicalAssetPlugin):
+        return value
+
+    configuration_type = (
+        value.get("configuration_type")
+        if isinstance(value, dict)
+        else getattr(value, "configuration_type", None)
+    )
+    if not configuration_type:
+        raise ValueError("configuration_type is required")
```
stijndehaes: This should be based on name I suppose

backend/app/technical_asset_configuration/schema_union.py
```diff
@@ -1,59 +1,49 @@
...
+from fastapi import HTTPException, status
+from pydantic import PlainSerializer, PlainValidator, WithJsonSchema
+
+from app.plugins.registry import plugin_registry
+from app.technical_asset_configuration.base_schema import TechnicalAssetPlugin
+
+plugin_registry.discovered()
+
+
+def _resolve_configuration(value: Any) -> TechnicalAssetPlugin:
```
stijndehaes: Maybe rename to `technical_asset_validator`?

stijndehaes (changes requested):
I think this might break provisioner, suddenly we don't have types for technical assets. 
Perhaps check with @stijn-janssens if this would break things when trying to privision technical assets.

We also need a solution ofcourse, since out of tree plugins can't change our generated SDK

backend/app/technical_asset_configuration/schema_union.py
HannesDS: Agreed, done

backend/app/technical_asset_configuration/schema_response.py
HannesDS: I added them again, could be useful I guess, I rather have proper documentation on it than this.

backend/app/technical_asset_configuration/base_schema.py
HannesDS: It has now :)

backend/app/technical_asset_configuration/azure_blob/schema.py
HannesDS: I made this a computed field, since it is still needed to not break the schema but we keep the field only once.

backend/app/technical_asset_configuration/schema_response.py
HannesDS: Yes they still do, I was planning in a future PR to remove that, but not in this one. So nice comment :)

HannesDS:
> I think this might break provisioner, suddenly we don't have types for technical assets. Perhaps check with @stijn-janssens if this would break things when trying to privision technical assets.
> 
> We also need a solution ofcourse, since out of tree plugins can't change our generated SDK

Inded, so I added a line in the release notes to show that the provisioner breaks. They are typed again, due to being added/copied to the sdk now. This runs in a pre-commit hook. However it still breaks since the import path changes.

backend/app/technical_asset_configuration/base_schema.py
```diff
@@ -109,22 +114,22 @@ class PlatformMetadata(ORMModel):
...
+
     version: ClassVar[str] = "1.0"
     mcp_instructions: ClassVar[str] = ""
-    configuration_type: DataOutputTypes
 
-    # Platform metadata - should be overridden in subclasses
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def configuration_type(self) -> str:
+        return self.name
```
stijndehaes: Can't we just remove this?

backend/app/technical_asset_configuration/schema_response.py
stijndehaes: Well the still are not here anymore :D And indeed better documentation is actually the right answer. Let's add that to todo, and just remove comments

backend/app/technical_asset_configuration/schema_union.py
stijndehaes: It's still based on `configuration_type`, but I think we should use name no? And just remove `configuration_type`

scripts/check_generated_plugin_configurations.sh
```diff
@@ -0,0 +1,31 @@
...
+    if [[ -z "$(git status --porcelain ./sdk/plugins)" ]];
+    then
+      exit 0
+    else
+      echo "Git is dirty"
+      git status --porcelain ./sdk/plugins
+      git --no-pager diff
+      exit 1
+    fi
+  fi
```
stijndehaes, line 23: I think some mixup happened, there are other ways to check that new files where generated `git status --porcelain ./sdk/plugins` already reports newly generated untracked files no?

THe add was on purpose in the other script I was fed up with forgetting to add certain code that was generated, but perhaps that was not the right move? Then we can just remove the add

sdk/generation/build_plugin_configurations.py
stijndehaes, line 1: This is a very complex file that is hard to understand can you add some coments to wat is tries to achieve?


Also `KNOWN_ENUMS` could use some info

stijndehaes (changes requested):

backend/app/technical_asset_configuration/base_schema.py
HannesDS: Well yes and no, I renamed to name, and removed name as a CLassvar above in the base. it runs out that name: str = CONFIGURATION_TYPE looks like it should behave like a ClassVar, but in Pydantic v2 it categorically doesn't. The computed_field is what makes that ClassVar's value show up in model_dump(). It sounds fair to me, although I am not an expert.

sdk/generation/build_plugin_configurations.py
HannesDS: Me and my ai added a docstring in the beginning.

scripts/check_generated_plugin_configurations.sh
HannesDS: Yes indeed. The git add `.sdk/plugins` has been removed

backend/app/shared/schema.py
```diff
@@ -12,6 +12,16 @@ def is_pydantic(cls, obj: object) -> bool:
         """Checks whether an object is pydantic."""
         return type(obj).__class__.__name__ == "ModelMetaclass"
 
+    @staticmethod
+    def _orm_kwargs(schema: "ORMModel") -> dict:
```
stijndehaes, line 16: Have we added tests for this behavior? I assume this class is rather untested, OK to add this as a ticket

sdk/generation/build_plugin_configurations.py
stijndehaes: Ok now I understand it, but why does this have to be so difficult. It's doing more then just copying right, it's constructing a new file. Can't we just make sure that the file we want can be copied without having to change it?

OK to simplify in follow up PR, if possible

stijndehaes (approved):

sdk/generation/build_plugin_configurations.py
HannesDS: Fair enough, I added a todo in my ticket.

PR #4328, opened 2026-09-16
Dynamically Loadable plugins:: Migrate plugin tables from the plugin's own revisions

HannesDS:
A plugin that owns a table needs to ship changes to it, and the portal's
own Alembic history is the wrong place: it lives in the core project,
which is what a plugin is supposed to avoid touching.

A plugin declares the revision it wants and the package holding its
versions directory, and the existing deploy-time migration step brings
its table to that revision, upgrading or downgrading, after the core
migrations run.

Every plugin's versions directory goes into one Alembic environment
through version_locations. Each plugin's first revision has no
down_revision, so each is a separate branch with its own base and head,
and Alembic's own one-row-per-head tracking gives one shared
plugin_migration_state table rather than one table per plugin. That only
works while every call knows every row, so a row left by a plugin that
was uninstalled or failed to import is reported by name instead of
surfacing as Alembic's "Can't locate revision".

No plugin declares a target revision yet, so this runs and does nothing.

## Deploy-time migration

```mermaid
sequenceDiagram
    participant Deploy as python -m app.db_tool migrate
    participant Core as Alembic (core)
    participant Rec as reconcile_all
    participant DB as Postgres

    Deploy->>Core: upgrade heads
    Deploy->>Rec: reconcile_all(discovered plugins)
    Rec->>Rec: one Config, version_locations = every plugin's versions/
    Note over Rec: one environment, many bases.<br/>Each plugin's first revision has<br/>no down_revision, so each is a branch.
    Rec->>DB: read plugin_migration_state
    alt a row belongs to no installed plugin
        Rec-->>Deploy: error naming the orphan revisions
    end
    loop each plugin
        Rec->>DB: upgrade or downgrade to its target_revision
    end
```

All of our own plugins from now on use this system; So changes to those configuration tables should happen through those alembic files. 

its base is the PR below it.

| # | PR | |
|---|----|-|
| 0 | `fix/precommit-sdk-hook-staging` | #4324 |
| 1 | `feat/4325-plugin-registry` | #4326 |
| 2 | `feat/4325-generic-configuration` | |
| 3 | `feat/4325-plugin-migrations` | |
| 4 | `feat/4325-optional-platform-service` | |
| 5 | `feat/4325-plugins-out-of-tree` | |

Each is functional on its own: none leaves the portal broken if the ones above
it never merge.

---


backend/app/plugins/alembic_runner/env.py
```diff
@@ -0,0 +1,37 @@
+from alembic import context
+from sqlalchemy import engine_from_config, pool
+
+config = context.config
+target_metadata = None
+
+
+def run_migrations_online() -> None:
```
stijndehaes, line 8: We can't have a single migrations runner? I believe we already have one for our main app? This question might be stupid :D

backend/app/plugins/migrations.py
```diff
@@ -0,0 +1,144 @@
...
+@contextmanager
+def _config(
+    plugins: Sequence[type[TechnicalAssetPlugin]], url: str
+) -> Iterator[Config]:
+    with ExitStack() as stack:
+        version_locations = [
+            str(
+                stack.enter_context(
+                    resources.as_file(
+                        resources.files(plugin.migrations_package) / "versions"
```
stijndehaes, line 33: Should `migrations_package` include the versions? 

Btw we can also make this a default location that is relative from where the plugin is defined no?

backend/app/technical_asset_configuration/azure_blob/schema.py
```diff
@@ -31,6 +31,8 @@
 class AzureBlobTechnicalAssetConfiguration(TechnicalAssetPlugin):
     name: ClassVar[str] = NAME
     version: ClassVar[str] = "1.0"
+    target_revision: ClassVar[str] = "azure_blob_0001_baseline"
+    migrations_package: ClassVar[str] = "app.technical_asset_configuration.azure_blob"
```
stijndehaes, line 35: Shouldn't this target revision be auto detected? I believe that is how our backend works.

backend/app/plugins/migrations.py
stijndehaes, line 1: I would review this file, I think maybe `target_revision` should not be defined by the plugin at all.

And maybe `migrations_package` that location could be auto detected, if there is a versions folder we try to load sql alchemy migrations automatically and run them. I don't think the user even has to explicitly set that location

stijndehaes (changes requested):

backend/app/plugins/alembic_runner/env.py
HannesDS: Not a stupid question! We can have a single runner, but there is a slight difference in runners as I understand it. THe "core" app has always a single head, the single version we are on now. While the plugins have N heads with N the amount of plugins installed. So currently for the core alembic we always assume there exists only one. While now the lpugins there are multiple. So it would break a CI job and I thought we would want them split. But I am open for discussion.

HannesDS: Discussed, will try to move it to a single runner for simplicity

backend/app/plugins/migrations.py
HannesDS: Target revesion is overkill and makes thing unneededly complex. 
- migraions package, hardcode it under versions folder where the plugin class is actually defined. Make sure to put into the docs where users need to write it.

backend/app/plugins/alembic_runner/env.py
vlieven: I believe what Stijn means is if this should not be a single module that handles all the migrations of the application (both of portal core + plugins), instead of having these separate in different areas of the codebase.

backend/app/plugins/registry.py
```diff
@@ -25,6 +25,11 @@ def _discover(self) -> dict[str, type[TechnicalAssetPlugin]]:
                 raise Exception(
                     "The registered plugin does not subclass TechnicalAssetPlugin or is not a class"
                 )
+            if bool(plugin.target_revision) != bool(plugin.migrations_package):
```
vlieven: Is target_revision just not always the latest? Seems a bit of an edge case to allow plugins to already declare future revisions without applying them

HannesDS: After reading the comments "mijne frank is gevallen" and understood that indeed you always want the latest and there should not be a reason to want something else, so target revesion is removed.

backend/app/database/alembic/env.py
```diff
@@ -65,7 +65,11 @@ def run_migrations_online() -> None:
     and associate a connection with the context.
     """
     config_section = config.get_section(config.config_ini_section, {})
-    config_section["sqlalchemy.url"] = get_url()
+    # A plugin's reconciliation config supplies its own sqlalchemy.url, since
+    # its version_locations aren't known to this file's static alembic.ini.
+    # Bare `alembic` CLI usage never sets it, so it falls back to get_url().
```
stijndehaes:
```suggestion
```

I don't think this is needed?

backend/app/technical_asset_configuration/azure_blob/versions/__init__.py
stijndehaes, line 1: init files aren't needed in our backend

backend/app/technical_asset_configuration/azure_blob/versions/azure_blob_0001_baseline.py
```diff
@@ -0,0 +1,40 @@
...
+        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
+        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
+        sa.Column("deleted_at", sa.DateTime(timezone=False), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    if TABLE not in sa.inspect(op.get_bind()).get_table_names():
+        return
+    op.drop_table(TABLE)
```
stijndehaes:
```suggestion
def upgrade() -> None:
    op.create_table(
        TABLE,
        sa.Column(
            "id",
            postgresql.UUID(as_uuid=True),
            sa.ForeignKey("data_output_configurations.id", ondelete="CASCADE"),
            primary_key=True,
        ),
        sa.Column("domain", sa.String(), nullable=False),
        sa.Column("container_name", sa.String(), nullable=False),
        sa.Column("path", sa.String(), nullable=True),
        sa.Column("created_on", sa.DateTime(timezone=False), server_default=utcnow()),
        sa.Column("updated_on", sa.DateTime(timezone=False), onupdate=utcnow()),
        sa.Column("deleted_at", sa.DateTime(timezone=False), nullable=True),
        if_not_exists=True,
    )


def downgrade() -> None:
    op.drop_table(TABLE, if_exists=True))
```

stijndehaes: Same for all the others

backend/app/plugins/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+            "give up its migration history."
+        )
+
+    before = set(_current_heads(engine))
+
+    # Core's head first, then heads: core and a plugin can create the same table and only the plugin's create is guarded, so upgrading everyone to heads in one call could let Alembic run the plugin's create before core's unguarded one and fail on a duplicate table.
+    command.upgrade(config, _core_head(url))
+    command.upgrade(config, "heads")
+
+    results: dict[str, str] = {}
```
stijndehaes: I think we can drop the results, they are not used in any function I think except to log. But we can just log from here

backend/app/plugins/migrations.py
```diff
@@ -0,0 +1,159 @@
...
+            results[plugin.name] = f"upgraded {previous} to {head}"
+        logger.info(f"Reconciled plugin '{plugin.name}': {results[plugin.name]}")
+
+    return results
+
+
+def check_latest_migration_core(
+    plugins: Sequence[type[TechnicalAssetPlugin]], engine: Engine
+) -> str:
+    """Downgrade core's latest migration by one step, then reapply it.
```
stijndehaes:
```suggestion
    """ONLY USE IN TESTINGS: Downgrade core's latest migration by one step, then reapply it.
```

backend/app/db_tool.py
```diff
@@ -38,16 +40,33 @@ def migrate():
...
     try:
-        cfg = Config(
-            os.path.join(os.path.dirname(os.path.abspath("__file__")), "alembic.ini")
-        )
-        command.upgrade(cfg, "heads")
+        results = migrate_all(plugin_registry.discovered(), engine)
+        for plugin_name, outcome in results.items():
+            print(f"  {plugin_name}: {outcome}")
+        if not results:
+            print("  no plugins own a table")
```
stijndehaes: I don't think we should ever use print, I know the script already had it. Also let's not return result but just let `migrate_all` log results no?

backend/tests/app/search_output_ports/test_router.py
```diff
@@ -214,11 +223,25 @@ def setup(session) -> tuple[OutputPort, OutputPort, OutputPort]:
...
-        )
+        # A bare alembic.ini Config only knows core's own versions folder, but
+        # the shared version table also holds every plugin's tracked revision
+        # by this point in the suite - it can't resolve those rows at all.
+        # Build the same all-branches config migrate_all does. Core's own
+        # downgrade unconditionally drops a plugin's table too (the same
+        # overlap migrate_all works around going up), so bring core down to
+        # base first, same order as migrate_all's own upgrade, before the
+        # rest of the table: a plugin's own downgrade only no-ops once its
+        # table is already gone.
```
stijndehaes:
```suggestion
        # Downgrade plugins first, then downgrade core. 
```

I am unsure if my simple comment is needed even, it's pretty obvious no?

stijndehaes (changes requested):
Should we also remove the specific technical asset migrations from our core migrations?
Now they overlap with the plugin migrations...

HannesDS:
> Should we also remove the specific technical asset migrations from our core migrations? Now they overlap with the plugin migrations...

I removed as much as possible, some migrations are history dependant and the core migration still does something, but it's not table creation.

backend/app/plugins/migrations.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def _core_head(plugins: Sequence[type[TechnicalAssetPlugin]], url: str) -> str:
+    # A few core migrations `depends_on` a plugin's baseline (see those files
+    # for why), so resolving core's own revisions requires every such
+    # plugin's directory to be present too - a core-only config raises
+    # `KeyError` on that dependency. Building the same full config
+    # migrate_all uses, then picking out the head whose file lives under
+    # core's own versions folder, works regardless of which plugins that
+    # happens to include.
```
stijndehaes, line 77: I don't understand this comment, which are the few core migrations that depends_on plugin base line?

Maybe we can discuss in a call?

backend/app/plugins/migrations.py
```diff
@@ -0,0 +1,166 @@
...
+    for plugin in plugins_owning_tables:
+        head = _own_head(plugin, url)
+        previous = next((h for h in before if h in _own_revisions(plugin, url)), None)
+        if previous == head:
+            outcome = f"up to date at {head}"
+        elif previous is None:
+            outcome = f"installed at {head}"
+        else:
+            outcome = f"upgraded {previous} to {head}"
+        logger.info(f"Reconciled plugin '{plugin.name}': {outcome}")
```
stijndehaes, line 136: Is this usefull logging? Doesn't sql alchemy maybe log enough? By removing this we can remove `_own_revisions`, `_own_head` making thing simpler

backend/app/plugins/migrations.py
stijndehaes, line 1: In general this file is very complex, can we see if we can reduce some complexity? 
THe rest of the PR is now super simpel, which is great.

stijndehaes (approved):
Approved, but maybe we can have a quick look still to see if we can reduce the complexity of migrations a bit?

PR #4329, opened 2026-09-16
Dynamically loadable plugins: Let a plugin skip platforms and platform services

HannesDS:
A plugin that has no platform behind it, a repository link for instance,
still had to name a platform and a platform service to create a technical
asset. The database already allowed both to be absent; only the schemas
and two computed fields required them.

platform_id, service_id and service become optional on the request, the
response and the internal schema. result_string reads its template from
the plugin itself when there is no platform service, through the same
render_template every built-in type uses, and technical_info returns
nothing, because technical info is per environment platform configuration
and there is no platform to have one.

A technical asset backed by a platform service behaves exactly as before,
which the existing suite and a regression test both cover.

The form stops asking the backend to render a result string when there is
no platform service to render it against, since that endpoint resolves
its template from the platform and service pair.

its base is the PR below it.

---

- [ ] Updated sample data so the new feature can be easily tested.


frontend/src/components/data-products/technical-asset-form/technical-asset-form.component.tsx
```diff
@@ -162,8 +162,11 @@ export function TechnicalAssetForm({ mode, formRef, dataProductId, modalCallback
         }
     };
 
+    const resultRequestId = useRef(0);
```
stijndehaes: What is the point of this thing? I have the feeling it can just be removed?

frontend/src/tests/pages/data-product/components/data-product-tabs/data-output-tab/components/add-technical-asset-popup/add-technical-asset-popup.test.tsx
```diff
@@ -125,4 +132,49 @@ describe('TechnicalAssetPopup', async () => {
...
+        );
+
+        await fillInNameAndDescription(user);
+        await fillInGitHub(user);
+
+        await waitFor(() => expect(screen.getByLabelText(/repository link/i)).toHaveValue(''));
+        expect(renderSpy).not.toHaveBeenCalled();
+    }, 15000);
+
+    it('should not restore a stale access path when switching to a platformless plugin while a render request is in flight', async () => {
```
stijndehaes: Aha I assume this is the reason for the request counter, another way to achieve this is just blocking selecting another plugin until the request is done no? Maybe that is a better solution

stijndehaes (changes requested):

stijndehaes (approved):

PR #4345, opened 2026-09-17
Add OIDC enabled tests in device router to catch issues we missed before

stijn-janssens:


backend/tests/app/auth/test_device_router.py
```diff
@@ -1,14 +1,62 @@
...
+from app.core.auth.device_flows.service import verify_auth_header
+from app.database.database import get_system_db_session
+from app.settings import settings
+from tests.conftest import override_unauthenticated_get_db
 from tests.factories import DeviceFlowFactory
 
 ENDPOINT = "/api/v2/authn/device"
 
 
+@pytest.fixture(params=["oidc_disabled", "oidc_enabled"])
```
stijndehaes, line 20: This override the client fixutre used in other tests right? Should we call it another name to avoid confusion?

stijndehaes (approved):

PR #4357, opened 2026-09-18
Feature: add support for groups and machine users

dixie-flatline999:
- [X] Updated sample data so the new feature can be easily tested.

This PR introduces groups and machine users through the polymorphic Identity entity described in ADR 0023.

High-level changes:

- Adds the Identity, Group, MachineUser, and GroupMembership entities and their corresponding database tables.
- Adds database migrations that create identity records for existing users while preserving their IDs and existing relationships. Also moves common identity fields such as external_id from User to Identity.
- Updates Data Product and global role assignments to target identities, allowing assignments to users, groups, and machine users.
- Doesn't touch output port role assignments against users as these are deprecated.
- Only users and machine users can belong to groups, avoiding nested groups.
- Updates Casbin synchronization so users inherit Data Product and global permissions assigned to their groups.
- Updates Data Product visibility queries to consider both direct user assignments and assignments inherited through group membership.
- Keeps the existing public role assignment API temporarily user specific until the frontend is migrated, so this PR doesn't become even bigger. These endpoints continue to expose user and user_id and currently exclude group and machine user assignments.
- Adds tests covering identity polymorphism, group membership, assignment cleanup, Casbin inheritance, and Data Product visibility.

As a side note, the reformatted old migration is required as it was using the service, which uses the current model, thus referencing the identity table which doesn't exist at that point.


backend/app/authorization/role_assignments/data_product/router.py
```diff
@@ -102,9 +103,10 @@ def list_data_product_role_assignments(
     return ListDataProductRoleAssignmentsResponse(
         role_assignments=RoleAssignmentService(db).list_assignments(
             data_product_id=data_product_id,
-            user_id=user_id,
+            identity_id=user_id,
             role_id=role_id,
             decision=decision,
+            users_only=True,
```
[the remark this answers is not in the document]
dixie-flatline999: users_only is passed to the `list_assignments` method because it is both used in that api endpoint and in the auth flow, so the restriction to only users or not needs to be made depending on the caller

backend/app/authorization/role_assignments/global_/router.py
```diff
@@ -114,7 +114,9 @@ def list_global_role_assignments(
 ) -> ListGlobalRoleAssignmentsResponse:
     return ListGlobalRoleAssignmentsResponse(
         role_assignments=RoleAssignmentService(db).list_assignments(
-            user_id=user_id, role_id=role_id
+            identity_id=user_id,
+            role_id=role_id,
+            users_only=True,
```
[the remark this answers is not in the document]
dixie-flatline999: same logic as above

backend/app/database/alembic/versions/2026_09_04_1307-1d1b811732f3_add_identity_hierarchy.py
```diff
@@ -0,0 +1,180 @@
...
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+from app.shared.model import utcnow
+
+revision: str = "1d1b811732f3"
+down_revision: Union[str, None] = "c8fbf0cf31e4"
```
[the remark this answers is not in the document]
dixie-flatline999: done

backend/app/authorization/role_assignments/data_product/email.py
```diff
@@ -20,8 +20,7 @@ def send_role_assignment_request_email(
     action = emailgen.Table(["User", "Request", "Data Product", "Owned By"])
     action.add_row(
         [
-            f"{role_assignment_user.first_name} {role_assignment_user.last_name}",
-            "Wants to join ",
+            f"{identity_display_name} wants to join ",
```
stijn-janssens: You have merged 2 columns of the generated table in the email, without actually removing the columns.
I think it is better to keep the `identity_display_name` in its own column


```suggestion
            identity_display_name, 
            "Wants to join ",
```

backend/tests/factories/machine_user.py
```diff
@@ -0,0 +1,14 @@
...
+from tests.factories.identity import IdentityFactory
+
+fake = Faker()
+
+
+class MachineUserFactory(IdentityFactory):
+    class Meta:
+        model = MachineUser
+
+    display_name = factory.Sequence(lambda n: f"Machine USer {n}")
```
stijn-janssens:
```suggestion
    display_name = factory.Sequence(lambda n: f"Machine User {n}")
```

stijn-janssens (changes requested):
Overall, super clean PR 👏 . I have made 2 tiny requests to be changed.
Something to consider for a follow-up PR: Do we want to change the API key behavior?
Currently we create a system user who has permanent admin rights, this is how the global api key works (you log in as that user). Might make more sense to make this a machine user instead that somehow has global admin rights.
Could also be a topic for us to pick up, not you 😉

backend/app/users/notifications/service.py
```diff
@@ -106,7 +106,7 @@ def create_data_product_notifications(
 
         receivers = set(
             chain(
-                (assignment.user_id for assignment in assignments),
+                (assignment.identity_id for assignment in assignments),
```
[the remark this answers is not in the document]
stijn-janssens: this one might be a valid comment to consider

backend/app/authorization/service.py
```diff
@@ -136,3 +142,35 @@ def _sync_output_ports_reader_role(self):
...
+                group_id=membership.group_id,
+            ):
+                changes += 1
+
+            for data_product_id in assigned_data_products.get(
+                membership.group_id,
+                set(),
+            ):
+                total += 1
+                if self.authorizer.assign_resource_group_membership(
```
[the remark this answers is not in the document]
dixie-flatline999: Correct! I forgot to enforce membership edges on group membership updates and data product role assignments against groups. implemented now

backend/app/authorization/role_assignments/data_product/email.py
dixie-flatline999: absolutely... I joined them for styling without realizing they are part of a table....

backend/app/authorization/role_assignments/data_product/email.py
```diff
@@ -20,8 +20,7 @@ def send_role_assignment_request_email(
...
             role_assignment_data_product.name,
             ", ".join([f"{user.first_name} {user.last_name}" for user in approvers]),
```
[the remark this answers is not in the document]
dixie-flatline999: already accepted Stijn's suggestion

backend/tests/factories/machine_user.py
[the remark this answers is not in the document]
dixie-flatline999: already accepted Stijn's suggestion

backend/app/users/notifications/service.py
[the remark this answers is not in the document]
dixie-flatline999: good catch!! solved!

backend/app/authorization/role_assignments/data_product/service.py
```diff
@@ -187,7 +205,7 @@ def get_pending_data_product_role_assignments(
         data_product_ids = (
             select(DataProductRoleAssignmentModel.data_product_id)
             .where(
-                DataProductRoleAssignmentModel.user_id == user.id,
+                DataProductRoleAssignmentModel.identity_id == user.id,
```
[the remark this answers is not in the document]
dixie-flatline999: also valid, added group membership to the logic

backend/app/data_products/output_port_technical_assets_link/service.py
```diff
@@ -175,7 +175,8 @@ def get_user_pending_actions(
                         TechnicalAssetOutputPortAssociationModel.output_port.has(
                             OutputPortModel.data_product.has(
                                 DataProductModel.assignments.any(
-                                    DataProductRoleAssignmentModel.user_id == user.id
+                                    DataProductRoleAssignmentModel.identity_id
+                                    == user.id
```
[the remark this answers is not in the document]
dixie-flatline999: also valid remark, fixed

backend/app/database/alembic/versions/2026_09_22_1315-18fb80f18c7d_migrate_role_assignments_to_identities.py
```diff
@@ -0,0 +1,180 @@
...
+            sa.ForeignKey("groups.id", ondelete="CASCADE"),
+            primary_key=True,
+            nullable=False,
+        ),
+        sa.Column(
+            "member_identity_id",
+            postgresql.UUID(as_uuid=True),
+            sa.ForeignKey("identities.id", ondelete="CASCADE"),
+            primary_key=True,
+            nullable=False,
```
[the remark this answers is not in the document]
dixie-flatline999: also valid and a nice optimization as the lookup by member_identity_id is somewhat frequent

backend/app/data_products/output_ports/input_ports/service.py
```diff
@@ -290,7 +290,8 @@ def get_user_pending_actions(self, user: User) -> Sequence[InputPortRequest]:
                         InputPortModel.output_port.has(
                             OutputPortModel.data_product.has(
                                 DataProductModel.assignments.any(
-                                    DataProductRoleAssignmentModel.user_id == user.id
+                                    DataProductRoleAssignmentModel.identity_id
+                                    == user.id
```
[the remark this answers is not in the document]
dixie-flatline999: same issue as the previous remark, also fixed now

dixie-flatline999:
> Overall, super clean PR 👏 . I have made 2 tiny requests to be changed. Something to consider for a follow-up PR: Do we want to change the API key behavior? Currently we create a system user who has permanent admin rights, this is how the global api key works (you log in as that user). Might make more sense to make this a machine user instead that somehow has global admin rights. Could also be a topic for us to pick up, not you 😉

Could be done but will need some work. Although now machine users can have a global role assigned, only users are considered in the auth layer.

backend/app/authorization/role_assignments/data_product/model.py
```diff
@@ -50,5 +53,5 @@ def to_event(self) -> DataProductRoleAssignmentEvent:
         return DataProductRoleAssignmentEvent(
             id=self.id,
             data_product_id=self.data_product_id,
-            user_id=self.user_id,
+            user_id=self.identity_id,
```
vlieven, line 56: I think the user_id can be dropped from this event, no?

backend/app/authorization/role_assignments/data_product/router.py
```diff
@@ -68,13 +70,20 @@ def delete_data_product_role_assignment(
 
     if assignment.decision is DecisionStatus.APPROVED:
         DataProductAuthAssignment(assignment).remove()
+        # Enforces group role inheritance removal for its members
+        group_service = GroupService(db)
+        if group_service.is_group(assignment.identity_id):
+            group_service.remove_data_product_membership_edges(
+                group_id=assignment.identity_id,
+                data_product_id=assignment.data_product_id,
+            )
```
vlieven, line 79: Why is the `remove_data_product_membership_edges` needed? Should the AuthAssignment not be to the Group, and the members of said group inherit the permissions of the group?

backend/app/authorization/role_assignments/data_product/router.py
```diff
@@ -212,10 +222,18 @@ def create_data_product_role_assignment(
...
         )
         DataProductAuthAssignment(assignment).add()
+
+        # Enforces group role inheritance for its members
+        group_service = GroupService(db)
+        if group_service.is_group(assignment.identity_id):
+            group_service.add_data_product_membership_edges(
+                group_id=assignment.identity_id,
+                data_product_id=assignment.data_product_id,
+            )
```
vlieven, line 232: Same remark, should the assignment not be to the group?

backend/app/data_products/service.py
```diff
@@ -155,10 +160,22 @@ def get_data_products(
...
-                        decision=DecisionStatus.APPROVED,
+                        and_(
+                            DataProductRoleAssignment.decision
+                            == DecisionStatus.APPROVED,
+                            or_(
+                                DataProductRoleAssignment.identity_id
+                                == current_user.id,
+                                DataProductRoleAssignment.identity_id.in_(
+                                    user_group_ids
+                                ),
```
vlieven, line 176: This assignment condition comes back a few times, would it make sense to extract it somewhere?

backend/app/database/alembic/versions/2025_05_21_1055-6441623a586b_grant_api_user_admin.py
vlieven, line 1: Is this supposed to be a refactor to not having to import the role_assignment functions?
I would keep the admin UUID and system account identifier as separate variables for readability

dixie-flatline999: ~~well this was agreed on the standup. the migration is wrong is the first place as it uses the ORM model instead of explicit SQL queries, so now that the user model has changed, this migration tries to reference a table (via the model) that doesn't exist yet, as the migration that includes the identity table is down the line~~ <- this answered your previous message.

technically using an all 0 UUID is more correct as that was the value of the Admin UUID at that moment, which can change at any time. functionality stays the same, a fresh launch works the same and in already running systems this migration is way in the past

backend/app/data_products/service.py
dixie-flatline999: you are right I'll see if I can put it somewhere elegantly

backend/app/database/alembic/versions/2025_05_21_1055-6441623a586b_grant_api_user_admin.py
```diff
@@ -1,92 +1,55 @@
-"""grant API user admin
-
-Revision ID: 6441623a586b
-Revises: 7727032896e7
-Create Date: 2025-05-21 10:55:00.964055
-
-"""
-
```
vlieven, line 8: This does not have to be removed

backend/app/authorization/role_assignments/data_product/model.py
dixie-flatline999: absolutely. stijn explained to me the user_id shouldn't have been there in the first place but forgot to remove it

backend/app/authorization/role_assignments/data_product/router.py
dixie-flatline999: the users do not automatically inherit the permissions and it needs to be explicitly set. so a group has a role against a data product which in casbin is `g(group_id, role_id, data_product_id)` but then the user needs also to have a relationship with the data product via the group, in casbin `g(user_id, group_id, data_product_id)`.

in this method the group is getting the role revoked via `DataProductAuthAssignment(assignment).remove()` and if the entity is a group, all its members need to get it revoked also.

backend/app/authorization/role_assignments/data_product/router.py
dixie-flatline999: see answer above

backend/app/database/alembic/versions/2026_09_07_1315-18fb80f18c7d_migrate_role_assignments_to_identities.py
vlieven, line 1: These two migrations can be merged into one (less steps to execute)

dixie-flatline999: right!

backend/app/authorization/role_assignments/data_product/model.py
dixie-flatline999: right, now I remember. this changes the api spec, which can be considered a breaking change. I'd rather leave that out of this PR

backend/app/data_products/service.py
dixie-flatline999: honestly I checked it out and I don't see a clear place to place a helper method. also the queries where it is used are also kinda repeated, but imo this way makes the query more explicit which i think is nice

backend/app/users/model.py
```diff
@@ -128,5 +105,5 @@ def __hash__(self) -> int:
         return hash(self.id) if self.id is not None else id(self)
 
 
-def ensure_user_exists(user_id: UUID, db: Session) -> User:
-    return ensure_exists(user_id, db, User)
+def ensure_user_exists(user_id: UUID, db: Session, options: list = []) -> User:
```
vlieven: `[]` is a mutable default (I'm surprised that the linter does not complain about this, but that's a different issue), this is a safer way to express the same thing

```suggestion
def ensure_user_exists(user_id: UUID, db: Session, options: Sequence[Any] = ()) -> User:
```

backend/app/groups/service.py
```diff
@@ -0,0 +1,130 @@
...
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Member not found.",
+            )
+        return membership
+
+    def get_group(self, group_id: UUID) -> Group:
+        return ensure_group_exists(group_id, self.db)
+
+    def is_group(self, identity_id: UUID) -> bool:
+        return self.db.get(Group, identity_id) is not None
```
vlieven: This is only used in a test, I think it can be removed now?

vlieven (approved):
One minor style issue, which does not impact correctness I think

backend/app/groups/service.py
dixie-flatline999: right, leftover from yesterday changes

backend/app/users/model.py
dixie-flatline999: good one, PyCharm catches it without much fuss

stijn-janssens (approved):

PR #4392, opened 2026-09-22
Ensure the consumer has access

stijndehaes:
They have access to private output port
or hidden data product as long as you
are owner of a data product that consumes
it.

In support of: #4162


backend/app/data_products/output_ports/input_ports/service.py
```diff
@@ -85,6 +85,11 @@ def get_link(
             )
         return current_link
 
+    def _sync_hidden_data_product_access(self, data_product_id: UUID) -> None:
+        from app.data_products.service import DataProductService
+
+        DataProductService(self.db)._sync_consumer_reader_grouping(data_product_id)
```
HannesDS, line 104: Not sure why this method is added and just not directly called?

backend/app/authorization/service.py
```diff
@@ -59,6 +59,11 @@ def reload_enforcer(self) -> None:
         self._sync_data_products_reader_role()
         logger.info("Synced data products reader role permissions")
 
+        from app.data_products.service import DataProductService
```
HannesDS, line 63: I have never seen an import mid method, mayb better to put it at the top?

backend/app/core/authz/authorization.py
```diff
@@ -20,11 +30,48 @@
...
+if TYPE_CHECKING:
+    from data_products.model import DataProduct
+
 ID: TypeAlias = Union[str, UUID]
 
 tracer = trace.get_tracer(__name__)
 
 
+@dataclass(frozen=True)
+class ConsumerLink:
```
HannesDS: `InputPortConsumer` ? I don't like Link here again, a bit confusing giving the hustory

backend/app/data_products/service.py
```diff
@@ -84,6 +85,64 @@ def _sync_public_reader_grouping(
             case _:
                 assert_never(visibility)
 
+    def _sync_consumer_reader_grouping(self, data_product_id: UUID) -> None:
+        """
+        Will sync consumer relations for data products and output ports.
+        This is made to ensure hidden data products and private output ports are only accessible to consumers that have been approved.
+        We sync it for every data product, since output ports can be changed from unrestricted to private.
+        """
+        active_links = set(
```
HannesDS: `active_input_ports` Sorry in the jetbrains UI I can't make a suggestion haha

backend/app/data_products/service.py
```diff
@@ -84,6 +85,64 @@ def _sync_public_reader_grouping(
...
+                .where(
+                    OutputPortModel.data_product_id == data_product_id,
+                    InputPortModel.status == InputPortStatus.APPROVED,
+                )
+                .distinct()
+            ).tuples()
+        )
+
+        authorizer = Authorization()
+        current_links = {
```
HannesDS: current_input_ports

backend/app/data_products/service.py
```diff
@@ -84,6 +85,64 @@ def _sync_public_reader_grouping(
...
+            (str(link.consumer_id), str(link.producer_output_port_id))
+            for link in authorizer.get_consumer_links(
```
HannesDS: All referecnes to link -> input_ports

backend/app/data_products/service.py
```diff
@@ -84,6 +85,64 @@ def _sync_public_reader_grouping(
...
+                producer_data_product_id=data_product_id,
+                producer_output_port_id=output_port_id,
+            )
+
+    def sync_data_product_consumer_access(self) -> None:
+        data_product_ids = self.db.scalars(
+            select(DataProductModel.id),
+            execution_options={"skip_data_product_visibility_filter": True},
+        ).all()
+        for data_product_id in data_product_ids:
```
HannesDS, line 152: Well I am wondering what is the most efficicent, but here you loop over all ids, and then again later over all connections (input ports) for all connectison, is it not just simpler, to just calculate ALL active link from teh get go, instead of doing it for every product? I am not sure what the impact is on the DB, but in my mind would make it quicker and simpler?

HannesDS: Ah I just notice that the method to sync a single one, is reused. nevermind

HannesDS (approved):

stijn-janssens (approved):

backend/app/data_products/output_ports/input_ports/service.py
stijndehaes: We could but now we only have to change one location in input_ports if we need extra functionallity. Probably pre optimisation :)

backend/app/core/authz/authorization.py
stijndehaes: Well I renamed all methods to consumer link, should I rename everything.

Like `add_input_port_link` instead of `add_consumer_link` etc?

stijndehaes: Renamed everything

PR #4393, opened 2026-09-22
Fix issue with approve/deny input port

stijndehaes:
When the consumer is a hidden input port
there were issues with the filtering we do.

The issue was first trigger by posthog
analytics not working, so refactored that to
make it so our service always report, and the
analytics mock decided to report yes or no
if posthog is enabled. Ensuring this trigger
always runs in our services.

In support of: #4162


backend/app/data_products/output_ports/input_ports/service.py
```diff
@@ -42,10 +45,24 @@
...
 
+@dataclass
+class ApprovedInputPort:
+    output_port_id: UUID
+    consuming_abstract_data_product_id: UUID
+    requested_by_id: UUID
+
+
+@dataclass
+class DeniedInputPort:
```
HannesDS: Is it really worht it to create 2 identical dataclasses just for the sake of the Denied vs Approved?

backend/tests/app/data_products/output_ports/input_ports/test_service.py
```diff
@@ -98,18 +98,18 @@ def test_approve__renewal_window_starts_at_current_date(self, session):
...
+        InputPortService(session).approve_output_port_as_input_port(
             data_product_id=port.data_product.id,
             output_port_id=port.id,
             consuming_data_product_id=consumer.id,
             actor=actor,
         )
 
-        requests = _by_id(current_link)
-        assert current_link.status == InputPortStatus.APPROVED
+        requests = _by_id(link)
```
HannesDS: Since you touch the file anyway, maybe best to rename link to input_port?

HannesDS (approved):

backend/app/data_products/output_ports/input_ports/service.py
stijndehaes: I guess not, I can call it RedactedInputPort, or just InputPort?

HannesDS: I like RedactedInputPort

